Issue #20, opened 2023-04-06
ClobberWarnings from thrust, cub, spdlog


bdice:
It appears that libucxx is shipping headers for spdlog, thrust, and cub. These headers should not be included in the libucxx package, but rather should be supplied by librmm or spdlog itself. We need to fix the conda dependency list to avoid including these files. It produces hundreds of lines of warnings in the CI output.

Example:
```
ClobberWarning: This transaction has incompatible packages due to a shared path.
  packages: rapidsai-nightly/linux-64::librmm-23.04.00a-cuda11_230321_ge8fbd06e_34, file:///tmp/conda-bld-output/linux-64::libucxx-0.31.00a-cuda11_230406_g06a10c0_16
  path: 'include/rapids/thrust/system/omp/detail/transform_reduce.h'
```

jakirkham:
Think CPM is pulling in RMM (and therefore these other libraries). AFAIK they are not needed for this step. Is there a good way to disable them?

jakirkham:
Maybe we can just remove them manually for now ( https://github.com/rapidsai/ucxx/pull/22 ). Likely some more work is needed on the CMake side

jakirkham:
The PR above should fix the issue. Though it would be good to understand why these headers are getting installed so that we can disable that behavior (as opposed to simply removing them afterwards). Reopening for follow up

jakirkham:
Possibly PR ( https://github.com/rapidsai/ucxx/pull/23 ) has addressed some of the underlying issues. Might be worth rechecking things in a follow up to see what still needs to be done

pentschev:
> Possibly PR ( https://github.com/rapidsai/ucxx/pull/23 ) has addressed some of the underlying issues. Might be worth rechecking things in a follow up to see what still needs to be done

No, unfortunately not. Those changes were already in https://github.com/rapidsai/ucxx/pull/18 where you saw the warnings. I only split those changes into a new PR to make https://github.com/rapidsai/ucxx/pull/18 easier to review that PR and because that was not directly related to CI.

PR #22, opened 2023-04-06
Remove RAPIDS headers from `libucxx`

jakirkham:
Fixes https://github.com/rapidsai/ucxx/issues/20

This is a workaround to remove RAPIDS headers from `libucxx`. Ideally they shouldn't be installed in the first place. However for now just remove them to avoid issues.


pentschev (approved):
LGTM. Thanks John!

PR #23, opened 2023-04-06
Remove unused CMake dependencies and build variables

pentschev:
Remove unused CUDA CMake dependencies and `--build_metrics`/`--incl_cache_stats` build configurations.


vyasr (approved):
Approving assuming that removing the CUDA language is fine (which it appears to be from the build, just waiting on confirmation).

PR #18, opened 2023-04-03
Add CI scripts to build conda packages and run tests

pentschev:

PR #24, opened 2023-04-06
Test fixes and workarounds

pentschev:
Some test fixes, as well as workarounds for some tests that occur in CI but hard to reproduce locally.

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -93,9 +93,6 @@ skip = [
     "__init__.py",
 ]
 
-[tool.pytest.ini_options]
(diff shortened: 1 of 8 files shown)


pentschev:
Changes here were introduced as part of https://github.com/rapidsai/ucxx/pull/18, but since they're not part of the CI setup I cherry-picked them into this PR. To expedite review in https://github.com/rapidsai/ucxx/pull/18 I'm merging this myself.

@wence- would appreciate if you could take a look at the changes here if you have the chance some time this week, we can address anything in a follow-up PR.

Issue #15, opened 2023-04-03
Wakeup may hang for 20 seconds


pentschev:
The wakeup mechanism in UCX may hang up to 20 seconds when using the progress thread, and thus appear that there was a deadlock in the application, for example when running:

```
python python/ucxx/benchmarks/send_recv.py -s 0 -d 1 -o cupy -n 1000000000 -l ucxx-async
```

There are currently two ways to prevent the hang:

1. Specify `--progress-mode thread-polling` to the benchmark above (or `UCXPY_PROGRESS_MODE=thread-polling` for other applications); or
2. Specifying a low keep alive interval: `UCX_TCP_KEEPINTVL=1ms UCX_KEEPALIVE_INTERVAL=1ms`.

Ideally we wouldn't need either one of them. Using `thread-polling` mode means the progress thread will run at 100% CPU all the time, whereas a low keep alive interval may be the source of other issues, perhaps even causing premature timeouts.

pentschev:
This also doesn't happen with UCX 1.14+ and protov2 enable (`UCX_PROTO_ENABLE=y`), protov2 is expected to become default in UCX 1.16.

PR #28, opened 2023-04-11
Signal the worker progress event from the worker progress thread

pentschev:
Without signaling the worker progress event from the worker progress thread while shutting down the thread it may occur that the thread hangs as the event never gets awaken due to the lack of tasks to progress. This change ensures the worker progress event is signaled before joining the thread.

### DIFF
--- a/cpp/src/worker.cpp
+++ b/cpp/src/worker.cpp
@@ -254,19 +254,18 @@ void Worker::startProgressThread(const bool pollingMode)
   if (!pollingMode) initBlockingProgressMode();
   auto progressFunction = pollingMode ? std::bind(&Worker::progress, this)
                                       : std::bind(&Worker::progressWorkerEvent, this);
+  auto signalWorkerFunction =
+    pollingMode ? std::function<void()>{[]() {}} : std::bind(&Worker::signal, this);
 
   _progressThread = std::make_shared<WorkerProgressThread>(pollingMode,
                                                            progressFunction,
+                                                           signalWorkerFunction,
                                                            _progressThreadStartCallback,
                                                            _progressThreadStartCallbackArg,
                                                            _delayedSubmissionCollection);
 }
 
-void Worker::stopProgressThreadNoWarn()
-{
-  if (_progressThread && !_progressThread->pollingMode()) signal();
-  _progressThread = nullptr;
-}
+void Worker::stopProgressThreadNoWarn() { _progressThread = nullptr; }
 
 void Worker::stopProgressThread()
 {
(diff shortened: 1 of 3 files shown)


cpp/src/worker_progress_thread.cpp
```diff
@@ -49,6 +53,7 @@ WorkerProgressThread::~WorkerProgressThread()
   }
 
   _stop = true;
+  _signalWorkerFunction();
   _thread.join();
 }
```
wence-, line 58: Should the `_thread`'s destructor be called when deleting the worker progress thread?

cpp/src/worker.cpp
```diff
@@ -254,19 +254,18 @@ void Worker::startProgressThread(const bool pollingMode)
...
                                                            _progressThreadStartCallbackArg,
                                                            _delayedSubmissionCollection);
 }
 
-void Worker::stopProgressThreadNoWarn()
-{
-  if (_progressThread && !_progressThread->pollingMode()) signal();
-  _progressThread = nullptr;
-}
+void Worker::stopProgressThreadNoWarn() { _progressThread = nullptr; }
```
wence-, line 268: This doesn't need to signal now because `~WorkerProgressThread` arranges that for us?

wence- (approved):
I think this makes sense.

wence-:
> I think this makes sense.

Although that said, it looks like the test suite [segfaulted](https://github.com/rapidsai/ucxx/actions/runs/4667908487/jobs/8264574976?pr=28#step:7:684)?

cpp/src/worker_progress_thread.cpp
pentschev: Not sure I understand the question. The `_thread` destructor will run when the `~WorkerProgressThread` goes out-of-scope, did you expect something different?

cpp/src/worker.cpp
pentschev: Precisely, it must be signaled before we join the thread, which can't be done from here. In any case, I think it makes more sense that `~WorkerProgressThread` performs this operation too, making it less error prone.

pentschev:
> Although that said, it looks like the test suite [segfaulted](https://github.com/rapidsai/ucxx/actions/runs/4667908487/jobs/8264574976?pr=28#step:7:684)?

It didn't segfault, it timed out. I can reproduce that locally and it's hitting https://github.com/rapidsai/ucxx/issues/15 , although in CI it seems to hang indefinitely. That issue is not related to this PR specifically and is happening since we added CI. I'm still investigating it but going to go ahead and merge this when CI passes, as it fixed timeouts when running C++ examples.

Thanks @wence- !

PR #29, opened 2023-04-12
Fix `ucxx::Context` TLS config parser and improve tests

pentschev:
Fix `ucxx::Context` TLS config parser that caused errors when passing `UCX_TLS=^[transports]`. Additionally improve tests.

### DIFF
--- a/cpp/tests/context.cpp
+++ b/cpp/tests/context.cpp
@@ -3,12 +3,18 @@
  * SPDX-License-Identifier: BSD-3-Clause
  */
 #include <cstdlib>
+#include <string>
 
 #include <gtest/gtest.h>
 
 #include <ucxx/api.h>
 
 namespace {
+
+static std::vector<std::string> TlsConfig{"^tcp", "^tcp,sm", "tcp", "tcp,sm", "all"};
+
+class ContextTestCustomConfig : public testing::TestWithParam<std::string> {
+};
 
 TEST(ContextTest, HandleIsValid)
 {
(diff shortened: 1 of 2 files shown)


pentschev:
@wence- FYI, this fixes a bug in the logic you implemented, likely due to an oversight.

cpp/tests/context.cpp
```diff
@@ -19,23 +27,29 @@ TEST(ContextTest, DefaultConfigsAndFlags)
...
   ASSERT_GT(configMapOut.size(), 1u);
   ASSERT_NE(configMapOut.find("TLS"), configMapOut.end());
-  ASSERT_EQ(configMapOut["TLS"], "all");
+  if (const char* envTls = std::getenv("UCX_TLS"))
+    ASSERT_EQ(configMapOut["TLS"], envTls);
+  else
+    ASSERT_EQ(configMapOut["TLS"], "all");
 
   ASSERT_EQ(context->getFeatureFlags(), featureFlags);
 }
```
wence-, line 40: Does it make sense to test the corner cases for the parsing properly? I don't know this test system well, but a naive thing would be to write:

```
  std::vector<std::pair<std::string_view, bool>> cases{{"all", true},
                                                       {"cuda", true},
                                                       {"cuda_copy", true},
                                                       {"rc,cuda", true},
                                                       {"^rc,cuda_ipc", true},
                                                       {"tcp,rc", false},
                                                       {"^rc,tcp,cuda", false},
                                                       {"^cuda_copy", false},
                                                       {"^rc,cuda_copy", false}};
  for (auto&& [v, expect]: cases) {
    ASSERT_EQ(has_support(v), expect);
  }
```
if the has_support call were exposed. I guess otherwise one creates the context and then checks for `_cudaSupport`?

wence- (commented):
Thanks for mixing by logic error! Just wonder if it makes sense to explicitly test for cuda support parsing correctness?

cpp/tests/context.cpp
wence-: Ah, OK, makes sense.

pentschev:
Thanks @wence- for the review. I'll merge this as it has been reviewed already but I'll follow-up on this PR with some improvements. In fact, changes here are not doing anything right now, I overlooked this before and checking whether CUDA support is available isn't even exposed publicly.

quasiben (approved):

Issue #26, opened 2023-04-08
Failing or hanging tests


pentschev:
Some of the tests may non-deterministically fail. They seem to be more prevalent in non-InfiniBand setups, such as CI.

C++ tests (some are possibly related to https://github.com/rapidsai/ucxx/issues/15):

- [ ] `ProgressModes/WorkerProgressTest.ProgressTag/2`
- [ ] Examples

Python (unable to reproduce locally so far):

- [ ] `test_send_recv{,_multi}[{cupy,numba}]`

pentschev:
Does it hang or is it just slow (takes ~20 seconds, as described in #15 )? Is this happening with or without changes from https://github.com/rapidsai/ucxx/pull/30 ? And does `UCX_PROTO_ENABLE=y` change anything?

pentschev:
I actually see the same, but in my case it's mostly `DelayedSubmission/RequestTest.ProgressTagMulti/5` (but `/2` hung once too), `UCX_PROTO_ENABLE=y` didn't help, neither did #30 .

wence-:
Here's a backtrace (not tried with #30 yet)
```
Thread 6 (Thread 0x7fbf69109000 (LWP 34708) "UCXX_TEST"):
#0  0x00007fbf73f49fde in epoll_wait () from /usr/lib/x86_64-linux-gnu/libc.so.6
#1  0x00007fbf73dbd502 in ucs_event_set_wait (event_set=<optimized out>, num_events=num_events@entry=0x7fbf69106774, timeout_ms=timeout_ms@entry=0, event_set_handler=event_set_handler@entry=0x7fbf73e0b6b0 <uct_tcp_iface_handle_events>, arg=arg@entry=0x7fbf69106770) at sys/event_set.c:198
#2  0x00007fbf73e0a5ce in uct_tcp_iface_progress (tl_iface=0x5608c496a9c0) at tcp/tcp_iface.c:333
#3  0x00007fbf745e1b9a in ucs_callbackq_dispatch (cbq=<optimized out>) at /usr/local/src/conda/ucx-1.14.0/src/ucs/datastruct/callbackq.h:211
#4  uct_worker_progress (worker=<optimized out>) at /usr/local/src/conda/ucx-1.14.0/src/uct/api/uct.h:2768
#5  ucp_worker_progress (worker=0x5608c4996e80) at core/ucp_worker.c:2807
[...]

PR #32, opened 2023-04-19
Execute user's callback after setting request status

pentschev:
Once the user-defined callback is executed, it's important that the status for the request is already set, this may be used by the callback itself, as for example in the `ucxx::RequestTagMulti` where the request status must be known to ensure it did result in an error.


cpp/tests/request.cpp
```diff
@@ -212,6 +212,49 @@ TEST_P(RequestTest, ProgressTagMulti)
...
+  requests[1] = _ep->tagRecv(_recvPtr[0], _messageSize, 0, false, checkStatus, recvIndex);
+  waitRequests(_worker, requests, _progressWorker);
+
+  copyResults();
+
+  // Assert status was set before user callback is executed
+  for (const auto status : requestStatus)
+    ASSERT_THAT(status, UCS_OK);
+  for (const auto request : requests)
+    ASSERT_THAT(request->getStatus(), UCS_OK);
```
wence-, line 252: So the logic here is that the callback runs with an updated request status (so we shouldn't observe `UCS_INPROGRESS` in `requestStatus`), and then we need to check that all the requests have a status of `UCS_OK`. Is that right?

If the first fails, it either means that the callback ran too late, or the messages somehow otherwise failed. (In which case the latter check also fails).

is that right?

cpp/src/request.cpp
```diff
@@ -154,6 +147,13 @@ void Request::process()
   }
 
   setStatus(status);
+
+  ucxx_trace_req_f(_ownerString.c_str(),
+                   _request,
+                   _operationName.c_str(),
+                   "callback %p",
+                   _callback.target<void (*)(void)>());
+  if (_callback) _callback(_callbackData);
```
wence-, line 156: This movement aligns the callback call site with the docstring of process.

wence- (approved):
Thanks.

pentschev:
@wence- I pushed the changes we discussed previously in https://github.com/rapidsai/ucxx/issues/26#issuecomment-1527804036 via https://github.com/rapidsai/ucxx/pull/32/commits/caa129f4b3cc8b703770e6914e2ca01eeb486f9b. It seems that the issue of the request result being set before the callback is executed is now resolved and it seems more stable. However, I still do see segfaults if I run tests in a loop, I'm not sure if they are related to changes in this PR or if they are just showing up now because other issues were resolved. Also the issue I was attempting to workaround with https://github.com/rapidsai/ucxx/pull/37 is still present, and is quite annoying since it seems to be only happening frequently only in the `11.2.2, centos7, amd64, 3.9, v100, earliest`, I'm not sure why that is, the conda packages that are installed in other builds are the same, what changes really seems to be limited to OS and CUDA versions.

wence-:
> However, I still do see segfaults if I run tests in a loop, I'm not sure if they are related to changes in this PR or if they are just showing up now because other issues were resolved.

Can you provide a reproducer? Just `while 1; ./UCXX_TESTS` ?

wence-:
> > However, I still do see segfaults if I run tests in a loop, I'm not sure if they are related to changes in this PR or if they are just showing up now because other issues were resolved.
> 
> Can you provide a reproducer? Just `while 1; ./UCXX_TESTS` ?

FWIW, the issues I saw in #26 are not reproducible with this new change.

pentschev:
> Can you provide a reproducer? Just while 1; ./UCXX_TESTS ?

Yes, exactly that.

wence-:
> > Can you provide a reproducer? Just while 1; ./UCXX_TESTS ?
> 
> Yes, exactly that.

I guess my system is slightly different (no nvlink, but not sure if that is relevant for these tests). Using 
```
# Library version: 1.14.0
# Library path: /home/wence/Documents/src/rapids/compose/etc/conda/cuda_11.8/envs/rapids/bin/../lib/libucs.so.0
# API headers version: 1.14.0
# Git branch '', revision 8a97995
```

pentschev:
I can't remember exactly what tests were segfaulting but I can't reproduce that now, furthermore if I skip the same tests that we're already skipping in CI they all complete. On a DGX-1 running the following in a loop seems to always complete successfully:

```
while true; do UCX_TLS=tcp,cuda_copy UCX_TCP_CM_REUSEADDR=y timeout 10m ./cpp/build/gtests/UCXX_TEST --gtest_filter=-*DelayedSubmission*ProgressTagMulti*:ListenerTest.CloseCallback:ListenerTest.IsAlive:ListenerTest.RaiseOnError  ; done
```

I believe those issues may all be related to https://github.com/rapidsai/ucxx/pull/24 and we should tackle them in the near future but are probably not related to this PR.

cpp/include/ucxx/request_tag_multi.h
```diff
@@ -218,7 +218,7 @@ class RequestTagMulti : public std::enable_shared_from_this<RequestTagMulti> {
    *
```
wence-, line 218:
```suggestion
   *
   * @param[in] status the status of the current request.
```

cpp/include/ucxx/request_tag_multi.h
```diff
@@ -233,7 +233,7 @@ class RequestTagMulti : public std::enable_shared_from_this<RequestTagMulti> {
...
```
wence-, line 233:
```suggestion
   *
   * When called, the callback receives a single argument, the status of the current request.
   *
   * @param[in] status the status of the request being completed.
```

wence- (changes requested):
Minor docstring updates for the new `callback` interface, otherwise LGTM!

cpp/include/ucxx/request_tag_multi.h
pentschev: I think the latter makes it more obvious as to what is happening with the request, and it aligns better with the suggestion below too. I've edited the comment and am applying that.

pentschev (commented):
Suggestions are now in, thanks @wence- !

wence- (approved):
Thanks Peter!

ajschmidt8 (approved):
Approving `ops-codeowner` file changes

pentschev:
Thanks @wence- and @ajschmidt8 for reviews!

PR #33, opened 2023-04-26
Dropping Python 3.8

divyegala:
This PR drops Python 3.8 in favor of 3.9.


pentschev (approved):
Thanks @divyegala !

divyegala:
@pentschev let's wait on this a second, I think CI may fail. Has a `branch-0.32` not been released for this repo? I see many references to `branch-23.04` right now

pentschev:
This is a new repository, it's likely behind indeed as I've overlooked it. I'll submit another PR to update that.

pentschev:
> This is a new repository, it's likely behind indeed as I've overlooked it. I'll submit another PR to update that.

Perhaps I should instead be asking if this is something we could automate as we do for other projects @divyegala ? Not sure if this is something I could do or does it need to be ops?

divyegala:
@pentschev other projects have a `ci/release/update-version.sh`. Ops should be setting up a forward-merger so once `branch-xx` is created, it's continuously updated from `branch-x(x-1)` until release. When ops creates `branch-xx` for a new release, they just run `ci/release/update-version.sh` which uses `sed` to find and replace all occurrences manually. 

While this is doable manually and we could both go ahead and create a release, I'd like to tag @ajschmidt8 and ask for his advice to setup the right processes here.

pentschev:
Everything passes now. Thanks @divyegala !

ajschmidt8 (approved):

PR #36, opened 2023-04-26
Update release script

pentschev:
Update release script to make it consistent with other RAPIDS packages.


conda/environments/ucxx-cuda118_arch-x86_64.yaml
```diff
@@ -8,9 +8,11 @@ dependencies:
...
   - pip
 # RAPIDS
-  - cudf=23.06
-  - dask-cudf=23.06
-  - dask-cuda=23.06
+  - cudf==23.4.*
+  - dask-cuda==23.4.*
+  - dask-cudf==23.4.*
+  - librmm==23.4.*
+  - rmm==23.4.*
```
ajschmidt8, line 15: these should all be `23.6`, right?

pentschev: Yes! Thanks for catching. I did a big mess locally and screwed up when fixing, thankfully you noticed! 🙂

pentschev: It's fixed now.

ci/release/update-version.sh
ajschmidt8, line 1: I've been discouraging the use of variables in the search pattern of `sed` commands across RAPIDS (e.g. `sed "s/cudf==${CURRENT_SHORT_TAG_PEP440}/cudf==${NEXT_SHORT_TAG_PEP440}/g"`).

The reason for this is that those search patterns aren't robust and will stop working entirely if the versions fall out of sync (which can happen from time to time).

Ideally we should be able to run consecutive commands like this without issues:

```sh
bash ci/release/update-version.sh 0.29.00
bash ci/release/update-version.sh 0.31.00
bash ci/release/update-version.sh 0.28.00
```
That's not possible with the current changes in this PR.

Instead, I've been suggesting we use more generic commands like this:

```sh
sed "/- cudf==/ s/==.*/==${NEXT_SHORT_TAG_PEP440}\.*/g"
```

This command differs from the first `sed` command above in the following ways:

- It doesn't use any variables for the search pattern
  - this makes it robust. It will always work regardless of the current value of `cudf`
- It uses [sed addresses](https://www.gnu.org/software/sed/manual/sed.html#sed-addresses) to remove some redundancy from the `sed` command
  - that command basically says: "only perform the following replacement on lines that match `/- cudf==/`". `cudf` is only written out once instead of twice.
  - this also makes it really clear which dependency we're trying to update, eliminating ambiguity from dependencies like `cudf` vs. `dask-cudf` and `rmm` vs. `librmm`

For these reasons, I'd like to suggest the changes in the following diff.


<details><summary>diff</summary>
<p>

```diff
diff --git a/ci/release/update-version.sh b/ci/release/update-version.sh
index 46ab4b8..ef6c348 100755
--- a/ci/release/update-version.sh
+++ b/ci/release/update-version.sh
@@ -11,40 +11,34 @@
 # Example: 0.30.00
 NEXT_FULL_TAG=$1
 
-# Get current version
-CURRENT_TAG=$(git tag | grep -xE 'v[0-9\.]+' | sort --version-sort | tail -n 1 | tr -d 'v')
-CURRENT_MAJOR=$(echo $CURRENT_TAG | awk '{split($0, a, "."); print a[1]}')
-CURRENT_MINOR=$(echo $CURRENT_TAG | awk '{split($0, a, "."); print a[2]}')
-CURRENT_PATCH=$(echo $CURRENT_TAG | awk '{split($0, a, "."); print a[3]}')
-CURRENT_SHORT_TAG=${CURRENT_MAJOR}.${CURRENT_MINOR}
-
 #Get <major>.<minor> for next version
 NEXT_MAJOR=$(echo $NEXT_FULL_TAG | awk '{split($0, a, "."); print a[1]}')
 NEXT_MINOR=$(echo $NEXT_FULL_TAG | awk '{split($0, a, "."); print a[2]}')
 NEXT_SHORT_TAG=${NEXT_MAJOR}.${NEXT_MINOR}
 
 # Get RAPIDS version associated w/ ucx-py version
-CURRENT_RAPIDS_VERSION="$(curl -sL https://version.gpuci.io/ucx-py/${CURRENT_SHORT_TAG})"
 NEXT_RAPIDS_VERSION="$(curl -sL https://version.gpuci.io/ucx-py/${NEXT_SHORT_TAG})"
 
 # Need to distutils-normalize the versions for some use cases
-CURRENT_SHORT_TAG_PEP440=$(python -c "from setuptools.extern import packaging; print(packaging.version.Version('${CURRENT_RAPIDS_VERSION}'))")
 NEXT_SHORT_TAG_PEP440=$(python -c "from setuptools.extern import packaging; print(packaging.version.Version('${NEXT_RAPIDS_VERSION}'))")
-echo "Current RAPIDS is ${CURRENT_SHORT_TAG_PEP440}, next is ${NEXT_SHORT_TAG_PEP440}"
+echo "Next tag is ${NEXT_SHORT_TAG_PEP440}"
 
-echo "Preparing release $CURRENT_TAG => $NEXT_FULL_TAG"
+echo "Preparing release: $NEXT_FULL_TAG"
 
 # Inplace sed replace; workaround for Linux and Mac
 function sed_runner() {
     sed -i.bak ''"$1"'' $2 && rm -f ${2}.bak
 }
 
-# bump cudf, dask-cuda and rmm
-sed_runner "s/rmm =${CURRENT_RAPIDS_VERSION}/rmm =${NEXT_RAPIDS_VERSION}/g" conda/recipes/ucxx/meta.yaml
+# bump RAPIDS libs
+sed_runner "/- rmm =/ s/=.*/=${NEXT_RAPIDS_VERSION}/g" conda/recipes/ucxx/meta.yaml
 for FILE in conda/environments/*.yaml dependencies.yaml; do
-  sed_runner "s/cudf==${CURRENT_SHORT_TAG_PEP440}/cudf==${NEXT_SHORT_TAG_PEP440}/g" ${FILE};
-  sed_runner "s/dask-cuda==${CURRENT_SHORT_TAG_PEP440}/dask-cuda==${NEXT_SHORT_TAG_PEP440}/g" ${FILE};
-  sed_runner "s/rmm==${CURRENT_SHORT_TAG_PEP440}/rmm==${NEXT_SHORT_TAG_PEP440}/g" ${FILE};
+  sed_runner "/- cuda==/ s/==.*/==${NEXT_SHORT_TAG_PEP440}\.*/g" ${FILE};
+  sed_runner "/- cudf==/ s/==.*/==${NEXT_SHORT_TAG_PEP440}\.*/g" ${FILE};
+  sed_runner "/- dask-cuda==/ s/==.*/==${NEXT_SHORT_TAG_PEP440}\.*/g" ${FILE};
+  sed_runner "/- dask-cudf==/ s/==.*/==${NEXT_SHORT_TAG_PEP440}\.*/g" ${FILE};
+  sed_runner "/- librmm==/ s/==.*/==${NEXT_SHORT_TAG_PEP440}\.*/g" ${FILE};
+  sed_runner "/- rmm==/ s/==.*/==${NEXT_SHORT_TAG_PEP440}\.*/g" ${FILE};
 done
 
 # rapids-cmake version

```

</p>
</details> 

At some point we will be revamping the entire process around `update-version.sh` for all RAPIDS.

The changes above will be included in that revamp.

pentschev: I'm 100% onboard with this suggestion. I literally copied what's currently being done in cuDF as I thought it was the preferred way, I should have asked first. Thanks for doing the changes and submitting the diff here, it's now been applied. Could you take another look?

ajschmidt8 (approved):

pentschev:
Thanks @ajschmidt8 !

PR #37, opened 2023-04-27
Temporarily increase Python benchmark timeout

pentschev:
Increase the timeout of the Python benchmarks with `thread` progress mode to avoid hitting https://github.com/rapidsai/ucxx/issues/15 . This can be reverted once that issue is resolved.


pentschev:
Thanks @ajschmidt8 !

PR #40, opened 2023-04-28
Revert shared-action-workflows pin

divyegala:
This PR reverts the shared-action-workflows branch pin that was used to drop Python 3.8 to branch-23.06


pentschev (approved):
Thanks @divyegala !

ajschmidt8 (approved):

PR #44, opened 2023-05-02
Update conda pinning

pentschev:
Incorrect conda pinning causes an older UCX version to be picked up, causing various warnings, this change attempts to fix that.


jakirkham (approved):

conda/recipes/ucxx/conda_build_config.yaml
```diff
@@ -15,4 +15,4 @@ python:
   - 3.10
 
 ucx:
-  - ">=1.13.0,<1.15.0"
+  - ">=1.13.0"
```
jakirkham, line 18:
```suggestion
  - 1.13.0
```

pentschev: Won't this be a tight pin though? What if we want to install 1.14.0, will that still work?

jakirkham: This only pins it during the build

As `ucx` has [`run_exports`]( https://github.com/conda-forge/ucx-split-feedstock/blob/8a97995a5cd61a740bbe5a82e55b946fc8399621/recipe/meta.yaml#L21-L22 ), this would become `ucx >=1.13.0,<1.14`

If we wanted to use `1.14.0` as the minimum, we could change this to `1.14.0` instead

pentschev: Can we have a build that supports both `1.13.x` and `1.14.x` though? This is what I was trying to achieve.

jakirkham: We could turn this build into a matrix

```suggestion
  - 1.13.0
  - 1.14.0
```

The pinning itself comes from [the `run_exports`]( https://github.com/conda-forge/ucx-split-feedstock/blob/8a97995a5cd61a740bbe5a82e55b946fc8399621/recipe/meta.yaml#L21-L22 ). Related discussion ( https://github.com/conda-forge/ucx-split-feedstock/pull/92#discussion_r836882363 )

pentschev: That would be better, I think. Would this mean we end up with two different package builds?

jakirkham: Yeah exactly

The challenge is the current constraint that is in `conda_build_config.yaml` doesn't do what we would want either. Conda would look at `ucx >=1.13.0,<1.15.0` in `host` and would pick `ucx 1.14.0` to match that. As a result the `run` line would become `ucx >=1.14.0,<1.15.0`. IOW not compatible with `1.13.x`

pentschev: Does that mean that UCX-Py is also wrong? We have (as far as I understand) precisely the same we have here, [`ucx` in the `host` section](https://github.com/rapidsai/ucx-py/blob/7861326787abccb3cfb7b0f6da2234469cfc3248/conda/recipes/ucx-py/meta.yaml#L30) and pinning to [`>=1.13.0,<1.15.0`](https://github.com/rapidsai/ucx-py/blob/7861326787abccb3cfb7b0f6da2234469cfc3248/conda/recipes/ucx-py/conda_build_config.yaml#L7).

jakirkham (approved):

ajschmidt8 (commented):
Hmm. As far as I can tell from the build output below, there aren't multiple packages being produced here for the different `ucx` versions:

- https://github.com/rapidsai/ucxx/actions/runs/4865422854/jobs/8685792541?pr=44#step:7:8598

I would expect two `libucxx` packages to be built, one for each `ucx` version.

Aside from that, I will say that using the `conda_build_config.yaml` build matrix functionality isn't something that's used in any other RAPIDS recipes to my knowledge.

We might have to think about how to handle this a bit more. For example, presumably we'd want to test each `libucxx` package with the different `ucx` versions, right? So we'd have to figure out how to adjust our scripts to test that.

The easiest path forward here from my perspective would be to just support one version of `ucx` if possible.

If not, we'll need to explore everything I mentioned above (e.g. properly producing two packages, testing each package).

cc: @jakirkham

pentschev:
I'm ok with supporting a single version of UCX for now. Ideally we would still want to support at least a couple of UCX versions though, this allows to circumvent issues regressions that may be found in specific conditions and are _very_ hard to prevent entirely, given we can't test most of UCX in CI (e.g., InfiniBand and NVLink).

ajschmidt8 (approved):

pentschev:
@jakirkham any ideas on the above? Also I was checking locally and we should be good building against the oldest UCX build (e.g., 1.13) and running against newer builds. Would we be ok just building again 1.13 and switching `requirements.run.ucx >=1.13.0<1.15.0`?

pentschev:
I'll merge this to prevent those issues in other PRs, still appreciate if you could comment here @jakirkham .

PR #45, opened 2023-05-02
Fix condition to skip CUDA 11.2 Python async benchmarks in CI

pentschev:
The condition set in https://github.com/rapidsai/ucxx/pull/37 was wrong and this change fixes it.


ajschmidt8 (approved):

pentschev:
Thanks once again @ajschmidt8 !

PR #46, opened 2023-05-05
Add `PythonFutureTask`

pentschev:
Add a new class that allows creating and notifying a Python future when a C++ user-defined task completes.

This is not yet used in UCXX, but we intend to replace the current Python future notification strategy with this one in a follow-up PR.


cpp/python/include/ucxx/python/python_future_task.h
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+ public:
+  /**
+   * @brief Construct a Python future backed by C++ `std::packaged_task`.
+   *
+   * Construct a future object that receives a user-defined C++ `std::packaged_task` which
+   * runs asynchronously using an internal `std::async` that ultimately notifies a Python
+   * future that can be awaited in Python code.
+   *
```
madsbk, line 120:
```suggestion
   *
   * Note that this call will take the Python GIL and requires that the current thread have
   * an asynchronous event loop set.
   *
```

cpp/python/include/ucxx/python/python_future_task.h
```diff
@@ -0,0 +1,213 @@
...
+   * @param[in] result the C++ value that will be converted and set as the result of the
+   *                   Python future.
+   */
+  void setResult(const ReturnType result)
+  {
+    // PyLong_FromSize_t requires the GIL
+    if (_handle == nullptr) throw std::runtime_error("Invalid object or already released");
+    PyGILState_STATE state = PyGILState_Ensure();
+    ucxx::python::future_set_result_with_event_loop(
+      _asyncioEventLoop, _handle, _pythonConvert(result));
```
madsbk, line 54: Since `future_set_result_with_event_loop()` doesn't have a `noexcept`, we might not release the GIL on failure?

cpp/python/include/ucxx/python/python_future_task.h
```diff
@@ -0,0 +1,213 @@
...
+  ~PythonFutureTask() { Py_XDECREF(_handle); }
+
+  /**
+   * @brief Get the C++ future.
+   *
+   * Get the underlying C++ future that can be awaited and have its result read.
+   *
+   * @returns The underlying C++ future.
+   */
+  std::future<ReturnType>& getFuture() { return _future; }
```
madsbk, line 169: Since the other _getters_ are guarded against invalid object, we should do it here as well? 
```suggestion
  std::future<ReturnType>& getFuture() { 
    if (_handle == nullptr) throw std::runtime_error("Invalid object or already released");
    
    return _future; 
  }
```

cpp/python/src/future.cpp
```diff
@@ -167,6 +179,124 @@ PyObject* future_set_exception(PyObject* future, PyObject* exception, const char
...
+
+  if (init_ucxx_python() < 0) {
+    if (!PyErr_Occurred()) PyErr_SetString(PyExc_RuntimeError, "could not allocate internals.");
+    goto finish;
+  }
+
+  set_result_callable = PyObject_GetAttr(future, set_result_str);
+  if (PyErr_Occurred()) {
+    ucxx_trace_req("Error getting future `set_result` method.");
+    PyErr_Print();
```
madsbk, line 219: I think you missed a jump here?
```suggestion
    PyErr_Print();
    goto finish;
```

cpp/python/src/future.cpp
```diff
@@ -167,6 +179,124 @@ PyObject* future_set_exception(PyObject* future, PyObject* exception, const char
...
+
+  if (init_ucxx_python() < 0) {
+    if (!PyErr_Occurred()) PyErr_SetString(PyExc_RuntimeError, "could not allocate internals.");
+    goto finish;
+  }
+
+  set_exception_callable = PyObject_GetAttr(future, set_exception_str);
+  if (PyErr_Occurred()) {
+    ucxx_trace_req("Error getting future `set_exception` method.");
+    PyErr_Print();
```
madsbk, line 263:
```suggestion
    PyErr_Print();
    goto finish;
```

madsbk (changes requested):
Great work @pentschev ! 
I only have some few comments.

cpp/python/include/ucxx/python/python_future_task.h
```diff
@@ -0,0 +1,213 @@
...
+  PythonFutureTask& operator=(PythonFutureTask const&) = delete;
+  PythonFutureTask(PythonFutureTask&& o)               = delete;
+  PythonFutureTask& operator=(PythonFutureTask&& o)    = delete;
+
+  /**
+   * @brief Python future destructor.
+   *
+   * Decrement the reference count of the underlying Python future.
+   */
+  ~PythonFutureTask() { Py_XDECREF(_handle); }
```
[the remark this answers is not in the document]
pentschev: I guess it is safe as long as the current thread is able to take the GIL, i.e. the "main" thread releases the GIL at some point. However, that doesn't mean it couldn't cause deadlocks depending on how the application makes use of that. Definitely a Python cleanup routine could make things safer but I can't think of a cleanup routine that would not require some complex Python mechanism, meaning it can't be implemented here alone, perhaps you already have something in mind? The only alternative I can think of, which is definitely not a good one, would be to implement a public `release()` method that the user is supposed to call where it is known that taking the GIL is safe and then change the destructor to print a warning/error if `release()` wasn't called.

cpp/python/include/ucxx/python/python_future_task.h
pentschev: There are no C++ exceptions raised in `future_set_result_with_event_loop` (it is supposed to be a pure C function), so I think `noexcept` would be redundant, no?

cpp/python/include/ucxx/python/python_future_task.h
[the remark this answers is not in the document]
pentschev: Sorry, we already have a `release()` which has a different meaning (release the handle to the user). Replace `release()` in my comment above with `free()`.

cpp/python/include/ucxx/python/python_future_task.h
```diff
@@ -0,0 +1,213 @@
...
+        return r;
+      } catch (std::exception& e) {
+        this->setException(e);
+      }
+    });
+  }
+  PythonFutureTask(const PythonFutureTask&)            = delete;
+  PythonFutureTask& operator=(PythonFutureTask const&) = delete;
+  PythonFutureTask(PythonFutureTask&& o)               = delete;
+  PythonFutureTask& operator=(PythonFutureTask&& o)    = delete;
```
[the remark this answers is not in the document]
pentschev: This is a bit more complicated that I initially anticipated, https://github.com/rapidsai/ucxx/pull/46/files#diff-3a0bf4e47549f5e676a1a7e7808b95fb9d6b74ff625a8c13ae1b11c765f2966fR141-R151 has several references to `this` which do not get update after the move constructor/assignment operator gets called. I think for a proper move constructor we'll have to encapsulate `_task`/`_future` into another `struct`/`class` and make it a `std::unique_ptr` that we can move without moving the actual `_task/`_future`. Perhaps there's a smarter way to do it that I'm just not seeing yet.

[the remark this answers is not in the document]
pentschev: Went ahead and moved implementation details into a new class via https://github.com/rapidsai/ucxx/pull/46/commits/e4b928d8dc5203e13b1ab56fd5584a708b5ea387 and https://github.com/rapidsai/ucxx/pull/46/commits/24989febd928d2859c79c4058d4d7462f0ecaffa. This has also the benefit of not requiring us to define explicit move constructor/assignment operator.

cpp/python/include/ucxx/python/python_future_task.h
madsbk: I think `noexcept` would be good both for the sake of clarity and potential compiler optimizations.

cpp/python/include/ucxx/python/python_future_task.h
```diff
@@ -0,0 +1,293 @@
...
+  PythonFutureTask(PythonFutureTask&& o)               = default;
+  PythonFutureTask& operator=(PythonFutureTask&& o)    = default;
+
+  /**
+   * @brief Python future destructor.
+   *
+   * Decrement the reference count of the underlying Python future.
+   */
+  ~PythonFutureTask() {}
+
```
madsbk, line 251:
```suggestion
```

cpp/python/include/ucxx/python/python_future_task.h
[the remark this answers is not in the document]
madsbk: Do we need `_handle` after `setResult()` or `setPythonException()` has been called? If not, maybe let them release the handle?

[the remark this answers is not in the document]
pentschev: The `_handle` is the `asyncio.Future` object, so if the user never got a reference to it via `getHandle()`/`release()` before the C++ future completed and notified the Python future then the user will never be able to get it later if we release it in `setResult()`/`setPythonException()`. Therefore, I don't think this would work.

cpp/python/include/ucxx/python/python_future_task.h
pentschev: Just to be sure I'm not misunderstanding your suggestion, https://github.com/rapidsai/ucxx/pull/46/commits/9a7491f2e96d5bee7db70e5d9f125d4841104bce is what you mean, right? You're talking about C++ exceptions and not Python exceptions, the latter might occur if a Python error occurs, e.g. in

cpp/python/include/ucxx/python/python_future_task.h
pentschev: Done via https://github.com/rapidsai/ucxx/pull/46/commits/1e68b191b6e4132d8485977a449616929d48eb65 .

cpp/python/include/ucxx/python/python_future_task.h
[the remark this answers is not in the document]
madsbk: I see, that won't work :)
The only alternative to implementing a`free()` method, is to save the `_handle` in a global struct that we can clean up periodically.

E.g. if `PyGILState_Check() == 0`, we add `_handle` to a global struct that is checked in some of our API calls that grabs the GIL.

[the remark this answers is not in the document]
pentschev: Agreed, that's probably our best option. It will take me until next week or the week after to get back to this, but I'll try to make it work.

python/ucxx/examples/CPPLINT.cfg
wence-, line 1: Does this one also want to live at the top-level of the repo?

cpp/python/include/ucxx/python/python_future_task.h
wence-: wrt python exceptions, `_pythonConvert` (which takes the C++ result and produces a Python result) could produce an error (e.g. return `NULL` rather than a valid `PyObject *`) which it looks like we are not handling here?

python/ucxx/examples/.clang-format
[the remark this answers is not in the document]
pentschev: The idea is that this piece of code may not live here long-term, thus I'd prefer that this file remains here, as this is the only other place it's used besides `cpp/`, thus avoiding pollution of the root directory of the repository with files that are not necessarily relevant outside of some subdirectories.

python/ucxx/examples/python_future_task.h
```diff
@@ -0,0 +1,159 @@
...
+      ApplicationThread::processLoop(processingPool, readyPool);
+    }
+  }
+};
+
+class Application {
+ private:
+  std::unique_ptr<ApplicationThread> _thread{nullptr};  ///< The progress thread object
+  std::shared_ptr<std::mutex> _incomingPoolMutex{
+    std::make_shared<std::mutex>()};  ///< Mutex to access the Python futures pool
```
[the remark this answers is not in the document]
pentschev: This is simply a `shared_ptr` because until now we've followed `cpplint`s default based on Google's styleguide, which [disallows non-const references](https://github.com/cpplint/cpplint/issues/148), which is what `std::mutex` would need to be to allow passing it to static member methods (e.g., `submit()`). I don't mind if we want to disable that and then rewrite this, but I'd prefer to do that (discuss/disable non-const references prohibition) in another issue or PR. Do you have strong opinions about it? Would you mind starting a discussion on it if so?

python/ucxx/examples/python_future_task.h
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  static void progressUntilSync(PyObject* asyncioEventLoop,
+                                std::shared_ptr<std::mutex> incomingPoolMutex,
+                                FuturePoolPtr incomingPool,
+                                FuturePoolPtr readyPool,
+                                const bool& stop)
+  {
+    ucxx_warn("Application thread started");
+    auto processingPool = std::make_shared<FuturePool>();
```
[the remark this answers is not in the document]
pentschev: Same reason as for the `std::shared_ptr<std::mutex>` mentioned in https://github.com/rapidsai/ucxx/pull/46#discussion_r1402692013 .

python/ucxx/examples/python_future_task.h
```diff
@@ -0,0 +1,159 @@
...
+};
+
+class Application {
+ private:
+  std::unique_ptr<ApplicationThread> _thread{nullptr};  ///< The progress thread object
+  std::shared_ptr<std::mutex> _incomingPoolMutex{
+    std::make_shared<std::mutex>()};  ///< Mutex to access the Python futures pool
+  FuturePoolPtr _incomingPool{std::make_shared<FuturePool>()};  ///< Incoming task pool
+  FuturePoolPtr _readyPool{
+    std::make_shared<FuturePool>()};  ///< Ready task pool, only to ensure task lifetime
```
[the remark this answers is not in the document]
pentschev: It was only here to ensure task lifetime, although I can't remember if the code was different when I first introduced that actually made it a hard-requirement. I agree this is not needed currently, thus it's now removed via https://github.com/rapidsai/ucxx/pull/46/commits/514627d1c4b9bba36e1300e6df5d431125c91b9a .

python/ucxx/examples/CPPLINT.cfg
pentschev: For the same reason mentioned in https://github.com/rapidsai/ucxx/pull/46#discussion_r1402691611, I'd prefer to keep this copy here for the time being, until we decide what to do with all files related to this example.

python/ucxx/examples/.clang-format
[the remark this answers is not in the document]
wence-: Ok, this makes sense, thanks.

cpp/python/include/ucxx/python/python_future_task.h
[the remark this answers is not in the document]
pentschev: I have now [introduced a new singleton class `PythonFutureTaskCollector`](https://github.com/rapidsai/ucxx/pull/46/commits/72eb53938496b34db64f71e9d6ca151e339b6996) which allows pushing `PyObject*` when `~PythonFutureTask()` runs instead of it requiring the GIL to decref those. The GIL and decref now happens when `PythonFutureTaskCollector::collect()` gets called, but that needs to be done by the application itself, which [currently runs it during `~ApplicationThread`](https://github.com/rapidsai/ucxx/pull/46/commits/6e6314033b01e4efed1d395b5c0c5f0e1c1276e0).

This is not an optimal solution, but I think it's probably the only one that guarantees objects are actually decrefed at some point, the alternative may be to register another `atexit` handler. A periodic call of that method could also improve things a bit during runtime but is not enough to ensure everything gets cleaned up before exiting, as we may always miss the last few destroyed `PythonFutureTask` if the periodic call happens to be called one last time before some `~PythonFutureTask`. Perhaps I'm overlooking some obvious better and guaranteed way to ensure we don't leave any `PyObject*` behind, if you can think of any please let me know @madsbk @wence- .

cpp/python/include/ucxx/python/python_future_task.h
```diff
@@ -0,0 +1,290 @@
...
+    } catch (const std::overflow_error& e) {
+      setPythonException(PyExc_OverflowError, e.what());
+    } catch (const std::range_error& e) {
+      setPythonException(PyExc_ArithmeticError, e.what());
+    } catch (const std::underflow_error& e) {
+      setPythonException(PyExc_ArithmeticError, e.what());
+    } catch (const std::exception& e) {
+      setPythonException(PyExc_RuntimeError, e.what());
+    } catch (...) {
+      setPythonException(PyExc_RuntimeError, "Unknown exception");
```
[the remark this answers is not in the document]
wence-: Ah sorry, I didn't spot that there was no exception to inspect in this branch.

cpp/python/src/python_future_task_collector.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+  PyGILState_Release(state);
+}
+
+PythonFutureTaskCollector::PythonFutureTaskCollector() {}
+
+PythonFutureTaskCollector::~PythonFutureTaskCollector()
+{
+  if (_toCollect.size() > 0)
+    ucxx_warn("Destroying PythonFutureTaskCollector with %lu uncollected tasks", _toCollect.size());
```
wence-, line 49: **suggestion:** this should hold the lock to be race-free I think.

cpp/python/include/ucxx/python/python_future_task.h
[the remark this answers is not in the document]
wence-: This kind of deferred collection with stop-the-world cleanup points is the way I (with others) solve morally the same problem in a distributed setting (where two processes have to agree to destruct objects collectively). This is kind of similar because we're needing to do collection at a point when we can guarantee synchronisation of some resource (in this case the gil + any other locks that might be required).

So I think this is fine. You might imagine that you'd want to also call this "collect" periodically during submission of tasks if that happens on the main thread that will hold the gil.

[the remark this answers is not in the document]
pentschev: > So I think this is fine. You might imagine that you'd want to also call this "collect" periodically during submission of tasks if that happens on the main thread that will hold the gil.

I agree with that idea, what I'm not sure is who should be responsible for that. I can think of a few different approaches:

1. Within the `PythonFutureTask`, for example every time a new one is constructed;
2. As part of the `ApplicationThread`, perhaps during `progressUntilSync()`;
3. Directly from Python with something similar to [`tornado.ioloop.PeriodicCallback`](https://www.tornadoweb.org/en/stable/ioloop.html#tornado.ioloop.PeriodicCallback);
4. Directly from Python with looping forever submitting a new async task for every iteration, similar to [`PollingMode`](https://github.com/rapidsai/ucxx/blob/branch-0.36/python/ucxx/_lib_async/continuous_ucx_progress.py#L64-L72).

I think we should be doing one or more of those, but it's difficult to decide the best approach. Furthermore, `ApplicationThread` is meant to be an example only, how to actually do it probably depends on the actual application needs so we don't need to cover every possible alternative. We will need to find an appropriate solution to this for UCXX though, but that will happen in a follow-up PR.

cpp/python/src/python_future_task_collector.cpp
pentschev: Done via https://github.com/rapidsai/ucxx/pull/46/commits/7e0ad1b7de36915eb3edf78898e1208e5eebcca6 .

python/ucxx/examples/python_future_task.h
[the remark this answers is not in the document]
wence-: Ah, ok. Nothing to see here then, I will collect my thoughts.

wence- (approved):
I think this looks good, @madsbk do you have any further thoughts/comments.

pentschev:
Thanks @madsbk and @wence- !

python/CMakeLists.txt
```diff
@@ -48,8 +48,10 @@ endif()
 
 rapids_cython_init()
 
+add_subdirectory(ucxx/examples)
 add_subdirectory(ucxx/_lib)
 
 if(DEFINED cython_lib_dir)
   rapids_cython_add_rpath_entries(TARGET ucxx PATHS "${cython_lib_dir}")
+  rapids_cython_add_rpath_entries(TARGET ucxx_examples PATHS "${cython_lib_dir}")
```
vyasr, line 56: Shouldn't need this.
```suggestion
```

python/ucxx/examples/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+# SPDX-License-Identifier: BSD 3-Clause License
+# =================================================================================
+
+set(cython_sources python_future_task_app.pyx)
+set(linked_libraries ucxx::ucxx ucxx::python Python3::Python)
+
+rapids_cython_create_modules(
+  CXX
+  SOURCE_FILES "${cython_sources}"
+  LINKED_LIBRARIES "${linked_libraries}" ASSOCIATED_TARGETS ucxx_examples
```
vyasr, line 12: I'm a bit confused, what is the ucxx_examples target? I see a custom `run_examples` target created in the cpp/examples/CMakeLists.txt, is that what you were trying to refer to? Does this Python application need to link to something created in that example?

python/CMakeLists.txt
pentschev: This is needed for a proper build/install of https://github.com/rapidsai/ucxx/pull/46/files/7e0ad1b7de36915eb3edf78898e1208e5eebcca6#diff-0cdf2e3fa52349932e0e3d2f84da5e2c9d3c664140dbe62c74db204c97ef4807.

python/ucxx/examples/CMakeLists.txt
pentschev: No, this refers to https://github.com/rapidsai/ucxx/pull/46/files/7e0ad1b7de36915eb3edf78898e1208e5eebcca6#diff-691a68ea554b5fd97545a2986bb6b915ecb28e20ea05ec5cb95b8b94340a57be, see the [`cython_sources`](https://github.com/rapidsai/ucxx/pull/46/files/7e0ad1b7de36915eb3edf78898e1208e5eebcca6#diff-0cdf2e3fa52349932e0e3d2f84da5e2c9d3c664140dbe62c74db204c97ef4807R6).

This is a Cython file that [wraps some C++ code](https://github.com/rapidsai/ucxx/pull/46/files/7e0ad1b7de36915eb3edf78898e1208e5eebcca6#diff-774dabed7ea310b8dfa9acde01482c4ce18a255ee4229f59e7c3b47d9a342446) which demonstrates how to set the result/exception of a Python future from a C++ future.

python/CMakeLists.txt
pentschev: I think you're actually right here, tested locally without it and seemed to build correctly. I'll try removing it and see if CI passes too.

python/ucxx/examples/CMakeLists.txt
vyasr: To clarify, I'm just confused about the `ASSOCIATED_TARGETS` argument here. I think that can be removed just like you removed the `rapids_cython_add_rpath_entries` above. However, if the example requires ucxx you may need to have ucxx rather than ucxx_examples here.

pentschev: Got it, it seems you're right and we don't need it. Removed it in https://github.com/rapidsai/ucxx/pull/46/commits/e55efdd7716786d8584e0cebd1ede9ca59091481 .

pentschev: Sorry, linked the wrong commit above, remove it in https://github.com/rapidsai/ucxx/pull/46/commits/8439f9527c38cd77d7ed425b5419b12a08de728d .

PR #47, opened 2023-05-08
Fix enabling delayed submission and Python future

pentschev:
Some issues were accidentally introduced when the C++ files were moved to `cpp/python`, changes contained here fix those issues and add C++ and Python tests to check that the correct UCXX worker capabilities are enabled.


pentschev:
Thanks @wence- ! 🙂

PR #48, opened 2023-05-19
Add Active Messages support

pentschev:
Introduce new `ucxx::RequestAm` class to allow transferring via the Active Messages UCX API.


cpp/include/ucxx/worker.h
```diff
@@ -643,6 +674,56 @@ class Worker : public Component {
...
+   *
+   * @code{.cpp}
+   * // context is `std::shared_ptr<ucxx::Context>`
+   * auto worker = context->createWorker(false);
+   *
+   * worker->registerAmAllocator(`UCS_MEMORY_TYPE_CUDA`, ucxx::RMMBuffer);
+   * @endcode
+   *
+   * @param[in] memoryType  the memory type the allocator will be used for.
+   * @param[in] allocator   the allocator callbable that will be used to allocate new
```
wence-, line 702:
```suggestion
   * @param[in] allocator   the allocator callable that will be used to allocate new
```

cpp/src/request.cpp
```diff
@@ -69,7 +69,7 @@ void Request::cancel()
                        ucs_status_string(status));
     } else {
       ucxx_trace_req_f(_ownerString.c_str(), _request, _operationName.c_str(), "canceling");
-      ucp_request_cancel(_worker->getHandle(), _request);
+      if (_request != nullptr) ucp_request_cancel(_worker->getHandle(), _request);
```
wence-, line 72: This is unrelated to the AM changes and just a check we need to add, right?

cpp/include/ucxx/worker.h
```diff
@@ -643,6 +674,56 @@ class Worker : public Component {
...
+   * on the host allocator is undesirable, for example when transferring CUDA buffers the
+   * destination is always going to be a host buffer and prevent the use of transports such
+   * as NVLink or InfiniBand+GPUDirectRDMA. For that reason it's important that the user
+   * defines those allocators that are important for the application.
+   *
+   * If the `memoryType` has already been registered, the previous allocator will be
+   * replaced by the new one. Be careful when doing this after transfers have started, there
+   * are no guarantees that inflight messages have not already been allocated with the old
+   * allocator for that type.
+   *
```
wence-, line 693: Do we want to add some text here about how the data received by active messages must be deallocated using the allocator's "matching" `free` implementation? E.g. I see that the `HostBuffer` implementation uses `malloc`/`free` pairs, but it could equally use `new`/`delete`. (This is only _really_ relevant when doing `HostBuffer.release` I guess).

wence- (commented):
A partial look, I haven't gone through the logical of the implementation properly yet.

cpp/include/ucxx/worker.h
pentschev: I haven't thought about this, but given the object the user will ultimately get via `request->getRecvBuffer()` is a `Buffer`, I would hope [the user reads `Buffer::release()` docs](https://github.com/rapidsai/ucxx/blob/95f0e4280488daec3ecdd1870adba587ecea1f83/cpp/include/ucxx/buffer.h#L122-L125) which writes that. Would you like to be redundant here? To me the danger of redundancy is in the event of changes in the behavior of `Buffer` and missing to update the relevant bits here too, perhaps the ideal case would be to have a cross-ref instead, but I haven't done that in this repo yet.

cpp/include/ucxx/worker.h
pentschev: Fixed in https://github.com/rapidsai/ucxx/pull/48/commits/bbf6e27c014efea57e6cb0484d3b663dbb1e258e

cpp/src/request.cpp
pentschev: That's right, I found it during development here and decided to throw it in here instead of opening a new PR, although I'll probably add this to another PR too to get this change in sooner.

pentschev (commented):
Thanks @wence- for looking at this, apologies for not addressing it earlier but somehow I completely missed the notification for this PR. I've addressed/responded to your comments now, please have a look when you've got the chance.

cpp/src/internal/request_am.cpp
```diff
@@ -0,0 +1,38 @@
...
+namespace internal {
+
+RecvAmMessage::RecvAmMessage(internal::AmData* amData,
+                             ucp_ep_h ep,
+                             std::shared_ptr<RequestAm> request,
+                             std::shared_ptr<Buffer> buffer)
+  : _amData(amData), _ep(ep), _request(request), _buffer(buffer)
+{
+  _request->_delayedSubmission =
+    std::make_shared<DelayedSubmission>(false, _buffer->data(), _buffer->getSize());
```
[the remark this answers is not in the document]
wence-: Also, the lifetime management of this object (and specifically in relation to that of `_buffer`) looks potentially bad.

The `DelayedSubmission` doesn't incref the buffer shared pointer, so if there's a way to grab the delayed submission out of a `Request` then it could have a longer lifetime than the request (and hence the buffer).

e.g.
```c++
std::shared_ptr<DelayedSubmission> long_lived = nullptr;
{
   auto buffer = Buffer...;
   auto request = Request(..., buffer);
   msg = RecvAmMsg(..., request, buffer);
   long_lived = msg->delayedSubmission;
}
long_lived->data...; // use-after-free
```

cpp/src/internal/request_am.cpp
```diff
@@ -0,0 +1,38 @@
...
+}
+
+void RecvAmMessage::setUcpRequest(void* request) { _request->_request = request; }
+
+void RecvAmMessage::callback(void* request, ucs_status_t status)
+{
+  _request->_buffer = _buffer;
```
wence-, line 28: Can `_request` be used for something else during the lifetime of `this`? if not, why not set up the `_request->_buffer` pointer in the constructor of `RecvAmMessage`, if yes, then this seems potentially racy.

cpp/src/endpoint.cpp
```diff
@@ -191,6 +192,27 @@ void Endpoint::removeInflightRequest(const Request* const request)
...
 size_t Endpoint::cancelInflightRequests() { return _inflightRequests->cancelAll(); }
 
+std::shared_ptr<Request> Endpoint::amSend(void* buffer,
+                                          size_t length,
+                                          ucs_memory_type_t memoryType,
+                                          const bool enablePythonFuture,
+                                          RequestCallbackUserFunction callbackFunction,
+                                          RequestCallbackUserData callbackData)
+{
+  auto endpoint = std::dynamic_pointer_cast<Endpoint>(shared_from_this());
```
wence-, line 202: Why do we need to cast `this` (which is an `Endpoint`?) to `Endpoint` (or is it that it might be a subclass?)

cpp/src/request.cpp
```diff
@@ -105,10 +105,11 @@ void Request::callback(void* request, ucs_status_t status)
                    _callback.target<void (*)(void)>());
   if (_callback) _callback(status, _callbackData);
 
-  ucp_request_free(request);
+  if (request != nullptr) ucp_request_free(request);
```
wence-, line 108: TODO: request that UCX accepts `nullptr` in its `XXX_free` functions so that we don't have to litter user code with this kind of nonsense error checking. That would align the UCX APIs more closely with that of `free`.

cpp/src/worker.cpp
```diff
@@ -374,4 +428,16 @@ std::shared_ptr<Listener> Worker::createListener(uint16_t port,
   return listener;
 }
 
+void Worker::registerAmAllocator(ucs_memory_type_t memoryType, AmAllocatorType allocator)
+{
+  if (_amData == nullptr)
+    throw std::runtime_error("Active Messages wasn not enabled during context creation");
+  _amData->_allocators.insert_or_assign(memoryType, allocator);
```
wence-, line 435: FWIW, since `ucs_memory_type_t` is an enum, the `_allocators` object could just be a `std::array<AmAllocatorType, UCS_MEMORY_TYPE_LAST + 1>` and accessing it could be `_allocators[memoryType]`.

python/ucxx/_lib/libucxx.pyx
```diff
@@ -951,6 +979,50 @@ cdef class UCXEndpoint():
...
+
+        with nogil:
+            req = self._endpoint.get().amSend(
+                buf,
+                nbytes,
+                UCS_MEMORY_TYPE_CUDA if cuda_array else UCS_MEMORY_TYPE_HOST,
+                self._enable_python_future
+            )
+
+        return UCXRequest(<uintptr_t><void*>&req, self._enable_python_future)
```
wence-, line 1013: Really the address of the shared pointer, and not the shared pointer itself?

python/ucxx/_lib/libucxx.pyx
```diff
@@ -951,6 +979,50 @@ cdef class UCXEndpoint():
...
+    def am_recv(self):
+        cdef shared_ptr[Request] req
+
+        if not self._context_feature_flags & Feature.AM.value:
+            raise ValueError("UCXContext must be created with `Feature.AM`")
+
+        with nogil:
+            req = self._endpoint.get().amRecv(self._enable_python_future)
+
+        return UCXRequest(<uintptr_t><void*>&req, self._enable_python_future)
```
wence-, line 1024: Same query here about `address-of-pointer` vs `pointer`.

python/ucxx/_lib_async/endpoint.py
```diff
@@ -85,6 +85,43 @@ async def close(self):
...
+        # Optimization to eliminate producing logger string overhead
+        if logger.isEnabledFor(logging.DEBUG):
+            nbytes = buffer.nbytes
+            log = "[AM Send #%03d] ep: %s, nbytes: %d, type: %s" % (
+                self._send_count,
+                hex(self.uid),
+                nbytes,
+                type(buffer.obj),
+            )
+            logger.debug(log)
```
wence-, line 112: This can be a little simpler since the logging module does at least accept format strings and arguments in its calls.

So

```suggestion
            logger.debug(
                "[AM Send #%03d] ep: 0x%x, nbytes: %d, type: %s", 
                self._send_count,
                self.uid,
                nbytes,
                type(buffer.obj),
            )
```

python/ucxx/_lib_async/endpoint.py
```diff
@@ -214,6 +249,44 @@ async def send_obj(self, obj, tag=None):
...
+            if self.closed():
+                raise UCXCloseError("Endpoint closed")
+
+        # Optimization to eliminate producing logger string overhead
+        if logger.isEnabledFor(logging.DEBUG):
+            log = "[AM Recv #%03d] ep: %s" % (
+                self._recv_count,
+                hex(self.uid),
+            )
+            logger.debug(log)
```
wence-, line 265: Same suggestion here:
```suggestion
            logger.debug(
                "[AM Recv #%03d] ep: 0x%x",
                self._recv_count,
                self.uid,
            )
```

python/ucxx/_lib_async/endpoint.py
```diff
@@ -214,6 +249,44 @@ async def send_obj(self, obj, tag=None):
...
+        buffer = req.get_recv_buffer()
+
+        if logger.isEnabledFor(logging.DEBUG):
+            log = "[AM Recv Completed #%03d] ep: %s, nbytes: %d, type: %s" % (
+                self._recv_count,
+                hex(self.uid),
+                buffer.nbytes,
+                type(buffer),
+            )
+            logger.debug(log)
```
wence-, line 280: And here... (Note also using `"0x%x"` to format the uid as hex, rather than calling `hex(self.uid)`).

wence- (commented):
I think there's a little cleanup we can do in the recvWait/recvPool mapping handling, and then some other minor cleanups/queries in the logic side of things.

python/ucxx/_lib_async/endpoint.py
pentschev: Done in https://github.com/rapidsai/ucxx/pull/48/commits/bd1787f0b17d505f6e4ec60037fc605d057bcd01

python/ucxx/_lib_async/endpoint.py
pentschev: Done in https://github.com/rapidsai/ucxx/pull/48/commits/bd1787f0b17d505f6e4ec60037fc605d057bcd01

python/ucxx/_lib_async/endpoint.py
pentschev: Done in https://github.com/rapidsai/ucxx/pull/48/commits/bd1787f0b17d505f6e4ec60037fc605d057bcd01

pentschev (commented):
Thanks @wence- for another great, thorough review. I've addressed your comments, could you please take another look when you have a chance?

wence- (approved):
Thanks for all the explanation, looks good I think.

cpp/src/internal/request_am.cpp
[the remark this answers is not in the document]
pentschev: You're right that the `shared_ptr` doesn't extend the lifetime of their owners, but the design we have in UCXX is different than the sample you wrote. In UCXX the user is responsible for guaranteeing the lifetime of (the equivalent of) `Child`, and `Child` holds a `shared_ptr` to (the equivalent of) `Parent`, thus ensuring `shared_ptr<Parent>`'s lifetime. What we have in UCXX is analogous to the following:

```c++
#include <memory>
#include <iostream>

struct Child;

struct Parent {
  
  Parent() {
    std::cout << "cons Parent " << std::endl;
  };
  ~Parent() {
    std::cout << "dest Parent " << std::endl;
  }

  
};

struct Child {

  std::shared_ptr<Parent> _p;

  Child() {
    std::cout << "cons Child" << std::endl;
  };
  ~Child() {
    std::cout << "dest Child" << std::endl;
  }

  void setParent(std::shared_ptr<Parent> p)
  {
    _p = p;
  }
};

int main(void)
{
  auto c = std::make_shared<Child>();
  std::shared_ptr<Child> d = nullptr;
  {
    auto p = std::make_shared<Parent>();
    c->setParent(p);
    d = c;
  }

  return 0;
}
```

Regarding the original point regarding `DelayedSubmission`, you are right. I missed making it a `ucxx::Component`, that's what we need to ensure its lifetime, thanks for catching that! The fix for this is now done via https://github.com/rapidsai/ucxx/pull/48/commits/d7dcea00b25a3ce04b0f389f5ec633b43e11303a.

cpp/src/internal/request_am.cpp
pentschev: That's a good idea, done in https://github.com/rapidsai/ucxx/pull/48/commits/6eb5f95b48c9f407d32a9a29267082ecaf1884bd.

cpp/src/worker.cpp
pentschev: Given UCX leaves this for the user to write any handler, the result type could be anything, even something that UCX doesn't support. One could receive on a CUDA buffer and move that data to a host buffer before returning it to the user, for example. Another possibility would be some kind of composition that returns multiple buffers (we may want to explore a multi-buffer transfer perhaps, similar to `RequestTagMulti`). In any of those cases we will probably want a non-`ucs_memory_type_t` assuming we make this more flexible in the future and allow the user to do any arbitrary transfers.

pentschev (commented):
Thanks @wence- for looking at it again. I think I've addressed all your comments, but please let me know if I missed something.

wence- (approved):

wence-:
Thanks, not sure what went on in the tests though

robertmaynard (approved):

cpp/src/internal/request_am.cpp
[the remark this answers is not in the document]
pentschev: It turns out that the change in https://github.com/rapidsai/ucxx/commit/d7dcea00b25a3ce04b0f389f5ec633b43e11303a is problematic, it creates a dependency cycle as `std::shared_ptr<ucxx::Request>` has a `shared_ptr<DelayedSubmission>`, and the latter has now also a reference to the former, thus preventing those objects from being destroyed. Thus, this has to be fixed in a different way: the `std::shared_ptr<ucxx::Request>` should continue to effectively own `shared_ptr<DelayedSubmission>`, but we must make sure that a reference to `std::shared_ptr<ucxx::Request>` is kept by the worker until it processes `shared_ptr<DelayedSubmission>`, only then the request should be allowed to go out-of-scope. Therefore, I reverted the first commit in favor of https://github.com/rapidsai/ucxx/pull/48/commits/88e8b349406468bda9d68aae893d430f72e75c4e. Would you mind taking a look at it @wence- and letting me know if you think that's reasonable?

cpp/src/delayed_submission.cpp
```diff
@@ -35,26 +35,25 @@ void DelayedSubmissionCollection::process()
...
-void DelayedSubmissionCollection::registerRequest(DelayedSubmissionCallbackType callback)
+void DelayedSubmissionCollection::registerRequest(std::shared_ptr<Request> request,
+                                                  DelayedSubmissionCallbackType callback)
 {
-  auto r = std::make_shared<DelayedSubmissionCallbackType>(callback);
-
   {
     std::lock_guard<std::mutex> lock(_mutex);
-    _collection.push_back(r);
+    _collection.push_back(std::make_pair(request, callback));
```
wence-, line 54: I think you can use an initializer list here to make it easier to read:
```suggestion
    _collection.push_back({request, callback});
```

cpp/src/request_stream.cpp
```diff
@@ -41,7 +34,12 @@ std::shared_ptr<RequestStream> createRequestStream(std::shared_ptr<Endpoint> end
   auto req = std::shared_ptr<RequestStream>(
     new RequestStream(endpoint, send, buffer, length, enablePythonFuture));
 
-  req->_delayedSubmission->setParent(req);
+  // A delayed notification request is not populated immediately, instead it is
+  // delayed to allow the worker progress thread to set its status, and more
+  // importantly the Python future later on, so that we don't need the GIL here.
+  req->_worker->registerDelayedSubmission(
+    req, std::bind(std::mem_fn(&Request::populateDelayedSubmission), req.get()));
```
wence-, line 41: Ah, this needs to be moved from the constructor because we need the `shared_ptr` to `this` which isn't available.

cpp/src/delayed_submission.cpp
```diff
@@ -35,26 +35,25 @@ void DelayedSubmissionCollection::process()
       toProcess = std::move(_collection);
```
wence-, line 35: What does `std::move` do to `_collection`? Does it just drain the vector, leaving it "as-if" it were empty again? I guess so.

wence- (approved):
A very minor comment.

pentschev:
All tests pass now, I'll leave this to be merged Monday in case you have any last comments @wence- , thanks again for reviewing!

PR #50, opened 2023-05-24
Various minor improvements

pentschev:
Adds various minor improvements:

- Add option to create `compile_commands.json` in build scripts
- Make log level string names consistent with UCX
- Simplify `ucxx::Endpoint::getWorker()` by making it a non-static method that doesn't require an argument


build.sh
```diff
@@ -180,6 +186,10 @@ if buildAll || hasArg libucxx; then
     compile_end=$(date +%s)
     compile_total=$(( compile_end - compile_start ))
 
+    if [[ ${BUILD_COMPILE_COMMANDS} == "ON" ]]; then
+      cp compile_commands.json ..
```
wence-, line 190: If the build directory is `cpp/build` then clangd automatically finds `compile_commands.json`, so there's not necessarily any need to do this copy.

build.sh
```diff
@@ -18,7 +18,7 @@ ARGS=$*
 # script, and that this script resides in the repo dir!
 REPODIR=$(cd $(dirname $0); pwd)
 
-VALIDARGS="clean libucxx libucxx_python ucxx benchmarks tests examples -v -g -n -l --show_depr_warn -h"
+VALIDARGS="clean libucxx libucxx_python ucxx benchmarks tests examples -v -g -n -c --show_depr_warn -h"
 HELP="$0 [clean] [libucxx] [libucxx_python] [ucxx] [benchmarks] [tests] [examples] [-v] [-g] [-n] [-h] [--cmake-args=\\\"<args>\\\"]
```
wence-, line 22:
```suggestion
HELP="$0 [clean] [libucxx] [libucxx_python] [ucxx] [benchmarks] [tests] [examples] [-vcgnh] [--cmake-args=\\\"<args>\\\"]
```

?

build.sh
```diff
@@ -18,7 +18,7 @@ ARGS=$*
 # script, and that this script resides in the repo dir!
 REPODIR=$(cd $(dirname $0); pwd)
 
-VALIDARGS="clean libucxx libucxx_python ucxx benchmarks tests examples -v -g -n -l --show_depr_warn -h"
```
wence-, line 21: Presume that `-l` never did anything?

cpp/include/ucxx/endpoint.h
```diff
@@ -421,20 +421,16 @@ class Endpoint : public Component {
...
+   * A `std::shared_ptr<ucxx::Endpoint>` needs to be created and registered by
+   * `std::shared_ptr<ucxx::Worker>`, but the endpoint may be a child of a
+   * `std::shared_ptr<ucxx::Listener>` object. For convenience, this method can be used to
+   * get the `std::shared_ptr<ucxx::Worker>` which the endpoint is associated with.
    *
-   * @param[in] workerOrListener the `std::shared_ptr<ucxx::Worker>` or
-   *                             `std::shared_ptr<ucxx::Listener>` object.
-   *
-   * @returns The `std::shared_ptr<ucxx::Worker>` derived from `workerOrListener` argument.
+   * @returns The `std::shared_ptr<ucxx::Worker>` which the endpoint is assocaited with.
```
wence-, line 431:
```suggestion
   * @returns The `std::shared_ptr<ucxx::Worker>` which the endpoint is associated with.
```

cpp/src/endpoint.cpp
```diff
@@ -23,12 +23,22 @@
 
 namespace ucxx {
 
+static std::shared_ptr<Worker> getWorker(std::shared_ptr<Component> workerOrListener)
+{
+  auto worker = std::dynamic_pointer_cast<Worker>(workerOrListener);
+  if (worker == nullptr) {
+    auto listener = std::dynamic_pointer_cast<Listener>(workerOrListener);
```
wence-, line 30: The type system will not save you if you pass a non worker-or-listener in here and you'll get some kind of bad de-reference. Should we do `if (listener == nullptr) { throw ...; }` ?

cpp/src/log.cpp
```diff
@@ -26,7 +26,7 @@ void parseLogLevel()
       });
 
     auto level = logLevelNames.find(logLevelName);
-    if (level != logLevelNames.end())
+    if (!logLevelName.empty() > 0 && level != logLevelNames.end())
```
wence-, line 29: `logLevelName.empty()` returns bool, why the check `> 0` ?

Is this just:
```
if (!logLevelName.empty() && level != logLevelNames.end())
```
?

cpp/include/ucxx/log.h
```diff
@@ -69,11 +69,12 @@ const std::unordered_map<std::string, ucxx_log_level_t> logLevelNames = {
...
-  {"TRACE_REQ", UCXX_LOG_LEVEL_TRACE_REQ},
-  {"TRACE_DATA", UCXX_LOG_LEVEL_TRACE_DATA},
-  {"TRACE_ASYNC", UCXX_LOG_LEVEL_TRACE_ASYNC},
-  {"TRACE_FUNC", UCXX_LOG_LEVEL_TRACE_FUNC},
-  {"TRACE_POLL", UCXX_LOG_LEVEL_TRACE_POLL},
+  {"REQ", UCXX_LOG_LEVEL_TRACE_REQ},
+  {"DATA", UCXX_LOG_LEVEL_TRACE_DATA},
+  {"ASYNC", UCXX_LOG_LEVEL_TRACE_ASYNC},
+  {"FUNC", UCXX_LOG_LEVEL_TRACE_FUNC},
+  {"POLL", UCXX_LOG_LEVEL_TRACE_POLL},
```
wence-, line 76: I think we should keep these names the same as UCX, WDYT?

cpp/include/ucxx/log.h
```diff
@@ -69,11 +69,12 @@ const std::unordered_map<std::string, ucxx_log_level_t> logLevelNames = {
...
+  {"", UCXX_LOG_LEVEL_LAST},
   {"PRINT", UCXX_LOG_LEVEL_PRINT}};
```
wence-, line 78: Should these be in the opposite order?

cpp/src/endpoint.cpp
wence-: I guess one can't use `std::variant` because the type of `parent` is just `Component`.

wence- (changes requested):
Minor.

cpp/include/ucxx/log.h
pentschev: No, the [print level](https://github.com/openucx/ucx/blob/662455a15c4befc26b2baee45f1fc3b9cd6661b1/src/ucs/debug/log.c#L75) is a sort of "temporary debug", which you can use for development debugging. I'm not even sure we can use it currently because [we're bound to UCX logger's internals checking for the highest level at compile-time](https://github.com/openucx/ucx/blob/662455a15c4befc26b2baee45f1fc3b9cd6661b1/src/ucs/debug/log.c#L288-L291), but I'm leaving it here for the time being as I would like to fix that in the future.

build.sh
pentschev: Hmm, at least for me clangd via [neovim+LSP](https://github.com/neovim/nvim-lspconfig) hasn't been finding it in `cpp/build`, it's possible I missed some config but couldn't get it to work oterhwise so far.

cpp/src/log.cpp
pentschev: Yes, thanks for catching, fixed in https://github.com/rapidsai/ucxx/pull/50/commits/a811406d5af37375c0d64c06f5c0dc80e0a81644. I remember when I was writing this I was doing it the Python way, `logLevelName.size() > 0`, then I remembered of `.empty()` but forgot to remove the comparison operator.

cpp/src/endpoint.cpp
pentschev: Good catch, fixed in https://github.com/rapidsai/ucxx/pull/50/commits/0a876bf40d11fda391d1b5f8708dabb2e6174212. I think your comment about `std::variant` is also right, it would be useful if we were passing the parent when it was already cast into its final form, but that is unfortunately not the case here.

pentschev (commented):
Thanks @wence- for reviewing. I think I addressed all your comments, would you mind taking another look when you have the chance?

pentschev:
Any more comments here @wence- ?

build.sh
wence-: Ah, no worries.

cpp/include/ucxx/log.h
wence-: Ah, I had misremembered, nothing to see here...

wence- (approved):
Looks good, thanks!

pentschev:
Thanks @wence- !

PR #54, opened 2023-06-06
Fix RMM dependency

bdice:
The `dependencies.yaml` file lists `librmm` for conda/pyproject/requirements, but the C++ package librmm isn't available for pyproject/requirements files. Only the Python package `rmm` is available as a wheel. This PR swaps the two dependencies in the list.

Also I updated `gmock` / `gtest` to align with the versions used by the rest of RAPIDS and sorted the dependency list.


pentschev (approved):
LGTM, thanks @bdice

PR #56, opened 2023-06-09
Update C++/Python version to 0.33.00 and remove `_version.py`

pentschev:
The C++ and Python distribution files were overlooked and were versioned as `0.0.1`, even though the conda packages already follow the same version as UCX-Py. This change fixes that problem, updates the release script so that versions will be properly updated in the future and remove legacy `_version.py` file.


wence- (approved):
Thanks Peter!

robertmaynard (approved):

raydouglass (approved):

PR #58, opened 2023-06-09
Fixes for various tests intermittently failing in CI

pentschev:
Changes included are:

- Use different ports for C++ benchmarks/examples in CI
- Set `UCX_KEEPALIVE_INTERVAL` for Python benchmarks in CI
- Wait Python async AM test message to arrive
- Move `ucxx::Request` status setting to base class

### DIFF
--- a/cpp/src/request.cpp
+++ b/cpp/src/request.cpp
@@ -169,6 +169,7 @@ void Request::setStatus(ucs_status_t status)
                    status,
                    ucs_status_string(status));
 
+  if (_status != UCS_INPROGRESS) ucxx_error("setStatus called but the status was already set");
   _status.store(status);
 
   if (_enablePythonFuture) {
--- a/cpp/src/request.cpp
+++ b/cpp/src/request.cpp
@@ -162,20 +162,14 @@ void Request::setStatus(ucs_status_t status)
   if (_endpoint != nullptr) _endpoint->removeInflightRequest(this);
   _worker->removeInflightRequest(this);
 
-  if (_status == UCS_INPROGRESS) {
-    // If the status is not `UCS_INPROGRESS`, the derived class has already set the
-    // status, a truncated message for example.
-    _status.store(status);
-  }
-
-  ucs_status_t s = _status;
-
   ucxx_trace_req_f(_ownerString.c_str(),
                    _request,
                    _operationName.c_str(),
                    "callback called with status %d (%s)",
-                   s,
-                   ucs_status_string(s));
+                   status,
+                   ucs_status_string(status));
+
+  _status.store(status);
 
   if (_enablePythonFuture) {
     auto future = std::static_pointer_cast<ucxx::Future>(_future);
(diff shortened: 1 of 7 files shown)


cpp/src/request.cpp
```diff
@@ -162,20 +162,14 @@ void Request::setStatus(ucs_status_t status)
...
   ucxx_trace_req_f(_ownerString.c_str(),
                    _request,
                    _operationName.c_str(),
                    "callback called with status %d (%s)",
-                   s,
-                   ucs_status_string(s));
+                   status,
+                   ucs_status_string(status));
+
+  _status.store(status);
```
wence-, line 172: OK so now we just always store the status (even if it is not `INPROGRESS`). Because the derived class might not set the status any more? Or, I guess, it is safer this way.

cpp/src/request_tag.cpp
```diff
@@ -70,8 +70,6 @@ void RequestTag::callback(void* request, ucs_status_t status, const ucp_tag_recv
     std::snprintf(_status_msg.data(), _status_msg.size(), fmt, info->length, _length);
   }
 
-  _status = status;
-
   Request::callback(request, status);
```
wence-, line 73: So we're just deferring to the superclass to set the status now?

python/ucxx/_lib_async/tests/test_send_recv_am.py
```diff
@@ -83,13 +83,17 @@ async def test_send_recv_am(size, recv_wait, data):
...
 
     if data["memory_type"] == "cuda" and msg.nbytes < rndv_thresh:
         # Eager messages are always received on the host, if no custom host
         # allocator is registered, UCXX defaults to `np.array`.
         np.testing.assert_equal(recv[0].view(np.int64), msg.get())
     else:
         data["validator"](recv[0], msg)
+
+    for c in clients:
+        await c.close()
```
wence-, line 98: Is it better to `await asyncio.gather(*(c.close() for c in clients))` ?

wence- (approved):
Thanks @pentschev nothing to adapt here I think, I just had a few small questions.

cpp/src/request.cpp
pentschev: It is safer this way, I wrote the reason in https://github.com/rapidsai/ucxx/pull/58/commits/874b4c377c8d5d00b69800e374623d171141593b but it's indeed not clear without it. Also, TBH, I didn't account for the case where this isn't `UCS_INPROGRESS` as that is not expected to happen, but I've logged an error now in https://github.com/rapidsai/ucxx/pull/58/commits/6f99cbbd436b558ee1c3cc3ec7d8d972e9d9a935 to catch this case should it happen.

cpp/src/request_tag.cpp
pentschev: Correct, I've commented in the question above regarding the reasons.

python/ucxx/_lib_async/tests/test_send_recv_am.py
pentschev: Should be equivalent, but I agree your suggestion is a more asyncio-nic solution, done in https://github.com/rapidsai/ucxx/pull/58/commits/4f3d5d819fefdbd25d857ac5c02547b55d41c750.

pentschev (commented):
Thanks @wence- for the review, I've responded and addressed your comments.

PR #60, opened 2023-06-14
Add distributed-ucxx subproject

pentschev:
Add new subproject `distributed-ucxx`, providing a plugin for Distributed with a new `protocol="ucxx"` that can be specified by the user to enable UCXX as backend for communication. This is completely independent of UCX-Py for now, which may still be chosen by specifying `protocol="ucx"`.

Most of the changes here are actually reimplementing the [UCX-Py backend from Distributed](https://github.com/dask/distributed/blob/main/distributed/comm/ucx.py), with minor changes such as `ucp`->`ucxx` and to adapt to API changes in UCXX. The tests in this PR are also those that currently test [UCX-Py in Distributed](https://github.com/dask/distributed/tree/main/distributed/comm/tests), similarly with `ucp`->`ucxx` and API adaptations.

Packaging and distribution may still require further work that will be addressed in follow-up PRs.


python/distributed-ucxx/pyproject.toml
```diff
@@ -0,0 +1,117 @@
...
+    "tests.*",
+]
+
+[tool.versioneer]
+VCS = "git"
+style = "pep440"
+versionfile_source = "distributed_ucxx/_version.py"
+versionfile_build = "distributed_ucxx/_version.py"
+tag_prefix = "v"
+parentdir_prefix = "distributed_ucxx-"
```
wence-, line 117: Should we use the same approach as elsewhere and _not_ introduce versioneer? Which I think we've migrated away from.

python/distributed-ucxx/distributed_ucxx/distributed_patches.py
```diff
@@ -0,0 +1,41 @@
...
+
+import ucxx
+
+_scheduler_close = Scheduler.close
+_worker_close = Worker.close
+
+
+def _stop_notifier_thread_and_progress_tasks():
+    ucxx.stop_notifier_thread()
+    ucxx.core._get_ctx().progress_tasks.clear()
```
wence-, line 12: This presumably means we can't have multiple schedulers running simultaneously (probably this is a dask restriction too). Since the context is global and not tied to a scheduler.

python/distributed-ucxx/distributed_ucxx/tests/test_ucx.py
wence-, line 1: **suggestion:** Rename to `test_ucxx.py` ?

wence- (approved):
I have some small suggestions but broadly this looks good to me.

Things that I think are worthwhile fixing:
- remove `versioneer` stuff
- rename `test_ucx.py -> test_ucxx.py`

python/distributed-ucxx/distributed_ucxx/distributed_patches.py
```diff
@@ -0,0 +1,41 @@
...
+    # This patch is insufficient for `dask worker` when `--nworkers=1` (default) or
+    # `--no-nanny` is specified because there's no good way to detect that the
+    # `distributed.Worker.close()` method should stop the notifier thread.
+
+    worker = args[0]  # args[0] == self
+
+    await _worker_close(*args, **kwargs)
+
+    if worker._protocol.startswith("ucxx") and worker.nanny is not None:
+        _stop_notifier_thread_and_progress_tasks()
```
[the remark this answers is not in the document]
pentschev: Correct, it's a no-op if already stopped.

python/distributed-ucxx/pyproject.toml
pentschev: Absolutely, I'll handle that in a follow-up PR. For now I'll keep it as is just to have packages installable for other projects to test.

python/distributed-ucxx/distributed_ucxx/distributed_patches.py
pentschev: TBH, I didn't even think of a scenario where we would have multiple schedulers. In any case, I'm not really sure how we could handle this in a better way than how it's currently written, so anything slightly less common (like multiple schedulers) will demand some further evaluation. Perhaps some sort of reference counting would be useful, but given the unpredictability of garbage collection in Python I'm not sure if that would work either.

python/distributed-ucxx/distributed_ucxx/tests/test_ucx.py
pentschev: Good idea, done via https://github.com/rapidsai/ucxx/pull/60/commits/d58bca4195c1f4d26e06e0977857a95247bf2eae .

python/distributed-ucxx/distributed_ucxx/distributed_patches.py
wence-: No, I think this is fine.

raydouglass (approved):

PR #61, opened 2023-06-27
Add CUDA 12.0 build matrix

### LINKED ISSUE #59: UCXX: CUDA 12 Conda Packages
Pushing for CUDA 12 Conda package builds of libraries for RAPIDS 23.08. This issue is for tracking that work.

Requirements (please recheck):

```[tasklist]
### Developer tasks
- [x] I've confirmed the list below is correct (or updated if not)
```

```[tasklist]
### Required dependencies from Conda
- [x] https://github.com/conda-forge/pynvml-feedstock/pull/11
- [x] https://github.com/conda-forge/ucx-split-feedstock/issues/116
- [ ] https://github.com/rapidsai/rmm/issues/1207
```

```[tasklist]
### Test dependencies from conda-forge
- [x] https://github.com/conda-forge/cupy-feedstock/issues/139
- [x] https://github.com/conda-forge/numba-feedstock/issues/112
- [x] `nvtx`
```

```[tasklist]
### RAPIDS test dependencies
- [ ] https://github.com/rapidsai/cudf/issues/12771
```

pentschev:
Closes #59

### DIFF
--- a/conda/recipes/ucxx/meta.yaml
+++ b/conda/recipes/ucxx/meta.yaml
@@ -135,6 +135,13 @@ outputs:
     requirements:
       build:
         - cmake
+      host:
+        {% if cuda_major == "11" %}
+        - cudatoolkit ={{ cuda_version }}
+        {% else %}
+        - cuda-cudart-dev
+        {% endif %}
+        - cuda-version ={{ cuda_version }}
       run:
         - {{ pin_subpackage('libucxx', exact=True) }}
         {% if cuda_major == "11" %}
--- a/conda/recipes/ucxx/meta.yaml
+++ b/conda/recipes/ucxx/meta.yaml
@@ -41,7 +41,7 @@ requirements:
     - sysroot_{{ target_platform }}
   host:
     {% if cuda_major == "11" %}
-    - cudatoolkit
+    - cudatoolkit ={{ cuda_version }}
     {% else %}
     - cuda-cudart
     {% endif %}
@@ -65,9 +65,9 @@ outputs:
         - ucx
       run:
         {% if cuda_major == "11" %}
-        - cudatoolkit
+        - cudatoolkit {{ cuda_spec }}
         {% endif %}
-        - cuda-version ={{ cuda_spec }}
+        - cuda-version {{ cuda_spec }}
         - ucx
     test:
       commands:
@@ -138,11 +138,11 @@ outputs:
       run:
         - {{ pin_subpackage('libucxx', exact=True) }}
         {% if cuda_major == "11" %}
-        - cudatoolkit
+        - cudatoolkit {{ cuda_spec }}
         {% else %}
         - cuda-cudart
         {% endif %}
-        - cuda-version ={{ cuda_spec }}
+        - cuda-version {{ cuda_spec }}
         - gtest
         - gmock
     about:
--- a/conda/recipes/ucxx/meta.yaml
+++ b/conda/recipes/ucxx/meta.yaml
@@ -180,7 +180,7 @@ outputs:
         {% if cuda_major == "11" %}
         - cudatoolkit {{ cuda_spec }}
         {% else %}
-        - cuda-cudart
+        - cuda-cudart-dev
         {% endif %}
         - cuda-version {{ cuda_spec }}
       run:
--- a/conda/recipes/ucxx/meta.yaml
+++ b/conda/recipes/ucxx/meta.yaml
@@ -43,7 +43,7 @@ requirements:
     {% if cuda_major == "11" %}
     - cudatoolkit ={{ cuda_version }}
     {% else %}
-    - cuda-cudart
+    - cuda-cudart-dev
     {% endif %}
     - cuda-version ={{ cuda_version }}
     - ucx
--- a/conda/recipes/ucxx/meta.yaml
+++ b/conda/recipes/ucxx/meta.yaml
@@ -69,7 +69,11 @@ outputs:
       build:
         - {{ compiler('c') }}
         - {{ compiler('cxx') }}
+        {% if cuda_major == "11" %}
+        - {{ compiler('cuda11') }} ={{ cuda_version }}
+        {% else %}
         - {{ compiler('cuda') }}
+        {% endif %}
         - sysroot_{{ target_platform }}
         - cmake
         - ninja
@@ -145,11 +149,19 @@ outputs:
     build:
       number: {{ GIT_DESCRIBE_NUMBER }}
       string: cuda{{ cuda_major }}_{{ date_string }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
+      ignore_run_exports_from:
+        {% if cuda_major == "11" %}
+        - {{ compiler('cuda11') }}
+        {% endif %}
     requirements:
       build:
         - {{ compiler('c') }}
         - {{ compiler('cxx') }}
+        {% if cuda_major == "11" %}
+        - {{ compiler('cuda11') }} ={{ cuda_version }}
+        {% else %}
         - {{ compiler('cuda') }}
+        {% endif %}
         - sysroot_{{ target_platform }}
         - cmake
         - ninja
--- a/conda/recipes/ucxx/meta.yaml
+++ b/conda/recipes/ucxx/meta.yaml
@@ -144,12 +144,6 @@ outputs:
         - cuda-version ={{ cuda_version }}
       run:
         - {{ pin_subpackage('libucxx', exact=True) }}
-        {% if cuda_major == "11" %}
-        - cudatoolkit {{ cuda_spec }}
-        {% else %}
-        - cuda-cudart
-        {% endif %}
-        - cuda-version {{ cuda_spec }}
         - gtest
         - gmock
     about:
@@ -183,6 +177,12 @@ outputs:
         - {{ pin_subpackage('libucxx', exact=True) }}
         - ucx
         - rmm =23.08
+        {% if cuda_major == "11" %}
+        - cudatoolkit {{ cuda_spec }}
+        {% else %}
+        - cuda-cudart
+        {% endif %}
+        - cuda-version {{ cuda_spec }}
       run:
         - python * *_cpython
         - packaging
--- a/conda/recipes/ucxx/meta.yaml
+++ b/conda/recipes/ucxx/meta.yaml
@@ -41,7 +41,7 @@ requirements:
     - sysroot_{{ target_platform }}
   host:
     {% if cuda_major == "11" %}
-    - cudatoolkit ={{ cuda_version }}
+    - cudatoolkit
     {% else %}
     - cuda-cudart-dev
     {% endif %}
@@ -65,7 +65,7 @@ outputs:
         - ucx
       run:
         {% if cuda_major == "11" %}
-        - cudatoolkit {{ cuda_spec }}
+        - cudatoolkit
         {% endif %}
         - cuda-version {{ cuda_spec }}
         - ucx
@@ -137,7 +137,7 @@ outputs:
         - cmake
       host:
         {% if cuda_major == "11" %}
-        - cudatoolkit ={{ cuda_version }}
+        - cudatoolkit
         {% else %}
         - cuda-cudart-dev
         {% endif %}
@@ -178,7 +178,7 @@ outputs:
         - ucx
         - rmm =23.08
         {% if cuda_major == "11" %}
-        - cudatoolkit {{ cuda_spec }}
+        - cudatoolkit
         {% else %}
         - cuda-cudart-dev
         {% endif %}
--- a/conda/recipes/ucxx/meta.yaml
+++ b/conda/recipes/ucxx/meta.yaml
@@ -5,7 +5,6 @@
 {% set minor_version = version.split('.')[0] + '.' + version.split('.')[1] %}
 {% set cuda_version = '.'.join(environ['RAPIDS_CUDA_VERSION'].split('.')[:2]) %}
 {% set cuda_major = cuda_version.split('.')[0] %}
-{% set cuda_spec = ">=" + cuda_major ~ ",<" + (cuda_major | int + 1) ~ ".0a0" %} # i.e. >=11,<12.0a0
 {% set date_string = environ['RAPIDS_DATE_STRING'] %}
 
 package:
--- a/conda/recipes/ucxx/meta.yaml
+++ b/conda/recipes/ucxx/meta.yaml
@@ -40,7 +40,12 @@ requirements:
     - ninja
     - sysroot_{{ target_platform }}
   host:
-    - cudatoolkit ={{ cuda_version }}
+    {% if cuda_major == "11" %}
+    - cudatoolkit
+    {% else %}
+    - cuda-cudart
+    {% endif %}
+    - cuda-version ={{ cuda_version }}
     - ucx
     - python
 
@@ -59,7 +64,10 @@ outputs:
       host:
         - ucx
       run:
-        - cudatoolkit {{ cuda_spec }}
+        {% if cuda_major == "11" %}
+        - cudatoolkit
+        {% endif %}
+        - cuda-version ={{ cuda_spec }}
         - ucx
     test:
       commands:
@@ -129,7 +137,12 @@ outputs:
         - cmake
       run:
         - {{ pin_subpackage('libucxx', exact=True) }}
-        - cudatoolkit {{ cuda_spec }}
+        {% if cuda_major == "11" %}
+        - cudatoolkit
+        {% else %}
+        - cuda-cudart
+        {% endif %}
+        - cuda-version ={{ cuda_spec }}
         - gtest
         - gmock
     about:
--- a/conda/recipes/ucxx/meta.yaml
+++ b/conda/recipes/ucxx/meta.yaml
@@ -61,6 +61,10 @@ outputs:
       string: cuda{{ cuda_major }}_{{ date_string }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
       run_exports:
         - {{ pin_subpackage("libucxx", max_pin="x.x") }}
+      ignore_run_exports_from:
+        {% if cuda_major == "11" %}
+        - {{ compiler('cuda11') }}
+        {% endif %}
     requirements:
       build:
         - {{ compiler('c') }}
--- a/conda/recipes/ucxx/meta.yaml
+++ b/conda/recipes/ucxx/meta.yaml
@@ -36,12 +36,16 @@ requirements:
     - cmake
     - {{ compiler('c') }}
     - {{ compiler('cxx') }}
+    {% if cuda_major == "11" %}
+    - {{ compiler('cuda11') }} ={{ cuda_version }}
+    {% else %}
+    - {{ compiler('cuda') }}
+    {% endif %}
+    - cuda-version ={{ cuda_version }}
     - ninja
     - sysroot_{{ target_platform }}
   host:
-    {% if cuda_major == "11" %}
-    - cudatoolkit
-    {% else %}
+    {% if cuda_major != "11" %}
     - cuda-cudart-dev
     {% endif %}
     - cuda-version ={{ cuda_version }}
@@ -59,7 +63,12 @@ outputs:
         - {{ pin_subpackage("libucxx", max_pin="x.x") }}
     requirements:
       build:
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
+        - {{ compiler('cuda') }}
+        - sysroot_{{ target_platform }}
         - cmake
+        - ninja
       host:
         - ucx
         - cuda-version ={{ cuda_version }}
@@ -134,16 +143,22 @@ outputs:
       string: cuda{{ cuda_major }}_{{ date_string }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
     requirements:
       build:
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
+        - {{ compiler('cuda') }}
+        - sysroot_{{ target_platform }}
         - cmake
+        - ninja
       host:
-        {% if cuda_major == "11" %}
-        - cudatoolkit
-        {% else %}
+        {% if cuda_major != "11" %}
         - cuda-cudart-dev
         {% endif %}
         - cuda-version ={{ cuda_version }}
       run:
         - {{ pin_subpackage('libucxx', exact=True) }}
+        {% if cuda_major == "11" %}
+        - cudatoolkit
+        {% endif %}
         - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
         - gtest
         - gmock
@@ -178,15 +193,16 @@ outputs:
         - {{ pin_subpackage('libucxx', exact=True) }}
         - ucx
         - rmm =23.08
-        {% if cuda_major == "11" %}
-        - cudatoolkit
-        {% else %}
+        {% if cuda_major != "11" %}
         - cuda-cudart-dev
         {% endif %}
         - cuda-version {{ cuda_version }}
       run:
         - python * *_cpython
         - packaging
+        {% if cuda_major == "11" %}
+        - cudatoolkit
+        {% endif %}
         - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
         - ucx
         - {{ pin_subpackage('libucxx', exact=True) }}
--- a/conda/recipes/ucxx/meta.yaml
+++ b/conda/recipes/ucxx/meta.yaml
@@ -63,11 +63,12 @@ outputs:
         - cmake
       host:
         - ucx
+        - cuda-version ={{ cuda_version }}
       run:
         {% if cuda_major == "11" %}
         - cudatoolkit
         {% endif %}
-        - cuda-version {{ cuda_spec }}
+        - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
         - ucx
     test:
       commands:
@@ -144,6 +145,7 @@ outputs:
         - cuda-version ={{ cuda_version }}
       run:
         - {{ pin_subpackage('libucxx', exact=True) }}
+        - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
         - gtest
         - gmock
     about:
@@ -182,10 +184,11 @@ outputs:
         {% else %}
         - cuda-cudart-dev
         {% endif %}
-        - cuda-version {{ cuda_spec }}
+        - cuda-version {{ cuda_version }}
       run:
         - python * *_cpython
         - packaging
+        - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
         - ucx
         - {{ pin_subpackage('libucxx', exact=True) }}
         - {{ pin_compatible('rmm', max_pin='x.x') }}
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -60,19 +60,27 @@ dependencies:
           - matrix:
               cuda: "11.2"
             packages:
-              - cudatoolkit=11.2
+              - cuda-version=11.2
+              - cudatoolkit
           - matrix:
               cuda: "11.4"
             packages:
-              - cudatoolkit=11.4
+              - cuda-version=11.2
+              - cudatoolkit
           - matrix:
               cuda: "11.5"
             packages:
-              - cudatoolkit=11.5
+              - cuda-version=11.2
+              - cudatoolkit
           - matrix:
               cuda: "11.8"
             packages:
-              - cudatoolkit=11.8
+              - cuda-version=11.2
+              - cudatoolkit
+          - matrix:
+              cuda: "12.0"
+            packages:
+              - cuda-version=12.0
   py_version:
     specific:
       - output_types: conda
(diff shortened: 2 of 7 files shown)


wence- (approved):
I believe this to be right!

jarmak-nv:
Hi @pentschev could you please link this PR to https://github.com/rapidsai/ucxx/issues/59 either by editing in closing keywords to the description or using the UI?

Thanks!

pentschev:
@bdice @jakirkham would you mind taking a look at this? I think it's good now.

conda/recipes/ucxx/meta.yaml
```diff
@@ -40,7 +40,12 @@ requirements:
     - ninja
     - sysroot_{{ target_platform }}
   host:
+    {% if cuda_major == "11" %}
     - cudatoolkit ={{ cuda_version }}
+    {% else %}
+    - cuda-cudart-dev
```
bdice, line 46: Can you describe how `cuda-cudart-dev` is being used? I see there isn't a CUDA compiler being installed here.

The headers in `cuda-cudart-dev` aren't usable by a host compiler (gcc) unless you have the `cuda-nvcc` compiler package installed as well. There is a circular dependency between those packages. 🙃

conda/recipes/ucxx/meta.yaml
```diff
@@ -59,7 +64,10 @@ outputs:
       host:
         - ucx
       run:
+        {% if cuda_major == "11" %}
         - cudatoolkit {{ cuda_spec }}
```
bdice, line 68: We're using `cuda-version` to constrain `cudatoolkit` for CUDA 11 packaging.
```suggestion
        - cudatoolkit
```

conda/recipes/ucxx/meta.yaml
```diff
@@ -59,7 +64,10 @@ outputs:
...
+        {% endif %}
+        - cuda-version {{ cuda_spec }}
```
bdice, line 70: In some cases we've specified `- cuda-version ={{ cuda_version }}` in `host` and then used `pin_compatible` in `run` to generate the min/max spec. [Example from cudf here](https://github.com/rapidsai/cudf/blob/aed7174eae6c6eb38fbf186938df44f88787cf29/conda/recipes/cudf/meta.yaml#L98). Do we want to do that here?

conda/recipes/ucxx/meta.yaml
```diff
@@ -127,9 +135,15 @@ outputs:
     requirements:
       build:
         - cmake
+      host:
+        {% if cuda_major == "11" %}
+        - cudatoolkit ={{ cuda_version }}
```
bdice, line 140:
```suggestion
        - cudatoolkit
```

conda/recipes/ucxx/meta.yaml
```diff
@@ -163,6 +177,12 @@ outputs:
         - {{ pin_subpackage('libucxx', exact=True) }}
         - ucx
         - rmm =23.08
+        {% if cuda_major == "11" %}
+        - cudatoolkit {{ cuda_spec }}
```
bdice, line 181:
```suggestion
        - cudatoolkit
```

bdice (changes requested):
Some questions/suggestions. Generally looks pretty close.

conda/recipes/ucxx/meta.yaml
pentschev: > Can you describe how cuda-cudart-dev is being used? I see there isn't a CUDA compiler being installed here.

It is linking RMM, which needs thrust, so we need to pull it in here. It took me a while to figure that out too.

I guess one alternative is to pull just `cuda-cccl`, but that brings in `cuda-cccl-impl` which isn't needed, so I'm not sure what would be preferred, thus left `cuda-cudart-dev`.

> The headers in cuda-cudart-dev aren't usable by a host compiler (gcc) unless you have the cuda-nvcc compiler package installed as well. There is a circular dependency between those packages. 🙃

Assuming `cuda_runtime.h` is coming from this package (as the name seems to imply), this statement is not true. It's perfectly acceptable to call `cudaMalloc*` from host compiler.

conda/recipes/ucxx/meta.yaml
pentschev: I based this off of an older cuDF review in my local environment, apologies for that. This is and others below are now fixed.

conda/recipes/ucxx/meta.yaml
pentschev: I'm actually now thinking if we should have this in `run` at all. CUDA is not a required dependency, so I believe we should even remove this.

conda/recipes/ucxx/meta.yaml
jakirkham: Done in commit ( https://github.com/rapidsai/ucxx/pull/61/commits/44cab6ffa3eb2609a7341036a0cb5da17781f1b3 )

conda/recipes/ucxx/meta.yaml
jakirkham: Done in commit ( https://github.com/rapidsai/ucxx/pull/61/commits/44cab6ffa3eb2609a7341036a0cb5da17781f1b3 )

conda/recipes/ucxx/meta.yaml
jakirkham: Done in commit ( https://github.com/rapidsai/ucxx/pull/61/commits/44cab6ffa3eb2609a7341036a0cb5da17781f1b3 )

conda/recipes/ucxx/meta.yaml
```diff
@@ -59,7 +64,10 @@ outputs:
       host:
         - ucx
       run:
-        - cudatoolkit {{ cuda_spec }}
+        {% if cuda_major == "11" %}
+        - cudatoolkit
+        {% endif %}
+        - cuda-version {{ cuda_spec }}
```
bdice, line 70: Remove this as a run dependency, per https://github.com/rapidsai/ucxx/pull/61/files#r1244290996
```suggestion
```

conda/recipes/ucxx/meta.yaml
```diff
@@ -163,6 +177,12 @@ outputs:
         - {{ pin_subpackage('libucxx', exact=True) }}
         - ucx
         - rmm =23.08
+        {% if cuda_major == "11" %}
+        - cudatoolkit
+        {% else %}
+        - cuda-cudart-dev
+        {% endif %}
+        - cuda-version {{ cuda_spec }}
```
bdice, line 185: The `cuda-version` in `host` should be a specific value corresponding to the CI matrix entry. If you want to support a wide range of CUDA minor versions at runtime, add `- {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}` in the run section -- assuming that CUDA is a hard dependency of `ucxx` (we discussed above that it isn't a hard dependency of `libucxx`).
```suggestion
        - cuda-version {{ cuda_version }}
```

pentschev: It isn't a hard-dependency, but we still want to support `>=11.2,<12.0`. Is `pin_compatible` not necessary then?

conda/recipes/ucxx/meta.yaml
bdice: RMM should be fetching and patching thrust via rapids-cmake, not using the CUDA Toolkit package `cuda-cccl`.

While it is acceptable to use `<cuda_runtime.h>` from a host compiler, there is a circular dependency (currently beyond our control) between that header and the `crt/` headers shipped in `cuda-nvcc`. As a result, you actually need `cuda-nvcc` installed even if you only intend to call CUDA runtime functions from a C++ host compiler.

conda/recipes/ucxx/meta.yaml
bdice: This is tricky. If CUDA is not a hard dependency, it shouldn't be in `run`. Yet if the optional CUDA bits differ between the two builds, I think we still need two package variants, one constrained to CUDA 11 and one constrained to CUDA 12. So maybe we need to put `cuda-version` in `run_constrained` so that it isn't _installed_, but limits that ucxx build to a particular CUDA major version?

conda/recipes/ucxx/meta.yaml
```diff
@@ -40,7 +40,12 @@ requirements:
     - ninja
```
bdice, line 40: Please copy these lines into `build` for `{{ compiler('cuda') }}`:

https://github.com/rapidsai/cudf/blob/de4a20ec9089a77e7c0a106794ba863e3d0083df/conda/recipes/libcudf/meta.yaml#L40-L45

And also these lines into `conda_build_config.yaml`:
https://github.com/rapidsai/cudf/blob/de4a20ec9089a77e7c0a106794ba863e3d0083df/conda/recipes/libcudf/conda_build_config.yaml#L7-L11

conda/recipes/ucxx/meta.yaml
```diff
@@ -59,7 +64,10 @@ outputs:
       host:
         - ucx
```
bdice, line 65:
```suggestion
        - ucx
        - cuda-version ={{ cuda_version }}
```

conda/recipes/ucxx/meta.yaml
bdice: (We discussed offline and CUDA _is_ a hard dependency. The comment below happens to be on the same line as the previous discussion thread but this comment is no longer related to the thread above.)

This will use the pinning from `host` and allow any compatible same-major version.
```suggestion
        - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
```

conda/recipes/ucxx/meta.yaml
```diff
@@ -127,9 +135,15 @@ outputs:
...
         - cmake
+      host:
+        {% if cuda_major == "11" %}
+        - cudatoolkit
+        {% else %}
+        - cuda-cudart-dev
+        {% endif %}
+        - cuda-version ={{ cuda_version }}
       run:
         - {{ pin_subpackage('libucxx', exact=True) }}
```
bdice, line 146:
```suggestion
        - {{ pin_subpackage('libucxx', exact=True) }}
        - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
```

conda/recipes/ucxx/meta.yaml
bdice: This needs to match the CI job's CUDA version:
```suggestion
        - cuda-version {{ cuda_version }}
```

conda/recipes/ucxx/meta.yaml
```diff
@@ -163,6 +177,12 @@ outputs:
...
         - rmm =23.08
+        {% if cuda_major == "11" %}
+        - cudatoolkit
+        {% else %}
+        - cuda-cudart-dev
+        {% endif %}
+        - cuda-version {{ cuda_spec }}
       run:
         - python * *_cpython
         - packaging
```
bdice, line 188:
```suggestion
        - packaging
        - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
```

bdice (commented):
Suggesting changes after offline discussion. At present, ucxx has a hard dependency on CUDA when built with RMM enabled. As such, we need to ensure that the packages are built with pinnings that constrain their compatible CUDA versions. These changes will align ucxx closer to other RAPIDS packages.

conda/recipes/ucxx/meta.yaml
pentschev: I'm not sure what you're saying here exactly, are you saying we need `cuda-nvcc` installed to build? That didn't seem to be necessary, and the error was completely resolved after installing `cuda-cudart-dev`.

bdice: 🤷‍♂️ I won't argue if CI is passing. But yes, that's what I'm saying -- `cuda-nvcc` is needed to use most of `cuda-cudart-dev` even if you're using a host C++ compiler and not nvcc to build. I'm not sure what piece of `cuda-cudart-dev` is needed here, though, so maybe this is an exception.

conda/recipes/ucxx/meta.yaml
bdice, line 1: You can remove the declaration of `cuda_spec` and `minor_version` from the top of this file. They are unused now.

bdice (approved):
A couple minor cleanup comments but otherwise this is probably complete? I feel a little less confident about this repo than some, just because I'm not as familiar with it, but I am approving. A second CUDA 12 approval would be good.

conda/recipes/ucxx/meta.yaml
pentschev: From what I could understand the only piece we need here is really `cuda-cccl`, but since that brought in `cuda-cccl-impl` that `cuda-cudart-dev` does not, I felt it was safer to just use `cuda-cudart-dev`. That is passing now, and to be fair I still don't get why `cuda-nvcc` would ever be required here, so since CI is passing anyway I'd prefer to keep it that way.

conda/recipes/ucxx/meta.yaml
pentschev: Done in https://github.com/rapidsai/ucxx/pull/61/commits/64ca814c0090252896f581f3f9ab60e9d97945a0 .

vyasr (changes requested):
Something is off here. I was suspicious about the same thing @bdice was, the fact that you didn't need cuda-nvcc, and sure enough the builds are finding the CTK in [the wrong place](https://github.com/rapidsai/ucxx/actions/runs/5416563552/jobs/9846479684?pr=61#step:7:643):
```
-- Found CUDAToolkit: /usr/local/cuda/include (found version "12.0.140")
```

This should say something like
```
-- Found CUDAToolkit: $PREFIX/... 
```
(or maybe `BUILD_PREFIX` depending on which env you find it in).

Moreover, the CUDA version being found is incorrect. I believe the latest patch version uploaded to cf is 12.0.107 (for example, see https://anaconda.org/conda-forge/cuda-cudart/files).

I would guess that the oddity that made Bradley uncomfortable (the lack of cuda-nvcc) is also the root cause behind why CMake is finding the wrong CTK. However, the fact that it's finding the CTK outside the conda environment is also probably masking the errors that you would otherwise see if attempting to use cuda-cudart-dev without cuda-nvcc. At least, that's my best guess.

(I suspect that future improved support for splayed CUDA layouts in CMake may improve support for the current setup as well)

conda/recipes/ucxx/meta.yaml
jakirkham: Given Vyas' observation below ( https://github.com/rapidsai/ucxx/pull/61#pullrequestreview-1506222578 ), think we need the CUDA compiler

For CCCL we may want to look into using `rapids-cmake` here

In terms of `cuda-cudart-dev`, it is also needed as UCXX appears to be dynamically linking to `libcudart`, which is provided by `cuda-cudart` (pulled in and exported to packages that use `cuda-cudart-dev`)

conda/recipes/ucxx/meta.yaml
jakirkham: This suggestion includes the first piece

```suggestion
    {% if cuda_major == "11" %}
    - {{ compiler('cuda11') }} ={{ cuda_version }}
    {% else %}
    - {{ compiler('cuda') }}
    {% endif %}
    - cuda-version ={{ cuda_version }}
    - ninja
```

<hr>

The second is to add a `conda_build_config.yaml` file with this content adjacent to `meta.yaml`:

```yaml
cuda_compiler:
  - cuda-nvcc

cuda11_compiler:
  - nvcc
```

conda/recipes/ucxx/meta.yaml
```diff
@@ -40,7 +39,12 @@ requirements:
     - ninja
     - sysroot_{{ target_platform }}
   host:
-    - cudatoolkit ={{ cuda_version }}
+    {% if cuda_major == "11" %}
+    - cudatoolkit
+    {% else %}
+    - cuda-cudart-dev
+    {% endif %}
```
jakirkham, line 46: Once the CUDA compiler portion is added, think we can drop the CUDA 11 portion

The CUDA 12 portion is still needed since `libcudart` is dynamically linked to

```suggestion
    {% if cuda_major != "11" %}
    - cuda-cudart-dev
    {% endif %}
```

conda/recipes/ucxx/meta.yaml
```diff
@@ -127,9 +135,16 @@ outputs:
...
+        {% if cuda_major == "11" %}
+        - cudatoolkit
+        {% else %}
+        - cuda-cudart-dev
+        {% endif %}
+        - cuda-version ={{ cuda_version }}
       run:
         - {{ pin_subpackage('libucxx', exact=True) }}
-        - cudatoolkit {{ cuda_spec }}
+        - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
```
jakirkham, line 147: To make sure the `libcudart` dependency is satisfied at `run` time, would suggest adding `cudatoolkit` here (especially since we are relaxing the version constraint on it above

```suggestion
        {% if cuda_major == "11" %}
        - cudatoolkit
        {% endif %}
        - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
```

conda/recipes/ucxx/meta.yaml
```diff
@@ -163,9 +178,16 @@ outputs:
...
+        {% if cuda_major == "11" %}
+        - cudatoolkit
+        {% else %}
+        - cuda-cudart-dev
+        {% endif %}
+        - cuda-version {{ cuda_version }}
       run:
         - python * *_cpython
         - packaging
+        - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
```
jakirkham, line 190: To make sure the `libcudart` dependency is satisfied at `run` time, would suggest adding `cudatoolkit` here (especially since we are relaxing the version constraint on it above

```suggestion
        {% if cuda_major == "11" %}
        - cudatoolkit
        {% endif %}
        - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
```

dependencies.yaml
```diff
@@ -60,19 +60,27 @@ dependencies:
...
           - matrix:
               cuda: "11.8"
             packages:
-              - cudatoolkit=11.8
+              - cuda-version=11.8
+              - cudatoolkit
+          - matrix:
+              cuda: "12.0"
+            packages:
+              - cuda-version=12.0
```
jakirkham, line 83: Would suggest either this or the `-dev` package depending on whether the goal is a working development environment (where the headers are needed) or a working runtime environment (where only the libraries are needed)

```suggestion
              - cuda-version=12.0
              - cuda-cudart
```

jakirkham (commented):
Thanks all! 🙏

Following up on the point Vyas observed above about CUDA Toolkit detection by CMake, it sounds like we need the CUDA compiler. This also makes sense when considering the usage of RMM and the linkage to `libcudart`

On a different note, am seeing references to [a newer GLIBC]( https://github.com/rapidsai/ucxx/actions/runs/5416563552/jobs/9846479684?pr=61#step:7:2133 ), which is missed. Think we need to [add `sysroot_{{ target_platform }}`]( https://github.com/rapidsai/cudf/blob/e7a1448816b9d83c3f69676cf23fa38f45891ab1/conda/recipes/cudf/meta.yaml#L56 ) throughout the `build` sections and [specify the `2.17` version]( https://github.com/rapidsai/cudf/blob/e7a1448816b9d83c3f69676cf23fa38f45891ab1/conda/recipes/cudf/conda_build_config.yaml#L7-L8 )

pentschev:
I see @bdice @jakirkham and @vyasr all agree we need `cuda-nvcc`, my understanding given Vyas' comment https://github.com/rapidsai/ucxx/pull/61#pullrequestreview-1506222578 is that we need it because the package will set the correct paths to find CUDA correctly, which makes sense, is that why we need it? If so, I don't think this is an obvious requirement for the community in general, I'd never expect we need `cuda-nvcc` if we're not using it for anything (we only need `gcc`/`g++`), does it make sense to then ensure `cuda-cudart-dev` (and possibly other `-dev` packages) set the appropriate paths instead?

jakirkham:
Essentially the CUDA compiler package is needed to configure the search paths correctly with CMake, Conda compilers, etc.. There's not a good way atm to handle the needs of these cases without that package as it requires things like add `nvcc.profile` files (and patching them) as well as altering the flags the Conda compilers get, which basically only makes sense in the CUDA compiler package

pentschev:
> Essentially the CUDA compiler package is needed to configure the search paths correctly with CMake, Conda compilers, etc.. There's not a good way atm to handle the needs of these cases without that package as it requires things like adding `nvcc.profile` files (and patching them) as well as altering the flags the Conda compilers get, which basically only makes sense in the CUDA compiler package

Do you think that can be (or already is) documented somewhere? I think this is not an obvious requirement and may cause confusion.

jakirkham:
Please make any documentation requests in issue ( https://github.com/conda-forge/conda-forge.github.io/issues/1927 )

pentschev:
@jakirkham I applied your suggestions and both 11.8 and 12.0 builds are failing:

```
The reported errors are:
- Encountered problems while solving:
-   - nothing provides requested cuda_linux-64
```

Are we missing something more here? Also tried adding [`ignore_run_exports_from` section](https://github.com/rapidsai/ucxx/pull/61/commits/3fea7758d0f025db96a8ada446e3ca487c83819c) as you suggested, but the same error persists.

jakirkham:
Thanks Peter! 🙏

Yeah think we need to add `conda_build_config.yaml` files as noted here ( https://github.com/rapidsai/ucxx/pull/61#discussion_r1247444774 ):

> ... add a `conda_build_config.yaml` file with this content adjacent to `meta.yaml`:
> 
> ```yaml
> cuda_compiler:
>   - cuda-nvcc
> 
> cuda11_compiler:
>   - nvcc
> ```

Unfortunately there's not a good way to create a code suggestion for a new file

pentschev:
> Thanks Peter! 🙏
> 
> Yeah think we need to add `conda_build_config.yaml` files as noted here ( [#61 (comment)](https://github.com/rapidsai/ucxx/pull/61#discussion_r1247444774) ):
> 
> > ... add a `conda_build_config.yaml` file with this content adjacent to `meta.yaml`:
> > ```yaml
> > cuda_compiler:
> >   - cuda-nvcc
> > 
> > cuda11_compiler:
> >   - nvcc
> > ```
> 
> Unfortunately there's not a good way to create a code suggestion for a new file

Thanks John. I'm a bit confused about some of the changes you've suggested though, like adding compilers to the [`build` section of `libucxx](https://github.com/rapidsai/ucxx/blob/3fea7758d0f025db96a8ada446e3ca487c83819c/conda/recipes/ucxx/meta.yaml#L70-L73)`, I don't see the same for [`libcudf`](https://github.com/rapidsai/cudf/blob/de4a20ec9089a77e7c0a106794ba863e3d0083df/conda/recipes/libcudf/meta.yaml#L91-L92). Would you mind explaining why we want it here if it's not in there? I think the requirements are somewhat similar.

conda/recipes/ucxx/meta.yaml
```diff
@@ -53,13 +61,26 @@ outputs:
...
         - {{ pin_subpackage("libucxx", max_pin="x.x") }}
+      ignore_run_exports_from:
+        {% if cuda_major == "11" %}
+        - {{ compiler('cuda11') }}
+        {% endif %}
     requirements:
       build:
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
+        - {{ compiler('cuda') }}
```
jakirkham, line 72: Think the original suggestion here needed a tweak for CUDA 11 handling, which I had missed

```suggestion
        {% if cuda_major == "11" %}
        - {{ compiler('cuda11') }} ={{ cuda_version }}
        {% else %}
        - {{ compiler('cuda') }}
        {% endif %}
```

conda/recipes/ucxx/meta.yaml
```diff
@@ -126,10 +147,23 @@ outputs:
       string: cuda{{ cuda_major }}_{{ date_string }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
     requirements:
       build:
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
+        - {{ compiler('cuda') }}
```
jakirkham, line 152: Think the original suggestion here needed a tweak for CUDA 11 handling, which I had missed

```suggestion
        {% if cuda_major == "11" %}
        - {{ compiler('cuda11') }} ={{ cuda_version }}
        {% else %}
        - {{ compiler('cuda') }}
        {% endif %}
```

conda/recipes/ucxx/meta.yaml
jakirkham, line 72: Think the original suggestion here needed a tweak for CUDA 11 handling, which I had missed

```suggestion
        {% if cuda_major == "11" %}
        - {{ compiler('cuda11') }} ={{ cuda_version }}
        {% else %}
        - {{ compiler('cuda') }}
        {% endif %}
```

conda/recipes/ucxx/meta.yaml
```diff
@@ -126,10 +147,23 @@ outputs:
       string: cuda{{ cuda_major }}_{{ date_string }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
```
jakirkham, line 147:
```suggestion
      string: cuda{{ cuda_major }}_{{ date_string }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
      ignore_run_exports_from:
        {% if cuda_major == "11" %}
        - {{ compiler('cuda11') }}
        {% endif %}
```

jakirkham (commented):
Thanks Peter! 🙏

Think the original suggestions I had made missed some subtleties with CUDA 11 / 12 compilers. Have proposed some revisions to them below

jakirkham:
As to whether we want compilers in each section or not. That is a good question. Am seeing [this portion of cuDF]( https://github.com/rapidsai/cudf/blob/de4a20ec9089a77e7c0a106794ba863e3d0083df/conda/recipes/libcudf/meta.yaml#L400-L404 ) where an output does include the compilers, but you are right others don't. Think we need to handle these consistently somehow

pentschev:
> As to whether we want compilers in each requirement section or not. That is a good question. Am seeing [this portion of cuDF](https://github.com/rapidsai/cudf/blob/de4a20ec9089a77e7c0a106794ba863e3d0083df/conda/recipes/libcudf/meta.yaml#L400-L404) where an output does include the compilers, but you are right others don't. Think we need to handle these consistently somehow

Makes sense, thanks John for checking that!

pentschev:
Alright, it seems everything is passing now. I've also checked the CUDA toolkit path being found, for [CUDA 11](https://github.com/rapidsai/ucxx/actions/runs/5421994969/jobs/9858135890) we see:

```
-- Found CUDAToolkit: /usr/local/cuda/include (found version "11.8.89") 
```

I believe the above is correct, as it is [consistent with a build from 2 weeks ago](https://github.com/rapidsai/ucxx/actions/runs/5271772385/jobs/9533168308?pr=60#step:7:641) for a different PR. For [CUDA 12](https://github.com/rapidsai/ucxx/actions/runs/5421994969/jobs/9858136041#step:7:658) we see what [Vyas mentioned CMake should be finding](https://github.com/rapidsai/ucxx/pull/61#pullrequestreview-1506222578):

```
-- Found CUDAToolkit: $BUILD_PREFIX/targets/x86_64-linux/include (found version "12.0.76") 
```

With the above said, I believe this PR is now consistent with what is generally expected. Could you take another, hopefully final, look at this @bdice @jakirkham @vyasr ?

vyasr:
> I see @bdice @jakirkham and @vyasr all agree we need `cuda-nvcc`, my understanding given Vyas' comment [#61 (review)](https://github.com/rapidsai/ucxx/pull/61#pullrequestreview-1506222578) is that we need it because the package will set the correct paths to find CUDA correctly, which makes sense, is that why we need it? If so, I don't think this is an obvious requirement for the community in general, I'd never expect we need `cuda-nvcc` if we're not using it for anything (we only need `gcc`/`g++`), does it make sense to then ensure `cuda-cudart-dev` (and possibly other `-dev` packages) set the appropriate paths instead?

That is not the reason that Bradley was originally pointing out that it would be needed. This is the first time we are seeing this particular failure mode (where not including cuda-nvcc results in a host CTK being found), or at least the first such case since the very early days when we were still tweaking all the conda-forge packages extensively. I just knew that it shouldn't work with the current approach, so I took a close look at the build outputs to understand what was happening and I observed that the wrong CTK was being found.

The reason that we definitely need cuda-nvcc is that this package installs headers in `include/crt/*`, and those headers are included by other headers installed in cuda-cudart-dev. For example: 
```
(main) vyasr@vyasr-mlt tmp % rg crt .
./cuda-cudart-dev_linux-64-12.0.107-h59595ed_6/targets/x86_64-linux/include/device_functions.h
60:#include "crt/device_functions.h"
```

It turns out that there is a dependency cycle in the way that nvcc, cccl, and cudart have been historically packaged that results in all of them needing each other. With the conda packages that becomes a bit more complex due to the `*impl` and target-specific packages.

Agree that we need good docs, and would be happy to see this added somewhere. In general there is a lot of information about these conda-forge packages that we plan to document.

conda/recipes/ucxx/meta.yaml
vyasr: This is now resolved.

vyasr (approved):

vyasr:
I agree that we need to be consistent with the compilers, but let's maybe come back and do that in a future pass?

jakirkham (approved):

bdice (approved):
Approving. Logs look mostly fine, aside from problems noted already in https://github.com/rapidsai/ucxx/issues/20.

raydouglass (approved):

pentschev:
Thanks everyone for all the reviews and fixes here! 😄

jakirkham:
Thanks Peter and everyone for the reviews! 🙏

PR #62, opened 2023-06-29
Derive `ucxx::RequestTagMulti` from `ucxx::Request`

pentschev:
Large parts of `ucxx::RequestTagMulti`'s structure was derived from `ucxx::Request`, thus making it a derived class is the natural way of reducing code duplication.

Additionally fix lifetime of `ucxx::RequestTagMulti` by ensuring a reference to `std::shared_ptr<ucxx::RequestTagMulti>` is held until all children callback finish execution. If the owner of the
`std::shared_ptr<ucxx::RequestTagMulti>` releases it too quickly, the object's refcount may go to 0 before the last child send/receive request completes its callback, thus causing invalid memory usage.

### DIFF
--- a/cpp/src/endpoint.cpp
+++ b/cpp/src/endpoint.cpp
@@ -17,6 +17,7 @@
 #include <ucxx/request_am.h>
 #include <ucxx/request_stream.h>
 #include <ucxx/request_tag.h>
+#include <ucxx/request_tag_multi.h>
 #include <ucxx/typedefs.h>
 #include <ucxx/utils/sockaddr.h>
 #include <ucxx/utils/ucx.h>
@@ -269,21 +270,21 @@ std::shared_ptr<Request> Endpoint::tagRecv(void* buffer,
     endpoint, false, buffer, length, tag, enablePythonFuture, callbackFunction, callbackData));
 }
 
-std::shared_ptr<RequestTagMulti> Endpoint::tagMultiSend(const std::vector<void*>& buffer,
-                                                        const std::vector<size_t>& size,
-                                                        const std::vector<int>& isCUDA,
-                                                        const ucp_tag_t tag,
-                                                        const bool enablePythonFuture)
+std::shared_ptr<Request> Endpoint::tagMultiSend(const std::vector<void*>& buffer,
+                                                const std::vector<size_t>& size,
+                                                const std::vector<int>& isCUDA,
+                                                const ucp_tag_t tag,
+                                                const bool enablePythonFuture)
 {
   auto endpoint = std::dynamic_pointer_cast<Endpoint>(shared_from_this());
-  return createRequestTagMultiSend(endpoint, buffer, size, isCUDA, tag, enablePythonFuture);
+  return registerInflightRequest(
+    createRequestTagMultiSend(endpoint, buffer, size, isCUDA, tag, enablePythonFuture));
 }
 
-std::shared_ptr<RequestTagMulti> Endpoint::tagMultiRecv(const ucp_tag_t tag,
-                                                        const bool enablePythonFuture)
+std::shared_ptr<Request> Endpoint::tagMultiRecv(const ucp_tag_t tag, const bool enablePythonFuture)
 {
   auto endpoint = std::dynamic_pointer_cast<Endpoint>(shared_from_this());
-  return createRequestTagMultiRecv(endpoint, tag, enablePythonFuture);
+  return registerInflightRequest(createRequestTagMultiRecv(endpoint, tag, enablePythonFuture));
 }
 
 std::shared_ptr<Worker> Endpoint::getWorker() { return ::ucxx::getWorker(_parent); }
--- a/cpp/src/request_tag_multi.cpp
+++ b/cpp/src/request_tag_multi.cpp
@@ -114,10 +114,8 @@ void RequestTagMulti::recvFrames()
     for (size_t i = 0; i < h.nframes; ++i) {
       auto bufferRequest = std::make_shared<BufferRequest>();
       _bufferRequests.push_back(bufferRequest);
-      const auto bufferType = h.isCUDA[i] ? ucxx::BufferType::RMM : ucxx::BufferType::Host;
-      auto buf              = allocateBuffer(bufferType, h.size[i]);
-      bufferRequest->requestTagMulti =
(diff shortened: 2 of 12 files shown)


cpp/src/request_tag_multi.cpp
```diff
@@ -127,8 +114,10 @@ void RequestTagMulti::recvFrames()
     for (size_t i = 0; i < h.nframes; ++i) {
       auto bufferRequest = std::make_shared<BufferRequest>();
       _bufferRequests.push_back(bufferRequest);
-      const auto bufferType  = h.isCUDA[i] ? ucxx::BufferType::RMM : ucxx::BufferType::Host;
-      auto buf               = allocateBuffer(bufferType, h.size[i]);
+      const auto bufferType = h.isCUDA[i] ? ucxx::BufferType::RMM : ucxx::BufferType::Host;
+      auto buf              = allocateBuffer(bufferType, h.size[i]);
+      bufferRequest->requestTagMulti =
+        std::static_pointer_cast<RequestTagMulti>(shared_from_this());
```
wence-, line 120: Bit confused here. `_bufferRequests` is a member variable of `this` (a `RequestTagMulti`), and this seems to be making a reference cycle by saving a shared pointer to `this` in the buffer request that is then saved on `this`.

So why does this not create a situation like:
```
this---->bufferRequest
 ^            |
 |            |
 `------------'
```
?

wence-: Is it that someone else then gets hold of the `bufferRequest` objects (rather than holding on to the `RequestTagMulti` object), so something like this might happen (pseudo-code):

```
multi_request = ...;
multi_request->recvFrames(...);
buf_requests = multi_request->get_requests(...);
// lose reference to multi_request
...;
buf_requests->complete(); // uh-oh, multi_request which is needed might be gone
```

wence-: IOW, whose job is it to hold on to the `multi_request`, and how does this change help, except by leaking?

Is this situation now not like:
```c++
#include <memory>
#include <iostream>
struct B;

struct A {
  std::shared_ptr<B> b;
  A() { std::cout << "Created A" << std::endl; }
  ~A() { std::cout << "Deleted A" << std::endl; }
};

struct B {
  std::shared_ptr<A> a;
  B() { std::cout << "Created B" << std::endl; }
  ~B() { std::cout << "Deleted B" << std::endl; }
};

int main(void)
{
  auto a = std::make_shared<A>();
  auto b = std::make_shared<B>();
  b->a = a;
  a->b = b;
}
```
Which never calls the destructors for `A` and `B`?

cpp/src/request_tag_multi.cpp
```diff
@@ -22,15 +20,20 @@ namespace ucxx {
...
+    _tag(tag)
 {
   ucxx_trace_req("RequestTagMulti::RequestTagMulti [recv]: %p, tag: %lx", this, _tag);
 
   auto worker = endpoint->getWorker();
   if (enablePythonFuture) _future = worker->getFuture();
 
   ucxx_debug("RequestTagMulti created: %p", this);
-  callback(UCS_OK);
+  recvCallback(UCS_OK);
```
wence-: Remind me why we call the callback on construction of the request?

wence-: Ah, is it to kick off the `recvHeader` call?

cpp/src/endpoint.cpp
```diff
@@ -269,21 +270,21 @@ std::shared_ptr<Request> Endpoint::tagRecv(void* buffer,
...
+    createRequestTagMultiSend(endpoint, buffer, size, isCUDA, tag, enablePythonFuture));
 }
 
-std::shared_ptr<RequestTagMulti> Endpoint::tagMultiRecv(const ucp_tag_t tag,
-                                                        const bool enablePythonFuture)
+std::shared_ptr<Request> Endpoint::tagMultiRecv(const ucp_tag_t tag, const bool enablePythonFuture)
 {
   auto endpoint = std::dynamic_pointer_cast<Endpoint>(shared_from_this());
-  return createRequestTagMultiRecv(endpoint, tag, enablePythonFuture);
+  return registerInflightRequest(createRequestTagMultiRecv(endpoint, tag, enablePythonFuture));
```
wence-, line 287: It seems like this registration of the inflight request (since it holds shared pointers) is the thing that should keep the multi request alive?

wence- (commented):
The inheritance changes look fine I think, but I am rather confused about the changes to keep the requesttagmulti alive. It _looks_ like you're just making a reference cycle so that things are leaked, but I might be missing something.

cpp/src/request_tag_multi.cpp
pentschev: Thanks @wence- for realizing this and taking the time to discuss that with me last week, you were spot on in your review. I've since reverted those changes and instead ensure that the callback takes a `shared_from_this()` reference [here](https://github.com/rapidsai/ucxx/pull/62/files#diff-525902688d0bc5f703bd2edfe719a1fe3593546160de89ab66284271c596bc48R105-R111) and [here](https://github.com/rapidsai/ucxx/pull/62/files#diff-b15e3d4a4461df6e15af9a85397fb8f5d2340cc6062d532405e8de68074855d8R150-R157), and thus ensure the object remains valid at least until `ucxx::Request::callback()`/`ucxx::RequestTagMulti::markCompleted()` finish running, independent of the user releasing that as soon as the status gets set. This should also prevent any cycles.

cpp/src/request_tag_multi.cpp
pentschev: That's right. I think this is not the best naming, but essentially `recvCallback` is a recursive method, so we kick it off here.

cpp/src/endpoint.cpp
pentschev: That's right too, now all request types should have their lifetime ensured by this in addition to the changes mentioned in https://github.com/rapidsai/ucxx/pull/62#discussion_r1255690494.

pentschev:
Thanks again @wence- for the review. I believe this is now ready for another round when you have the chance. I've fixed those issues you raised and added a few more fixes I found were required during that work.

wence- (approved):
Thanks @pentschev this now looks correct to me from a lifetime management point of view.

pentschev:
Thanks again @wence- , appreciate the review! 😄

PR #64, opened 2023-06-30
Add librmm and gtest to host environment.

### LINKED ISSUE #20: ClobberWarnings from thrust, cub, spdlog

bdice:
Currently conda builds are fetching `rmm` from CPM, rather than using conda packages. This PR aims to use conda for as many dependencies as possible to avoid repackaging and clobbering issues.

Resolves #20.

Related to #22, #54, similar to https://github.com/rapidsai/cudf/issues/13230

### DIFF
--- a/ci/release/update-version.sh
+++ b/ci/release/update-version.sh
@@ -40,6 +40,7 @@ sed_runner "s/^version = .*/version = \"${NEXT_FULL_TAG}\"/g" python/pyproject.t
 
 
 # bump RAPIDS libs
+sed_runner "/- librmm =/ s/=.*/=${NEXT_RAPIDS_VERSION}/g" conda/recipes/ucxx/meta.yaml
 sed_runner "/- rmm =/ s/=.*/=${NEXT_RAPIDS_VERSION}/g" conda/recipes/ucxx/meta.yaml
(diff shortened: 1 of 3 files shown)


bdice:
I confirmed that this resolves #20. The dependencies are no longer being fetched by CPM (it finds the local conda packages) and the headers are not being shipped in the resulting packages from what I can see in the logs.

jakirkham (approved):
Thanks Bradley! 🙏

pentschev (approved):
LGTM, thanks @bdice .

raydouglass (approved):

PR #66, opened 2023-07-10
Use gtest version from conda_build_config.yaml.

bdice:
This PR adds appropriate pinnings for `gtest` and `gmock` in `libucxx-tests`.

### DIFF
--- a/conda/recipes/ucxx/meta.yaml
+++ b/conda/recipes/ucxx/meta.yaml
@@ -178,8 +178,8 @@ outputs:
         - cudatoolkit
         {% endif %}
         - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
-        - gtest {{ gtest_version }}
-        - gmock {{ gtest_version }}
+        - gtest
+        - gmock
     about:
       home: https://rapids.ai/
       license: BSD-3-Clause
--- a/conda/recipes/ucxx/meta.yaml
+++ b/conda/recipes/ucxx/meta.yaml
@@ -52,7 +52,7 @@ requirements:
     - ucx
     - python
     - librmm =23.08
-    - gtest {{ gtest_version }}
+    - gtest
 
 outputs:
   - name: libucxx
--- a/conda/recipes/ucxx/meta.yaml
+++ b/conda/recipes/ucxx/meta.yaml
@@ -178,8 +178,8 @@ outputs:
         - cudatoolkit
         {% endif %}
         - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
-        - gtest
-        - gmock
+        - gtest {{ gtest_version }}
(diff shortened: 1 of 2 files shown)


pentschev (approved):
LGTM, thanks @bdice !

conda/recipes/ucxx/meta.yaml
```diff
@@ -178,8 +178,8 @@ outputs:
         - cudatoolkit
         {% endif %}
         - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
-        - gtest
-        - gmock
+        - gtest {{ gtest_version }}
+        - gmock {{ gtest_version }}
```
jakirkham, line 182: Would it be possible to change...

https://github.com/rapidsai/ucxx/blob/c3235d083bce472f7962c01a1a9c8867b1c3a034/conda/recipes/ucxx/conda_build_config.yaml#L26-L27

...to this?

```yaml
gmock:
  - ">=1.13.0"
gtest:
  - ">=1.13.0"
```

That way conda-build will pull in the versions for us (without needing templating) as is done with the other dependencies already in `conda_build_config.yaml`

bdice: Interesting, I didn't know that conda would do this automatic pinning. We haven't been using this feature for other RAPIDS repos and I'm not sure if I like it, since it hides the fact that there is a version pinning from readers of `meta.yaml`. It's not very friendly to non-experts in conda build behavior, unfortunately. I'm somewhat inclined to do the opposite and change `cmake` to use a `cmake_version` variable in ucxx since that's how we pin `cmake` in other RAPIDS repos. What do you think @jakirkham?

bdice: If you're confident that using pinnings like this in `conda_build_config.yaml` is best practice, we should file an issue and update the rest of RAPIDS. It just surprises me since I've been building recipes for some time without knowing about that feature.

bdice: Updated in 4246795.

bdice:
Just realized this PR is still open. @pentschev I'm going to retrigger CI and see if we can get this in for 0.33?

conda/recipes/ucxx/meta.yaml
bdice, line 182:
```suggestion
        - gtest
        - gmock
```

jakirkham:
One of the CI jobs failed. Tried restarting it, but it failed again. The test failure looks different (though unfortunately I don't remember the previous test that failed)

pentschev:
Thanks @bdice and @jakirkham . Indeed, CI is still unfortunately failing randomly, apologies for the hassle. After rerunning once more all tests now passed.

raydouglass (approved):

jakirkham:
Thanks Peter! 🙏

PR #69, opened 2023-07-20
Sync stream after allocation and result copy to host

wence-:
Stream synchronization was missing from C++ RMM tests, which caused data races with errors such as below:

```
[  FAILED  ] RMMProgressModes/RequestTest.ProgressAm/9, where GetParam() = (4-byte object <01-00 00-00>, false, false, 4-byte object <03-00 00-00>, 1024) (10 ms)
```


pentschev (approved):
Thanks so much for fixing this @wence- !

PR #77, opened 2023-08-17
Fixes for `RequestTagMulti` immediate completion

pentschev:
When the last underlying `RequestTag` transfer completed immediately during a `RequestTagMulti`, two issues may have occurred before the changes here:

1. The `BufferRequest` may not have been pushed into the `_bufferRequests` container, thus reordering is necessary; and
2. The `_bufferRequests->request` was still unpopulated as the request and `markCompleted()` callback were called before returning the `RequestTag` back to the caller.

### DIFF
--- a/cpp/src/request_tag_multi.cpp
+++ b/cpp/src/request_tag_multi.cpp
@@ -164,6 +164,9 @@ void RequestTagMulti::markCompleted(ucs_status_t status, RequestCallbackUserData
         if (s != UCS_OK) break;
       }
     }
+    // Check the status of the current (and final) message as it may have completed before
+    // `_bufferRequests->request` was populated.
+    if (s == UCS_OK) s = status;
 
     setStatus(s);
   }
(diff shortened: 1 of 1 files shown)


cpp/src/request_tag_multi.cpp
```diff
@@ -164,6 +164,9 @@ void RequestTagMulti::markCompleted(ucs_status_t status, RequestCallbackUserData
         if (s != UCS_OK) break;
       }
     }
+    // Check the status of the current (and final) message as it may have completed before
+    // `_bufferRequests->request` was populated.
+    if (s == UCS_OK) s = status;
```
wence-, line 169: You could initialise `s` to `status` on line 158? Would that have the same effect?

wence- (approved):
Makes sense, thanks.

cpp/src/request_tag_multi.cpp
pentschev: The current promise is that the [error (if any) is that of the first failed request](https://github.com/rapidsai/ucxx/blob/1c8e12d0cdd7abfc1ce2e4d3008dfb662beda24a/cpp/include/ucxx/request_tag_multi.h#L202-L205). I'm not saying that is the "right" promise, but I just didn't know what would be the most sensible promise to make. If we did initialize `s` to the value of `status` in line 158 we would break that promise.

wence-: Ah sure, as you were.

pentschev:
Thanks @wence- for another review! 😄

PR #78, opened 2023-08-23
Process `ucxx::Endpoint::close()` in progress thread

pentschev:
Use generic callbacks to process `ucxx::Endpoint::close()` in the progress thread when enabled.

### DIFF
--- a/cpp/src/endpoint.cpp
+++ b/cpp/src/endpoint.cpp
@@ -160,15 +160,17 @@ void Endpoint::close()
     });
     callbackNotifierPre.wait([](auto flag) { return flag; });
 
-    while (UCS_PTR_STATUS(status) == UCS_INPROGRESS) {
+    while (UCS_PTR_IS_PTR(status)) {
       utils::CallbackNotifier callbackNotifierPost{false};
-      worker->registerGenericPost([&callbackNotifierPost, &status]() {
-        if (UCS_PTR_IS_PTR(status)) {
-          auto s = ucp_request_check_status(status);
-          if (s != UCS_INPROGRESS) ucp_request_free(status);
-        }
-        if (UCS_PTR_STATUS(status) != UCS_OK) {
-          ucxx_error("Error while closing endpoint: %s", ucs_status_string(UCS_PTR_STATUS(status)));
+      worker->registerGenericPost([this, &callbackNotifierPost, &status]() {
+        ucs_status_t s = ucp_request_check_status(status);
+        if (UCS_PTR_STATUS(s) != UCS_INPROGRESS) {
+          ucp_request_free(status);
+          _callbackData->status = status = UCS_PTR_STATUS(s);
+          if (UCS_PTR_STATUS(status) != UCS_OK) {
+            ucxx_error("Error while closing endpoint: %s",
+                       ucs_status_string(UCS_PTR_STATUS(status)));
+          }
         }
 
         callbackNotifierPost.store(true);
@@ -178,9 +180,11 @@ void Endpoint::close()
   } else {
     status = ucp_ep_close_nb(_handle, closeMode);
     if (UCS_PTR_IS_PTR(status)) {
-      while (ucp_request_check_status(status) == UCS_INPROGRESS)
+      ucs_status_t s;
+      while ((s = ucp_request_check_status(status)) == UCS_INPROGRESS)
         if (!worker->isProgressThreadRunning()) worker->progress();
       ucp_request_free(status);
+      _callbackData->status = s;
     } else if (UCS_PTR_STATUS(status) != UCS_OK) {
       ucxx_error("Error while closing endpoint: %s", ucs_status_string(UCS_PTR_STATUS(status)));
     }
--- a/cpp/src/endpoint.cpp
+++ b/cpp/src/endpoint.cpp
@@ -148,14 +148,42 @@ void Endpoint::close()
     // the endpoint status is not UCS_OK
     closeMode = UCP_EP_CLOSE_MODE_FORCE;
   }
-  ucs_status_ptr_t status = ucp_ep_close_nb(_handle, closeMode);
-  if (UCS_PTR_IS_PTR(status)) {
-    auto worker = ::ucxx::getWorker(_parent);
-    while (ucp_request_check_status(status) == UCS_INPROGRESS)
-      if (!worker->isProgressThreadRunning()) worker->progress();
-    ucp_request_free(status);
-  } else if (UCS_PTR_STATUS(status) != UCS_OK) {
-    ucxx_error("Error while closing endpoint: %s", ucs_status_string(UCS_PTR_STATUS(status)));
+
+  auto worker = ::ucxx::getWorker(_parent);
+  ucs_status_ptr_t status;
+
+  if (worker->isProgressThreadRunning()) {
+    utils::CallbackNotifier callbackNotifierPre{false};
+    worker->registerGenericPre([this, &callbackNotifierPre, &status, closeMode]() {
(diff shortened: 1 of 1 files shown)


cpp/src/endpoint.cpp
```diff
@@ -148,14 +148,42 @@ void Endpoint::close()
...
+
+        callbackNotifierPost.store(true);
+      });
+      callbackNotifierPost.wait([](auto flag) { return flag; });
+    }
+  } else {
+    status = ucp_ep_close_nb(_handle, closeMode);
+    if (UCS_PTR_IS_PTR(status)) {
+      while (ucp_request_check_status(status) == UCS_INPROGRESS)
+        if (!worker->isProgressThreadRunning()) worker->progress();
```
wence-, line 182: This condition must be true, so:

```suggestion
        worker->progress();
```
?

cpp/src/endpoint.cpp
```diff
@@ -148,14 +148,42 @@ void Endpoint::close()
...
+    callbackNotifierPre.wait([](auto flag) { return flag; });
+
+    while (UCS_PTR_STATUS(status) == UCS_INPROGRESS) {
+      utils::CallbackNotifier callbackNotifierPost{false};
+      worker->registerGenericPost([&callbackNotifierPost, &status]() {
+        if (UCS_PTR_IS_PTR(status)) {
+          auto s = ucp_request_check_status(status);
+          if (s != UCS_INPROGRESS) ucp_request_free(status);
+        }
+        if (UCS_PTR_STATUS(status) != UCS_OK) {
```
wence-, line 170:
```suggestion
        } else if (UCS_PTR_STATUS(status) != UCS_OK) {
```
?

cpp/src/endpoint.cpp
```diff
@@ -148,14 +148,42 @@ void Endpoint::close()
...
+
+  if (worker->isProgressThreadRunning()) {
+    utils::CallbackNotifier callbackNotifierPre{false};
+    worker->registerGenericPre([this, &callbackNotifierPre, &status, closeMode]() {
+      status = ucp_ep_close_nb(_handle, closeMode);
+      callbackNotifierPre.store(true);
+    });
+    callbackNotifierPre.wait([](auto flag) { return flag; });
+
+    while (UCS_PTR_STATUS(status) == UCS_INPROGRESS) {
```
wence-, line 163: Who can be updating `status`?

The pre-callback closes the endpoint, which produces some status. That "happens-before" this read because of the `wait` on the callbacknotifier.

Now we loop and fire off another callback which writes to `status` by calling `ucp_request_free` (which is the thing that will eventually turn `UCS_PTR_STATUS(status)` into `UCS_OK` I guess).

Since there is a `wait` in this loop, the `ucp_request_free` must "happen-before" the next iteration of the loop, so that read is safe. I wonder if there is some circumstance under which the write on line 168 can conflict with the read on line 163. I think yes, so we need:

```suggestion
    while (UCS_PTR_STATUS(status) == UCS_INPROGRESS) {
      utils::CallbackNotifier barrier{false};
      worker->registerGenericPre([&barrier]() { barrier.store(true); });
      barrier.wait([](auto flag) { return flag; });
```
To ensure correct sequencing.

wence-: Otherwise the read of `status` in this thread can in theory migrate past the write on the progress thread (unlikely, but possible).

wence- (changes requested):
I think we need one more barrier-like call in the progress-thread version unfortunately :(.

cpp/src/endpoint.cpp
```diff
@@ -148,14 +148,42 @@ void Endpoint::close()
...
+        callbackNotifierPost.store(true);
+      });
+      callbackNotifierPost.wait([](auto flag) { return flag; });
+    }
+  } else {
+    status = ucp_ep_close_nb(_handle, closeMode);
+    if (UCS_PTR_IS_PTR(status)) {
+      while (ucp_request_check_status(status) == UCS_INPROGRESS)
+        if (!worker->isProgressThreadRunning()) worker->progress();
+      ucp_request_free(status);
```
wence-, line 183: Can `ucp_request_free` leave the status in a bad state? AKA, should we check the status after free in this branch?

cpp/src/endpoint.cpp
pentschev: > Now we loop and fire off another callback which writes to `status` by calling `ucp_request_free` (which is the thing that will eventually turn `UCS_PTR_STATUS(status)` into `UCS_OK` I guess).

No, `ucp_ep_close_nb` behaves the same way as any UCX request (like a tag request, for example). What will happen is that it will either return an actual status if it completed immediately, otherwise it will return a pointer to a request. If it completed immediately we're done, and when we check if it's still in progress it will just never enter the loop. `UCS_PTR_STATUS(status)` will check what is the status of the underlying request, if it completed immediately it will be either `UCS_OK` or `UCS_ERR_*`, if not the request is still `UCS_INPROGRESS` which will cause us to enter the loop. At this point, we have to wait for the request to complete, so while `UCS_PTR_IS_PTR(status)` (meaning it's still a request pointer) we need then to query again whether the request completed or not in line 167 `auto s = ucp_request_check_status(status)`, and if it's not `UCS_INPROGRESS` anymore (it completed) we can then free the request pointer, which won't modify it. I did miss updating `status` with the new value returned by `s`.

`ucp_request_free` just releases a request pointer, if it is a request pointer, and it won't modify `status`. For a while I was myself confused as to the differences of `UCS_PTR_STATUS` and `ucp_request_check_status`, so I went asking for confirmation. `UCS_PTR_STATUS` will basically return the status if the argument is a `ucs_status_t` value, and if it's a pointer it will return `UCS_INPROGRESS`, whereas `ucs_request_check_status` will check whether the request has actually completed and then return its final status, or `UCS_INPROGRESS` if it didn't.Therefore, the only place where we should assign it is after we've checked and released `status` (to update that it is _not_ a pointer anymore but a status now). This should now be fixed via https://github.com/rapidsai/ucxx/pull/78/commits/0260a50da0d010e0a011a682a5dae3dc7f840755 .

cpp/src/endpoint.cpp
pentschev: That's right, we need to first check and then act on the result.

cpp/src/endpoint.cpp
pentschev: This is now fixed by keeping a copy of the actual status, freeing the `status` pointer and finally updating `status` with the status again, which will be checked in the next iteration.

pentschev (commented):
Thanks @wence- , I think I've addressed your suggestion and questions, please have another look when you have time.

pentschev:
@wence- any further comments here?

wence- (approved):
Thanks Peter, apologies for the delay

pentschev:
Thanks @wence- !

wence-:
Somehow, while 7a644e2 passed fine, the merge commit https://github.com/rapidsai/ucxx/pull/87/commits/bef4260d46ed304410ebfa2947aa567a37b3ef46 did not :(

pentschev:
> Somehow, while [7a644e2](https://github.com/rapidsai/ucxx/commit/7a644e25833a254f86cc7eeaa5ecd9eb9ba0fa34) passed fine, the merge commit [bef4260](https://github.com/rapidsai/ucxx/commit/bef4260d46ed304410ebfa2947aa567a37b3ef46) did not :(

Yes, but this is unrelated. Unfortunately it's also the same problem I mentioned in the second paragraph of https://github.com/rapidsai/ucxx/pull/78#issuecomment-1738860626. I'm still investigating that, but don't have many leads currently.

PR #85, opened 2023-09-26
Refactor CI port retry logic for C++ and retry examples

pentschev:
Refactor the CI port retry logic for C++ processes that bind to a specific port so that the code is reused for both benchmarks and examples, and enable examples as those were not covered by the retry logic.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,7 +27,7 @@ repos:
                 types: [cython]
                 additional_dependencies: ["flake8-force"]
       - repo: https://github.com/pre-commit/mirrors-clang-format
-        rev: v11.1.0
+        rev: v16.0.6
         hooks:
               - id: clang-format
                 files: \.(h|cpp)$
--- a/ci/test_cpp.sh
+++ b/ci/test_cpp.sh
@@ -30,7 +30,7 @@ run_tests() {
   CMD_LINE="UCX_TCP_CM_REUSEADDR=y timeout 10m ${BINARY_PATH}/gtests/libucxx/UCXX_TEST"
 
   log_command "${CMD_LINE}"
-  UCX_TCP_CM_REUSEADDR=y timeout 10m ${BINARY_PATH}/gtests/libucxx/UCXX_TEST
+  ${CMD_LINE}
 }
 
 run_benchmark() {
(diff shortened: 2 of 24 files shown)


wence- (approved):
Minor suggestions to remove some repetition, looks good!

pentschev:
> Minor suggestions to remove some repetition, looks good!

I can't recall details anymore, but I remember seeing some weird issues when expanding variables to execute them  (maybe how bash expands arguments within `""`), this is why I wasn't doing that. But let's see if it will pass like this now or something will go wrong.

wence-:
> > Minor suggestions to remove some repetition, looks good!
> 
> I can't recall details anymore, but I remember seeing some weird issues when expanding variables to execute them (maybe how bash expands arguments within `""`), this is why I wasn't doing that. But let's see if it will pass like this now or something will go wrong.

Seems like no, sorry! https://github.com/rapidsai/ucxx/actions/runs/6337172491/job/17212036065?pr=85#step:7:443

wence-:
And this looks odd https://github.com/rapidsai/ucxx/actions/runs/6337172491/job/17212037078?pr=85#step:7:1027

pentschev:
> Seems like no, sorry! https://github.com/rapidsai/ucxx/actions/runs/6337172491/job/17212036065?pr=85#step:7:443

This one in particular seems to be bash doesn't allow expanding variables as the command line, I [fixed that](https://github.com/rapidsai/ucxx/pull/85/commits/2b611cba59eff6d256d19700a20bf8e6a099fc74) and maybe it will pass. Let's see if there will be other issues besides this.

> And this looks odd https://github.com/rapidsai/ucxx/actions/runs/6337172491/job/17212037078?pr=85#step:7:1027

This is the same issue that has been plaguing CI for a month or two. It can only be reproduced in CentOS 7 (I didn't see a single instace of that elsewhere), I'm suspicious it could be either an old compiler or stdlib bug, but how does one prove that?!

pentschev:
> > And this looks odd https://github.com/rapidsai/ucxx/actions/runs/6337172491/job/17212037078?pr=85#step:7:1027
> 
> This is the same issue that has been plaguing CI for a month or two. It can only be reproduced in CentOS 7 (I didn't see a single instace of that elsewhere), I'm suspicious it could be either an old compiler or stdlib bug, but how does one prove that?!

Also, this seems to have intensified this week for some reason, https://github.com/rapidsai/ucxx/actions/runs/6316749172/job/17185069857 required 9 attempts to pass (versus a common 1-3 in worst cases).

raydouglass (approved):

pentschev:
Thanks @wence- and @raydouglass !

PR #86, opened 2023-09-27
Fix shutdown of UCX getters Python async tests

pentschev:


python/ucxx/_lib_async/tests/test_ucx_getters.py
```diff
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 import pytest
+from utils import wait_listener_client_handlers
```
wence-, line 5: **nitpick (non-blocking):** ‏This is a bit weird and is relying (I think) on pytest's behaviour of inserting the current directory into `sys.path`. Is there a way this `utils` file can be moved into the ucxx package namespace proper so that this becomes a fully-qualified import? Or perhaps:

```suggestion
from .utils import wait_listener_client_handlers
```

Will work?

I worry that `utils` might be a global package (badly named, but ...)

wence- (approved):
Minor nit.

python/ucxx/_lib_async/tests/test_ucx_getters.py
pentschev: No, that won't work. For `.utils` to work it would either need to be in the current directory or in the installed module, but since tests are not an installed module it won't work.

In any case this is a valid concern, I now addressed it in https://github.com/rapidsai/ucxx/pull/86/commits/88b418a7cc482c7e09b192add74695b20b9c804e .

wence-: Thanks (sorry, a bigger fix than I was anticipating).

pentschev: Just a bunch of one-liner import changes, I wish changes were always like that! 🤣

pentschev:
Thanks @wence- . After some fourty-twelve-zillion-trillion attempts CI passed. I'm gonna go ahead and merge this and continue to try and fix the CI issue in https://github.com/rapidsai/ucxx/pull/89 and potentially further PRs.

PR #89, opened 2023-09-28
Separate `ucxx::DelayedSubmissionCollection` implementations

pentschev:
Separate `ucxx::DelayedSubmissionCollection` implementations for requests and generic callbacks, thus reducing some code duplication and making collection logic more self-contained.

This may help tackling issues that are only reproducible in CentOS 7.


pentschev:
We still see segfaults and deadlocks. Probably a good change from an organization perspective, but doesn't fix the issues.

pentschev:
What do you think of this change @wence- ? Better or worse than before?

cpp/src/delayed_submission.cpp
```diff
@@ -34,92 +76,27 @@ bool DelayedSubmissionCollection::isDelayedRequestSubmissionEnabled() const
...
-  if (callbacks.size() > 0) {
-    ucxx_trace_req("Submitting %lu generic", callbacks.size());
-
-    for (auto& callback : callbacks) {
-      ucxx_trace_req("Submitting generic");
-
-      if (callback) callback();
-    }
-  }
+  _genericPre.process();
```
wence-, line 81: **thought (non-blocking):** ‏ While I was reading this I was worried that these two objects were processed in the wrong order, but then I went back over the code and remembered that this is correct. The `genericPre` callbacks are to pass information between threads that the requests have been processed IIRC. Perhaps this is worth documenting here as a comment (or in the docstring?)

cpp/include/ucxx/delayed_submission.h
```diff
@@ -56,15 +57,135 @@ class DelayedSubmission {
...
+   * @brief Register a callable or complex-type for delayed submission.
+   *
+   * Register a simple callback, or complex-type with a callback (requires specialization),
+   * for delayed submission that will be executed when the request is in fact submitted when
+   * `process()` is called.
+   *
+   * @throws std::runtime_error if delayed request submission was disabled at construction.
+   *
+   * @param[in] item      the callback that will be executed by `process()` when the
+   *                      operation is submitted.
```
wence-, line 88: **nitpick (non-blocking):** ‏Document `contextInitialized`?

cpp/include/ucxx/delayed_submission.h
```diff
@@ -56,15 +57,135 @@ class DelayedSubmission {
                     const ucs_memory_type_t memoryType = UCS_MEMORY_TYPE_UNKNOWN);
 };
 
+template <typename T>
+class DelayedSubmissionCollectionTemplateSafe {
```
wence-, line 61: **nitpick (non-blocking):** ‏I'm not sure this name is the best. The fact it's templated is given by the type. And it's not really clear what "safe" means in this context. How about `BaseDelayedSubmissionCollection`?

wence- (commented):
I think this is reasonable (modulo I don't super-love the naming). I'm kind of a bit on the fence as to whether it improves readability.

cpp/include/ucxx/delayed_submission.h
pentschev: [Done](https://github.com/rapidsai/ucxx/pull/89/commits/c08cf4976cbc4a68e158eeeb15184550b1a0fa9f) along with some improvements.

cpp/src/delayed_submission.cpp
pentschev: That's a good idea, [done](https://github.com/rapidsai/ucxx/pull/89/commits/a679e2fe37632a49fce2f2e91215e79b359394e3).

pentschev (commented):
Thanks @wence- for another great review. I think I addressed everything, please take another look when you have a chance.

pentschev:
@wence- anything else here or do you think we're good merging this?

wence- (approved):
All looks good, thanks Peter!

pentschev:
Thanks @wence- for the review!

PR #93, opened 2023-10-02
Simplify callback notifier and provide (optional) spin-lock implementation

wence-:
Rather than passing information through the callback notifier object, merely use it as a barrier for memory ordering. Additionally, provide an alternate implementation (automatically enabled when running with glibc <= 2.25) that uses an atomic spinlock rather than condition variables.

PR #94, opened 2023-10-02
Properly check for `ucp_ep_create()` errors

pentschev:
The `UCS_INPROGRESS` state is not valid for `ucp_ep_create`, only `UCS_OK` and errors are returned. Therefore remove that check and additionally fix to check for the proper error, instead of the previous boolean error check.

### DIFF
--- a/cpp/src/endpoint.cpp
+++ b/cpp/src/endpoint.cpp
@@ -61,14 +61,15 @@ Endpoint::Endpoint(std::shared_ptr<Component> workerOrListener,
   params->err_handler.arg = _callbackData.get();
 
   if (worker->isProgressThreadRunning()) {
-    utils::CallbackNotifier callbackNotifier{UCS_INPROGRESS};
+    utils::CallbackNotifier callbackNotifier{false};
     auto worker = ::ucxx::getWorker(_parent);
-    worker->registerGenericPre([this, &params, &callbackNotifier]() {
+    auto status = UCS_OK;
+    worker->registerGenericPre([this, &status, &params, &callbackNotifier]() {
       auto worker = ::ucxx::getWorker(_parent);
-      auto status = ucp_ep_create(worker->getHandle(), params, &_handle);
-      callbackNotifier.store(status);
+      status      = ucp_ep_create(worker->getHandle(), params, &_handle);
+      callbackNotifier.store(true);
     });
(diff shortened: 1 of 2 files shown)


cpp/src/endpoint.cpp
```diff
@@ -61,14 +61,15 @@ Endpoint::Endpoint(std::shared_ptr<Component> workerOrListener,
...
-    worker->registerGenericPre([this, &params, &callbackNotifier]() {
+    auto status = UCS_INPROGRESS;
+    worker->registerGenericPre([this, &status, &params, &callbackNotifier]() {
       auto worker = ::ucxx::getWorker(_parent);
-      auto status = ucp_ep_create(worker->getHandle(), params, &_handle);
-      callbackNotifier.store(status);
+      status      = ucp_ep_create(worker->getHandle(), params, &_handle);
+      callbackNotifier.store(true);
     });
-    auto status = callbackNotifier.wait([](auto status) { return status != UCS_INPROGRESS; });
```
pentschev, line 71: Actually here `UCS_INPROGRESS` was possible, because that was the initial status in line 64.

cpp/src/endpoint.cpp
```diff
@@ -61,14 +61,15 @@ Endpoint::Endpoint(std::shared_ptr<Component> workerOrListener,
   params->err_handler.arg = _callbackData.get();
 
   if (worker->isProgressThreadRunning()) {
-    utils::CallbackNotifier callbackNotifier{UCS_INPROGRESS};
+    utils::CallbackNotifier callbackNotifier{false};
     auto worker = ::ucxx::getWorker(_parent);
-    worker->registerGenericPre([this, &params, &callbackNotifier]() {
+    auto status = UCS_INPROGRESS;
```
pentschev, line 66: I've also made the initial `status = UCS_INPROGRESS`. This is not something that `ucp_ep_create` will return, but we can use it ensure that it has changed after `ucp_ep_create` is called (we want either `UCS_OK` or `UCS_ERR_*`). In essence, the original code was correct w.r.t. to the `UCS_INPROGRESS` comparison, but the `ucsErrorThrow(status)` was wrong, since it was being cast down to a `bool`.

pentschev:
Closing in favor of #93 that also incorporates this fix.

PR #95, opened 2023-10-06
WIP: Only one notifier per worker

wence-:
Let's see if this works...


wence-:
No, I think this won't work...

pentschev:
For something like this to work I think you'd still require a pool of `condition_variable`s to provide ordering and synchronization (i.e., not being notified immediately because of some processing for another notifier has finished immediately after you waited on it). In that case we could at least reuse the `condition_variable`s, if that's really something that is expensive to construct.

PR #98, opened 2023-10-16
Exchange peer info fixes

pentschev:
Add a timeout to `exchange_peer_info` to prevent indefinite hangs in `exchange_peer_info`, specifically when executing `stream_recv` on the client side. The causes for this is unknown but believed to be due to messages being lost if there's either multiple stream messages being transferred simultaneously among various endpoints or being lost due to the receiving end taking too long to launch `stream_recv`, see https://github.com/rapidsai/ucx-py/pull/509 for a similar issue related to stream API. By adding a timeout doesn't allow recovery, but at least allows a UCX-Py client to retry upon failure to establish the endpoint.

Additionally, raise endpoint's error if an error occurs during `exchange_peer_info`.

### DIFF
--- a/python/ucxx/_lib_async/application_context.py
+++ b/python/ucxx/_lib_async/application_context.py
@@ -9,6 +9,7 @@ from queue import Queue
 
 import ucxx._lib.libucxx as ucx_api
 from ucxx._lib.arr import Array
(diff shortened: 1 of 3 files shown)


wence- (approved):

pentschev:
Thanks for the review here Lawrence!

PR #100, opened 2023-10-16
Prevent `_run_request_notifier()` uncatched exceptions

pentschev:
If an exception is raised during `_run_request_notifier()`, we should catch it, cancel the task and raise a debug log message, as those are only expected to occur during shutdown.

### DIFF
--- a/python/ucxx/_lib_async/notifier_thread.py
+++ b/python/ucxx/_lib_async/notifier_thread.py
@@ -12,7 +12,7 @@ logger = logging.getLogger("ucx")
 
 
 async def _run_request_notifier(worker):
-    worker.run_request_notifier()
+    return worker.run_request_notifier()
 
 
 async def _notifier_coroutine(worker):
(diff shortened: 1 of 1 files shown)


python/ucxx/_lib_async/notifier_thread.py
```diff
@@ -58,4 +58,7 @@ def _notifierThread(event_loop, worker, q):
         try:
             task.result(0.01)
         except TimeoutError:
+            task.cancel()
```
wence-, line 61: **question:** Can this cancellation raise an error, and do we need to catch that too, or is there nothing we can do in that circumstance?

wence- (approved):

python/ucxx/_lib_async/notifier_thread.py
pentschev: My understanding is that when scheduling from another thread, `cancel()` will not raise as that needs to be `awaited`, see how the `[asyncio.run_coroutin_threadsafe()` docs](https://docs.python.org/3/library/asyncio-task.html#task-cancellation) have an almost verbatim example.

pentschev:
Thanks @wence- !

PR #101, opened 2023-10-16
Do a roundtrip message transfer for `test_send_recv_am`

pentschev:
A roundtrip ensures the client receives the reply and thus prevents us from the checking for a transfer that didn't complete yet.

### DIFF
--- a/python/ucxx/_lib_async/tests/test_send_recv_am.py
+++ b/python/ucxx/_lib_async/tests/test_send_recv_am.py
@@ -77,21 +77,21 @@ async def test_send_recv_am(size, recv_wait, data):
         await ucxx.create_endpoint(ucxx.get_address(), listener.port)
         for i in range(num_clients)
     ]
-    for c in clients:
-        if recv_wait:
-            # By sleeping here we ensure that the listener's
-            # ep.am_recv call will have to wait, rather than return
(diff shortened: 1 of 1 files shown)


python/ucxx/_lib_async/tests/test_send_recv_am.py
```diff
@@ -83,16 +84,14 @@ async def test_send_recv_am(size, recv_wait, data):
             # immediately as receive data is already available.
             await asyncio.sleep(1)
         await c.am_send(msg)
-
-    while len(recv) == 0:
-        await asyncio.sleep(0)
+        recv_msg = await c.am_recv()
```
wence-, line 87: **polish:** ‏This waits for the result from every client, but overwrites with the most recent result. So where previously we checked only the first result, now we check only the last result.

Now, there's only a single client, so we're actually checking all the results. But perhaps we should make this test correct for more than one client. WDYT?

pentschev: Good catch, fixed in https://github.com/rapidsai/ucxx/pull/101/commits/7a0b1c3436e1b8a0cc92dac744312dc2a04228fd .

wence- (approved):

pentschev:
Thanks @wence- !

PR #102, opened 2023-10-16
Add UCP handle to create/destroy trace logging

pentschev:
Add the UCP handle to trace logging, thus making a distinction between the UCXX object (`this`) pointer and the underlying UCP handle, allowing to more easily match objects during debugging sessions.

### DIFF
--- a/cpp/src/context.cpp
+++ b/cpp/src/context.cpp
@@ -22,8 +22,8 @@ Context::Context(const ConfigMap ucxConfig, const uint64_t featureFlags)
   // UCP
   ucp_params_t params = {.field_mask = UCP_PARAM_FIELD_FEATURES, .features = featureFlags};
 
-  utils::ucsErrorThrow(ucp_init(&params, this->_config.getHandle(), &this->_handle));
-  ucxx_trace("Context created: %p", this->_handle);
+  utils::ucsErrorThrow(ucp_init(&params, _config.getHandle(), &_handle));
+  ucxx_trace("Context created: %p, UCP handle: %p", this, _handle);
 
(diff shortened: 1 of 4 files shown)


cpp/src/context.cpp
```diff
@@ -72,18 +72,18 @@ std::shared_ptr<Context> createContext(const ConfigMap ucxConfig, const uint64_t
 
 Context::~Context()
 {
-  if (this->_handle != nullptr) ucp_cleanup(this->_handle);
-  ucxx_trace("Context destroyed: %p", this->_handle);
+  if (_handle != nullptr) ucp_cleanup(_handle);
+  ucxx_trace("Context destroyed: %p, UCP handle: %p", this, _handle);
```
wence-, line 76: **polish:** ‏While we're here:

```suggestion
  if (_handle != nullptr) ucp_cleanup(_handle);
  ucxx_trace("Context destroyed: %p, UCP handle: %p", this, _handle);
  _handle = nullptr;
```

And similarly in the other destructors.

Or does this not matter because the object is about to disappear anyway, so we can't access `_handle` after `ucp_cleanup` is called.

wence- (approved):
Possibly it is worthwhile setting the member variable `_handle`s to `nullptr in the destructors, but I don't think it matters that much.

cpp/src/context.cpp
pentschev: Setting it to `nullptr` is unnecessary, I'd even argue bad practice (why not then reset/clear _every_ member?). It is undefined behavior to access `_handle` or any other non-shared member during/after destructor runs.

pentschev:
Thanks @wence- !

PR #108, opened 2023-10-20
Add compiler('cuda') to ucxx builds.

bdice:
Some build failures have been occurring on CUDA 12. This PR fixes the builds by adding a build dependency on `compiler('cuda')` for CUDA 12 in the ucxx package.

Also includes changes from #107.

```
-- Found Threads: TRUE  
CMake Error at /opt/conda/conda-bld/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib/cmake/rmm/rmm-config.cmake:157 (thrust_create_target):
  Unknown CMake command "thrust_create_target".
Call Stack (most recent call first):
  /opt/conda/conda-bld/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib/cmake/ucxx/ucxx-dependencies.cmake:23 (find_package)
  /opt/conda/conda-bld/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib/cmake/ucxx/ucxx-config.cmake:81 (include)
  CMakeLists.txt:28 (find_package)


-- Configuring incomplete, errors occurred!
Traceback (most recent call last):
  File "/opt/conda/conda-bld/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib/python3.10/site-packages/skbuild/setuptools_wrap.py", line 666, in setup
    env = cmkr.configure(
  File "/opt/conda/conda-bld/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib/python3.10/site-packages/skbuild/cmaker.py", line 357, in configure
    raise SKBuildError(msg)
```

PR #109, opened 2023-10-20
Update versions/recipe logic.

bdice:
Some improvements / modernizations for the package and dependencies. Depends on #108.


python/pyproject.toml
```diff
@@ -41,14 +42,16 @@ classifiers = [
...
     "pytest",
     "pytest-asyncio",
-    "cloudpickle",
-    "cudf",
-    "dask",
-    "distributed",
-    "cupy-cuda11x",
-]
+    "pytest-rerunfailures",
+] # This list was generated by `rapids-dependency-file-generator`. To make changes, edit ../dependencies.yaml and run `rapids-dependency-file-generator`.
```
pentschev, line 54: How did `rapids-dependency-file-generator` come up with this list? In particular the Dask/Distributed pins seem very arbitrary.

bdice: These are coming from `dependencies.yaml`. That file is supposed to be a "single source of truth" for all dependencies and their pinnings (with the exception of meta.yaml conda recipes and CMakeLists.txt which are not implemented at this time). Today, `rapids-dependency-file-generator` generates `pyproject.toml`, developer conda environments, and can generate dependency lists on-the-fly that aren't committed to the repo but are used during various builds (e.g. devcontainers, rapids-compose).

bdice: I just noticed that `conda/environments/ucxx-cuda118_arch-x86_64.yaml` is not being generated by `dependencies.yaml` so I'll try to adjust that.

bdice: There are some features like channel pinnings (`dask/label/dev::dask`) and comments that would be lost in the developer environment. @pentschev Do you have thoughts on whether we should rework this to be auto-generated like other RAPIDS packages or would you prefer to maintain it separately?

bdice: See what you think of commit 4fc2c974dbff425f73aaf711fbf22dad6571d113. I added a `dev` section with the dependencies like `libtool` and `autoconf` and `dask-cuda` that aren't otherwise specified by ucxx as build/run dependencies.

dependencies.yaml
```diff
@@ -112,13 +167,14 @@ dependencies:
...
   test_python:
     common:
-      - output_types: [conda, requirements]
+      - output_types: [conda, requirements, pyproject]
         packages:
+          - cloudpickle
           - cudf==23.12.*
           - cupy
           - dask>=2023.1.1
           - distributed>=2023.1.1
```
pentschev, line 176:
```suggestion
          - dask
          - distributed
```

pentschev: Let's make this [consistent with UCX-Py](https://github.com/rapidsai/ucx-py/blob/2a22a67a4abbafe3facfd5348030b272dfacfa5b/dependencies.yaml#L120-L121) for now. I'll have to review how we pin to Dask, whether we should follow the same practice to pin/unpin as we do for the rest of RAPIDS or otherwise.

python/pyproject.toml
pentschev: > These are coming from `dependencies.yaml`. That file is supposed to be a "single source of truth" for all dependencies and their pinnings (with the exception of meta.yaml conda recipes and CMakeLists.txt which are not implemented at this time). Today, `rapids-dependency-file-generator` generates `pyproject.toml`, developer conda environments, and can generate dependency lists on-the-fly that aren't committed to the repo but are used during various builds (e.g. devcontainers, rapids-compose).

Sorry, this was entirely my mistake. I knew it was coming from `dependencies.yaml` but I didn't realize it was being pinned to `>=2023.1.1` there. I'm guessing when this was last adjusted `python/pyproject.toml` was not regenerated so I got confused by this seemingly arbitrary versioning.

pentschev: > There are some features like channel pinnings (`dask/label/dev::dask`) and comments that would be lost in the developer environment. @pentschev Do you have thoughts on whether we should rework this to be auto-generated like other RAPIDS packages or would you prefer to maintain it separately?

I guess that has been addressed by https://github.com/rapidsai/ucxx/commit/4fc2c974dbff425f73aaf711fbf22dad6571d113, is that right? If so I think this question isn't relevant anymore.

> See what you think of commit [4fc2c97](https://github.com/rapidsai/ucxx/commit/4fc2c974dbff425f73aaf711fbf22dad6571d113). I added a `dev` section with the dependencies like `libtool` and `autoconf` and `dask-cuda` that aren't otherwise specified by ucxx as build/run dependencies.

I see that it included `dask/label/dev` as well which is great and addresses your question from above, these changes are perfect, thanks for doing that Bradley!

pentschev (changes requested):
These changes are awesome, thanks @bdice for the work here. I've left a minor request to make versioning consistent with UCX-Py for the moment, let me know if you have thoughts.

dependencies.yaml
```diff
@@ -82,6 +119,21 @@ dependencies:
...
+          # RAPIDS
+          - pip
+          - dask-cuda==23.12.*
+          - dask-cudf==23.12.*
+          # UCX Build
+          - libtool
+          - automake
+          - autoconf
+          # UCXX Build
+          - pkg-config
```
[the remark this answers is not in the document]
bdice: The `dev` section is for dependencies that were added by hand to the previous conda environment (perhaps by you?) but do not fall into one of the other categories like build-time dependencies or run-time dependencies. I didn't want to lose that information but (if I understand correctly) none of the packages in `dev` are needed for CI to build and run tests.

pentschev (approved):
Thanks a lot @bdice !

raydouglass (approved):

PR #110, opened 2023-10-25
Ignore UCX run exports and allow `ucx>=1.14.1,<1.16.0`

pentschev:
Build UCX with `ucx==1.14.*` and allow running with `ucx>=1.14.1,<1.16.0` which should all be ABI-compatible. This is required to allow resolution based on requirements from some dependencies (e.g., `libarrow`), while still providing a path to rollback for short-term testing.


raydouglass (approved):

pentschev:
> The pinning looks right, but there's a failing C++ test. Any idea why?

Some of the tests fail at times, I'm slowly trying to get them resolved but unfortunately they're often hard to reproduce locally, making them even more difficult. 😞

PR #113, opened 2023-10-30
Add argument to specify exchange peer info timeout

pentschev:
Under certain circumstances the current `stream_timeout` in `exchange_peer_info` may be too low, for example in `test_multiple_nodes` where hundreds of endpoints are being created simultaneously, thus causing those to timeout.

To prevent this, expose a new user-facing keyword argument `exchange_peer_info_timeout` to `create_endpoint`/`create_listener` that can be adjusted depending on the expected load of the application.


python/ucxx/_lib_async/application_context.py
```diff
@@ -201,6 +205,13 @@ def create_listener(
             but prevents a process from terminating unexpectedly that may
             happen when disabled. If `False` endpoint endpoint error handling
             is disabled.
+        exchange_peer_info_timeout: float
+            Timeout for exchanging peer info. In some cases, exchanging peer
+            information may hang indefinitely, a timeout prevents that. If the
+            is too high it may cause the operation to be stuck for too long
+            other than quickly raising a `TimeoutError` that may be recovered
+            from by the application, but under high-load a higher timeout may
+            be helpful to prevent exchanging peer info from failing too fast.
```
wence-, line 214:
```suggestion
            information may hang indefinitely, a timeout prevents that. If the
            chosen value is too high it may cause the operation to be stuck for too long
            rather than quickly raising a `TimeoutError` that may be recovered
            from by the application, but under high-load a higher timeout may
            be helpful to prevent exchanging peer info from failing too fast.
```

Modulo line-wrapping.

python/ucxx/_lib_async/application_context.py
```diff
@@ -251,6 +269,13 @@ async def create_endpoint(self, ip_address, port, endpoint_error_handling=True):
...
```
wence-, line 278: Same comment here.

python/ucxx/_lib_async/application_context.py
```diff
@@ -201,6 +205,13 @@ def create_listener(
             but prevents a process from terminating unexpectedly that may
             happen when disabled. If `False` endpoint endpoint error handling
             is disabled.
+        exchange_peer_info_timeout: float
+            Timeout for exchanging peer info. In some cases, exchanging peer
```
wence-, line 209:
```suggestion
            Timeout (in seconds) for exchanging peer info. In some cases, exchanging peer
```

Let's indicate what `10` means.

python/ucxx/_lib_async/application_context.py
```diff
@@ -251,6 +269,13 @@ async def create_endpoint(self, ip_address, port, endpoint_error_handling=True):
...
```
wence-, line 273: ibid.

wence- (approved):
Tiny wording suggestions

pentschev:
Thanks @wence- for the suggestions, I've applied them now, please let me know if you find any errors.

pentschev:
Thanks @wence- for yet another review!

PR #114, opened 2023-10-31
Remove versioneer and fix `ucxx` pin in `distributed-ucxx`

pentschev:
Remove versioneer in favor of RAPIDS update scripts, fix `ucxx` pin which had `exact=True` pin causing the `distributed-ucxx` version to build against a specific CUDA version and making the package uninstallable elsewhere, and pin Dask/Distributed versions.


raydouglass (approved):

madsbk (approved):
LGTM, thanks @pentschev

PR #115, opened 2023-10-31
Replace `std::string_view` usage with `std::string`

pentschev:
A `std::string_view` doesn't guarantee null-terminated strings and thus cannot be safely used with UCX and UCXX loggers. Replace `std::string_view` with `std::string` to guarantee safe conversion.


madsbk (approved):
Good catch :)

pentschev:
Thanks @madsbk for reviewing! 😄

PR #116, opened 2023-11-02
Add blocking progress mode to Python async

pentschev:
Implements the blocking progress mode (UCX-Py default), which was still not implemented in UCXX.


cpp/include/ucxx/worker.h
```diff
@@ -245,6 +245,23 @@ class Worker : public Component {
...
 
+  /**
+   * @brief Get the epoll file descriptor associated with the worker.
+   *
+   * Get the epoll file descriptor associated with the worker when running in blocking mode.
+   * The worker only has an associated epoll file descriptor after
+   * `initBlockingProgressMode()` is executed.
+   *
+   * The file descriptor is destroyed as part of the `ucxx::Worker` destructor, thus any
+   * reference to it shall not be used after that.
```
wence-, line 256: **polish/suggestion:** How about we reset the `_epollFileDescriptor` slot to `0` in `~Worker` after calling `close` on it. That way it is obvious if someone attempts to use a bad object.

python/ucxx/_lib_async/continuous_ucx_progress.py
```diff
@@ -70,3 +72,50 @@ async def _progress_task(self):
...
+
+
+class BlockingMode(ProgressTask):
+    def __init__(self, worker, event_loop, epoll_fd):
+        super().__init__(worker, event_loop)
+
+        # Creating a job that is ready straightaway but with low priority.
+        # Calling `await self.event_loop.sock_recv(self.rsock, 1)` will
+        # return when all non-IO tasks are finished.
+        # See <https://stackoverflow.com/a/48491563>.
```
wence-, line 84: Just a reminder that this approach (as mentioned in that updated answer) unfortunately doesn't work :(. I tried to figure out a way last time we noticed this, but didn't manage to.

python/ucxx/_lib_async/continuous_ucx_progress.py
```diff
@@ -70,3 +72,50 @@ async def _progress_task(self):
...
+        # since previous arm task is finished by now.
+        assert self.asyncio_task is None or self.asyncio_task.done()
+        self.asyncio_task = self.event_loop.create_task(self._arm_worker())
+
+    async def _arm_worker(self):
+        # When arming the worker, the following must be true:
+        #  - No more progress in UCX (see doc of ucp_worker_arm())
+        #  - All asyncio tasks that isn't waiting on UCX must be executed
+        #    so that the asyncio's next state is epoll wait.
+        #    See <https://github.com/rapidsai/ucx-py/issues/413>
```
wence-, line 110: Although I think I understand the constraint, I am not sure what it means for the "next state" to be epoll_wait. Surely there can be arbitrary non-ucx tasks?

python/ucxx/_lib_async/continuous_ucx_progress.py
```diff
@@ -70,3 +72,50 @@ async def _progress_task(self):
...
+        while True:
+            self.worker.progress()
+
+            # This IO task returns when all non-IO tasks are finished.
+            # Notice, we do NOT hold a reference to `worker` while waiting.
+            await self.event_loop.sock_recv(self.rsock, 1)
+
+            if self.worker.arm():
+                # At this point we know that asyncio's next state is
+                # epoll wait.
```
wence-, line 120: Who does the `epoll_wait` call?

wence- (commented):
I think the core of this looks fine, but I am reminded that this `epoll_wait` issue with asyncio doesn't actually quite work correctly with the code we are using (see discussion here https://github.com/rapidsai/ucx-py/issues/888)

pentschev:
> I think the core of this looks fine, but I am reminded that this `epoll_wait` issue with asyncio doesn't actually quite work correctly with the code we are using (see discussion here [rapidsai/ucx-py#888](https://github.com/rapidsai/ucx-py/issues/888))

Yes, I remember that. The purpose of this is not necessarily to be used long-term, but rather to have a fallback to original UCX-Py behavior. This will allow us testing as close as possible with what UCX-Py did in the past, and we may later deprecate/remove this if we are confident we have a better option (e.g., `thread` progress mode).

python/ucxx/_lib_async/continuous_ucx_progress.py
pentschev: Indeed, see [my reply to the "main" review comment](https://github.com/rapidsai/ucxx/pull/116#issuecomment-1795336173).

python/ucxx/_lib_async/continuous_ucx_progress.py
pentschev: I'm gonna be honest and say my understanding here is also a bit fuzzy, and as you noted yourself this "doesn't work" (except when it does), with the original being adapted from https://stackoverflow.com/a/48491563. In my understanding, what `epoll_wait` refers to here is the [socket state](https://github.com/rapidsai/ucxx/pull/116/files#diff-1b56e195b331ac6ca2c996f330a9392fddaf6f124da76d3c290b97cabd57d2daR85-R88), which is there solely to provide a mechanism to prevent `asyncio` from running out of "ready" tasks, so `epoll_wait` will ensure that if nothing useful happens in the event loop, `asyncio` will still be awaken at some point to allow the loop to reevaluate.

python/ucxx/_lib_async/continuous_ucx_progress.py
pentschev: As per my https://github.com/rapidsai/ucxx/pull/116#discussion_r1383603225, I believe this is related to the [sockets](https://github.com/rapidsai/ucxx/pull/116/files#diff-1b56e195b331ac6ca2c996f330a9392fddaf6f124da76d3c290b97cabd57d2daR85-R88).

cpp/include/ucxx/worker.h
wence-: I think that's OK. In Python we always go through the property which needs a live worker to access things. I agree that the user should be on the hook for managing the lifetime of arbitrary untyped pointer-like handles and making sure invariants are satisfied, so nothing to do here.

wence- (approved):
Thanks for the explanations!

ajschmidt8 (approved):

raydouglass (approved):

python/ucxx/ucxx/_lib_async/continuous_ucx_progress.py
```diff
@@ -70,3 +77,133 @@ async def _progress_task(self):
...
+        self,
+        worker: UCXWorker,
+        event_loop: asyncio.AbstractEventLoop,
+        progress_timeout: float = 1.0,
+    ):
+        """Progress the UCX worker in blocking mode.
+
+        The blocking progress mode ensure the worker is progress whenever the UCX
+        worker reports an event on its epoll file descriptor. In certain
+        circumstances the epoll file descriptor may not
```
wence-, line 93: *nit*: Grammar suggestion:

> Blocking progress mode ensures that the worker is progressed whenever the UCX worker reports and event on its epoll file descriptor.

*nit8: The second sentence appears to be incomplete.

wence- (approved):
I had some small docstring comments but I still think this looks good, is there anything else to do here?

python/ucxx/ucxx/_lib_async/continuous_ucx_progress.py
```diff
@@ -24,12 +29,19 @@ def __init__(self, worker, event_loop):
...
+            if not self.event_loop.is_closed() and self.asyncio_task is not None:
+                try:
+                    self.asyncio_task.cancel()
+                    self.event_loop.run_until_complete(self.asyncio_task)
+                except asyncio.exceptions.CancelledError:
+                    pass
+                finally:
+                    event_loop_close(*args, **kwargs)
+
+        self.event_loop.close = _event_loop_close
```
pentschev, line 44: @wence- would you mind having one more look at this? It is a _real solution_ for the years long `coroutine was never awaited`/`Task was destroyed but it is pending!` warning that we've attempted to resolve in many instances, including https://github.com/rapidsai/ucx-py/pull/929, yet it is very intrusive. To me it doesn't look like it can be too harmful, but maybe you'll have some other thoughts or opinions.

python/ucxx/ucxx/_lib_async/continuous_ucx_progress.py
```diff
@@ -24,12 +29,19 @@ def __init__(self, worker, event_loop):
...
-            # Running with polling mode will always cause
-            # `Task was destroyed but it is pending!` errors at ucxx.reset().
-            self.asyncio_task.cancel()
+        event_loop_close = self.event_loop.close
+
+        def _event_loop_close(*args, **kwargs):
+            if not self.event_loop.is_closed() and self.asyncio_task is not None:
+                try:
+                    self.asyncio_task.cancel()
+                    self.event_loop.run_until_complete(self.asyncio_task)
```
wence-, line 38: *suggestion*: should we set `self.asyncio_task = None` after running until complete?

wence-: OK, so the idea is that you don't have control over who is closing the event loop, so you instead hook into `close` and this task cancels itself during event loop closing/teardown?

python/ucxx/ucxx/_lib_async/continuous_ucx_progress.py
wence-: Some questions:

- This changes the behaviour of `EventLoop.close`. Does it only do so for this instance?
- What happens if multiple of these `ProgressTask`s are stacked up, I guess each one overwrites the `close` method, but remembers the previous one, so we do unwind everything?

python/ucxx/ucxx/_lib_async/continuous_ucx_progress.py
```diff
@@ -70,3 +82,112 @@ async def _progress_task(self):
...
+
+    def _fd_reader_callback(self):
+        """Schedule new progress task upon worker event.
+
+        Schedule new progress task when a new event occurs in the worker's epoll file
+        descriptor.
+        """
+        self.worker.progress()
+
+        # Notice, we can safely overwrite `self.dangling_arm_task`
```
wence-, line 145:
```suggestion
        # Notice, we can safely overwrite `self.blocking_asyncio_task`
```
?

python/ucxx/ucxx/_lib_async/continuous_ucx_progress.py
```diff
@@ -70,3 +82,112 @@ async def _progress_task(self):
...
+
+    async def _arm_worker(self):
+        """Progress the worker and rearm.
+
+        Progress and rearm the worker to watch for new events on its epoll file
+        descriptor.
+        """
+        # When arming the worker, the following must be true:
+        #  - No more progress in UCX (see doc of ucp_worker_arm())
+        #  - All asyncio tasks that isn't waiting on UCX must be executed
```
wence-, line 158:
```suggestion
        #  - All asyncio tasks that aren't waiting on UCX must be executed
```

python/ucxx/ucxx/_lib_async/continuous_ucx_progress.py
pentschev: > _suggestion_: should we set `self.asyncio_task = None` after running until complete?

That's a good idea, done in https://github.com/rapidsai/ucxx/pull/116/commits/279cb4c0b4d11fa9df6df29aaf4083e27e30018e .

> OK, so the idea is that you don't have control over who is closing the event loop, so you instead hook into `close` and this task cancels itself during event loop closing/teardown?

Exactly, there doesn't seem to be another way, since we have no control of whether the user will close the loop before resetting UCXX.

python/ucxx/ucxx/_lib_async/continuous_ucx_progress.py
pentschev: > * This changes the behaviour of `EventLoop.close`. Does it only do so for this instance?

That's right, the change only applies to `self.event_loop`, not the whole class, here's an example:

```python
import asyncio


async def run_patch():
    loop = asyncio.get_running_loop()
    print(f"run_patch: {loop}")

    loop_close = loop.close

    def _patch_close(*args, **kwargs):
        if not loop.is_closed():
            print("_patch_close")
            loop_close(*args, **kwargs)

    loop.close = _patch_close


async def run_orig():
    loop = asyncio.get_running_loop()
    print(f"run_orig: {loop}")


loop = asyncio.new_event_loop()
loop.run_until_complete(run_patch())
loop.close()

loop2 = asyncio.new_event_loop()
loop2.run_until_complete(run_orig())
loop2.close()
```

Which prints:

```
run_patch: <_UnixSelectorEventLoop running=True closed=False debug=False>
_patch_close
run_orig: <_UnixSelectorEventLoop running=True closed=False debug=False>
```

IOW, `_patch_close` only applies to `loop`, as expected.

> * What happens if multiple of these `ProgressTask`s are stacked up, I guess each one overwrites the `close` method, but remembers the previous one, so we do unwind everything?

This is a very good catch. Indeed this may not work and cause infinite recursion due to local `event_loop_close` in its original form, but it does work when passing the original `loop.close` function to a `partial`, such as:

```python
import asyncio
from functools import partial


async def run_patch():
    loop = asyncio.get_running_loop()
    print(f"run_patch: {loop}")

    loop_close = loop.close

    def _patch_close(loop_close, *args, **kwargs):
        if not loop.is_closed():
            print(f"_patch_close: {loop_close}")
            loop_close(*args, **kwargs)

    loop.close = partial(_patch_close, loop_close)

    loop_close = loop.close

    def _patch_close2(loop_close, *args, **kwargs):
        if not loop.is_closed():
            print(f"_patch_close2: {loop_close}")
            loop_close(*args, **kwargs)

    loop.close = partial(_patch_close2, loop_close)


loop = asyncio.new_event_loop()
loop.run_until_complete(run_patch())
loop.close()
```

If I'm not overlooking anything, the sample above is equivalent to having multiple `ProgressTask`s stacking up, so the unwinding occurs by always rewriting `loop.close` with the most recent `ProgressTask` that calls `loop.close` that was previously set. This change is now reflected in https://github.com/rapidsai/ucxx/pull/116/commits/c5c2ceb0514f0bf652a4ffe48036befcba6bfc66 .

wence- (approved):
Thanks Peter

pentschev:
Thanks all for the reviews here!

PR #120, opened 2023-11-07
Add new conda environments

pentschev:
UCXX is not necessarily CUDA dependent. This change adds two new environments: `cpp.yaml` for C++-only builds and `all.yaml` for C++ and Python builds.


README.md
```diff
@@ -24,10 +24,12 @@ $ conda install -c conda-forge mamba
 After that, one can proceed as before, but simply replacing `conda` with `mamba` in the environment creation command:
 
 ```
-$ mamba env create -n ucxx -f conda/environments/ucxx-cuda118_arch-x86_64.yml
-$ conda activate ucxx
+$ mamba env create -n ucxx -f conda/environments/all_cuda-118_arch-x86_64.yaml
+$ conda activate all_cuda-118_arch-x86_64
```
bdice, line 28: This should match the name from `-n ucxx` above.
```suggestion
$ conda activate ucxx
```

conda/environments/all.yaml
```diff
@@ -0,0 +1,34 @@
+# This file is generated by `rapids-dependency-file-generator`.
```
bdice, line 1: All environment files need to be generated by `dependencies.yaml`.

The failure in pre-commit is because `rapids-dependency-file-generator` [looks for files that it generated and removes them](https://github.com/rapidsai/dependency-file-generator/blob/78ca9ecbf01edaff93f898f2202fb34342f3820e/src/rapids_dependency_file_generator/rapids_dependency_file_generator.py#L39-L40) before re-generating the new files. You have the header here saying it's a generated file, but there's no declaration of this environment file in `dependencies.yaml`.

This is how you tell it to generate a given environment file: https://github.com/rapidsai/ucxx/blob/4464ca942d5ac71aabe4492e2a6d59da453f06dc/dependencies.yaml#L3-L17

You might want to add a section like this for the C++ dependency list.
```
files:
  cpp:     # This is the filename
    output: conda
    matrix:
      arch: [x86_64]
    includes:
      - build_cpp
      - checks
      - dev
      - test_cpp
```

pentschev: Thanks Bradley, this is helpful. I'd still like to trim that more though, for example the above will result in:

```
# This file is generated by `rapids-dependency-file-generator`.
# To make changes, edit ../../dependencies.yaml and run `rapids-dependency-file-generator`.
channels:
- rapidsai
- rapidsai-nightly
- dask/label/dev
- conda-forge
- nvidia
dependencies:
- autoconf
- automake
- c-compiler
- cmake>=3.26.4
- cxx-compiler
- dask-cuda==23.12.*
- dask-cudf==23.12.*
- fmt>=9.1.0,<10
- gmock>=1.13.0
- gtest>=1.13.0
- librmm==23.12.*
- libtool
- ninja
- pip
- pkg-config
- pre-commit
- spdlog>=1.11.0,<1.12
name: cpp_arch-x86_64
```

That include a bunch of things we don't need: `rapidsai`/`rapidsai-nightly`/`dask/label/dev`/`nvidia` channels and `dask-cuda`/`dask-cudf`/`librmm`/`pip`/`pkg-config`. Is there a way to minimize that?

Additionally, this is supposed to be valid for `aarch64` too, can we make it architecture agnostic?

bdice: There's currently no way to drop extra channels from the generated files. You can file an issue to rapids-dependency-file-generator if you have a proposal for how the `dependencies.yaml` schema should accommodate that idea, but I'm not sure if it's worth changing. https://github.com/rapidsai/dependency-file-generator

If you depend only on architecture-agnostic packages, you can just drop the arch from the `files:` spec.
```
files:
  cpp:     # This is used as part of the environment's filename
    output: conda
    includes:
      - build_cpp
      - checks
      - dev
      - test_cpp
```

pentschev:
Hi @jameslamb , thanks for the ping here. This is something that we'll need to get done at some point, so I would prefer to keep it open until we have time/solution for that. Do you by chance happen to know whether there's been any progress to the requirements to tackle this issue, based on the previous conversation with Bradley in this PR?

bdice:
If this is important, I would recommend that we update this PR to the latest branch, use `rapids-dependency-file-generator`, and merge it. We don't have a solution for dropping channels from the `rapids-dependency-file-generator` output but we have already removed our dependencies on several of those channels as part of dropping CUDA 11 support.

jameslamb:
Ok, thanks for confirming it's intentionally still open.

> _Do you by chance happen to know whether there's been any progress to the requirements to tackle this issue, based on the previous conversation with Bradley in this PR?_

I agree with @bdice , hopefully the lack of support for excluding conda channels (or including them selectively) in `rapids-dependency-file-generator` is less of a concern here in 2025 given that:

* the `nvidia` and `dask/label/dev` channels are not included by default
* `ucxx` has a build-time dependency on `rmm`, so probably wants the `rapidsai` / `rapidsai-nightly` channels anyway if using `conda` to set up a build environment

If you really really really do not want to have `rapidsai` / `rapidsai-nightly` in these conda environments either, it could be possible through some combination of removing those from `channels:` in `dependencies.yaml` and then adding them via CLI arguments `--prepend-channel` wherever they're needed.

A separate `pre-commit` hook configuration could then manage these other files that don't need these channels, similar to this:

```yaml
      - repo: https://github.com/rapidsai/dependency-file-generator
        rev: v1.19.0
        hooks:
            - id: rapids-dependency-file-generator
              args: ["--clean", "--prepend-channel=rapidsai", "--prepend-channel=rapidsai-nightly"]
            - id: rapids-dependency-file-generator-no-rapids-channels
              args: ["--output=conda", "--file-key=cpp-no-rapids"]
```

PR #123, opened 2023-11-10
Timeout generic callbacks used in destructors and inflight request cancelation

pentschev:
Destructors may be called from Python's garbage collector, disallowing us to disable the GIL for that operation. When that occurs, there might be a deadlock as other threads have the GIL, such as the progress thread when a Python's listener callback is invoked and executed in that thread.

### DIFF
--- a/cpp/src/utils/callback_notifier.cpp
+++ b/cpp/src/utils/callback_notifier.cpp
@@ -18,7 +18,7 @@ namespace utils {
 
 #ifdef __GLIBC__
 static const bool _useSpinlock = []() {
-  auto const libcVersion = std::string_view{gnu_get_libc_version()};
+  auto const libcVersion = std::string{gnu_get_libc_version()};
   auto const dot         = libcVersion.find(".");
   if (dot == std::string::npos) {
     return false;
@@ -27,7 +27,7 @@ static const bool _useSpinlock = []() {
     auto const glibcMajor = std::stoi(libcVersion.substr(0, dot).data());
     auto const glibcMinor = std::stoi(libcVersion.substr(dot + 1).data());
     auto const use        = glibcMajor < 2 || (glibcMajor == 2 && glibcMinor < 25);
-    ucxx_debug("glibc version %s detected, spinlock use is %d", libcVersion.data(), use);
+    ucxx_debug("glibc version %s detected, spinlock use is %d", libcVersion.c_str(), use);
     return use;
   }
 }();
--- a/cpp/src/utils/callback_notifier.cpp
+++ b/cpp/src/utils/callback_notifier.cpp
@@ -49,15 +49,24 @@ void CallbackNotifier::set()
     _conditionVariable.notify_all();
   }
 }
-void CallbackNotifier::wait()
+
+bool CallbackNotifier::wait(uint64_t period)
 {
   if (_useSpinlock) {
     while (!_flag.load(std::memory_order_acquire)) {}
   } else {
     std::unique_lock lock(_mutex);
     // Likewise here, the mutex provides ordering.
-    _conditionVariable.wait(lock, [this]() { return _flag.load(std::memory_order_relaxed); });
+    if (period > 0) {
+      return _conditionVariable.wait_for(
+        lock, std::chrono::duration<uint64_t, std::nano>(period), [this]() {
+          return _flag.load(std::memory_order_relaxed);
+        });
+    } else {
+      _conditionVariable.wait(lock, [this]() { return _flag.load(std::memory_order_relaxed); });
+    }
   }
+  return true;
 }
 
 }  // namespace utils
--- a/cpp/src/worker.cpp
+++ b/cpp/src/worker.cpp
@@ -150,7 +150,7 @@ std::shared_ptr<Worker> createWorker(std::shared_ptr<Context> context,
 
 Worker::~Worker()
 {
-  size_t canceled = cancelInflightRequests();
+  size_t canceled = cancelInflightRequests(3000000000 /* 3s */, 3);
   ucxx_debug("Worker %p canceled %lu requests", _handle, canceled);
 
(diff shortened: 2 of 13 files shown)


cpp/src/worker.cpp
```diff
@@ -413,16 +413,26 @@ size_t Worker::cancelInflightRequests()
...
     progressPending();
   } else if (isProgressThreadRunning()) {
-    utils::CallbackNotifier callbackNotifierPre{};
-    registerGenericPre([&callbackNotifierPre, &canceled, &inflightRequestsToCancel]() {
-      canceled = inflightRequestsToCancel->cancelAll();
-      callbackNotifierPre.set();
-    });
-    callbackNotifierPre.wait();
+    bool cancelSuccess = false;
+    for (uint64_t i = 0; i < maxAttempts; ++i) {
```
wence-, line 417:
```suggestion
    for (uint64_t i = 0; i < maxAttempts && !cancelSuccess; ++i) {
```

python/distributed-ucxx/distributed_ucxx/tests/test_ucxx.py
```diff
@@ -180,7 +180,7 @@ async def test_ucxx_deserialize(ucxx_loop):
     [
         lambda cudf: cudf.Series([1, 2, 3]),
         lambda cudf: cudf.Series([], dtype=object),
-        lambda cudf: cudf.DataFrame([]),
+        lambda cudf: cudf.DataFrame([], dtype=object),
```
wence-, line 183: This one has now arrived with the merge of #125 I think?

wence- (changes requested):
Some minor cleanups (to avoid excessive looping when `maxAttempts > 1` and a suggestion of how to implement a timeout for the spinlock case.

I was trying to think about how we could drop the gil (say) when destructing in python, but I am not sure if it is possible.

wence-:
I wonder if it is also a good idea to explicitly release the gil when dropping the C++ shared_ptr attributes of UCXX python cdef classes. I think something like:
```
diff --git a/python/ucxx/_lib/libucxx.pyx b/python/ucxx/_lib/libucxx.pyx
index 3e26bfb..ef8b4be 100644
--- a/python/ucxx/_lib/libucxx.pyx
+++ b/python/ucxx/_lib/libucxx.pyx
@@ -862,6 +862,10 @@ cdef class UCXEndpoint():
         bint _enable_python_future
         dict _close_cb_data
 
+    def __dealloc__(self):
+        with nogil:
+            self._endpoint = <shared_ptr[Endpoint]>nullptr;
+
     def __init__(
             self,
             uintptr_t shared_ptr_endpoint,
```

(For example) Makes sure that (if this is the last reference to the `Endpoint`) `~Endpoint` runs with the GIL released.

WDYT?

I think this change is worthwhile anyway so I can propose in a separate PR.

cpp/src/utils/callback_notifier.cpp
```diff
@@ -49,15 +49,24 @@ void CallbackNotifier::set()
     _conditionVariable.notify_all();
   }
 }
-void CallbackNotifier::wait()
+
+bool CallbackNotifier::wait(uint64_t period)
 {
   if (_useSpinlock) {
     while (!_flag.load(std::memory_order_acquire)) {}
```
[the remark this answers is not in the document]
pentschev: I'm not sure if this is happening in the cases where the spinlock is used, I think I haven't seen it at least. Therefore, I left it out on purpose and wanted to see what would happen. Do you have strong feelings whether we should add this right away in any case?

python/distributed-ucxx/distributed_ucxx/tests/test_ucxx.py
pentschev: Yes, I added this in all latest PRs to allow them to complete, it's still showing here because the branch wasn't rebased with latest, I'll do it now and it will go away.

cpp/src/utils/callback_notifier.cpp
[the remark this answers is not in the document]
wence-: No, let's leave it as is for now.

pentschev:
> I wonder if it is also a good idea to explicitly release the gil when dropping the C++ shared_ptr attributes of UCXX python cdef classes. I think something like:
> 
> ```
> diff --git a/python/ucxx/_lib/libucxx.pyx b/python/ucxx/_lib/libucxx.pyx
> index 3e26bfb..ef8b4be 100644
> --- a/python/ucxx/_lib/libucxx.pyx
> +++ b/python/ucxx/_lib/libucxx.pyx
> @@ -862,6 +862,10 @@ cdef class UCXEndpoint():
>          bint _enable_python_future
>          dict _close_cb_data
>  
> +    def __dealloc__(self):
> +        with nogil:
> +            self._endpoint = <shared_ptr[Endpoint]>nullptr;
> +
>      def __init__(
>              self,
>              uintptr_t shared_ptr_endpoint,
> ```
> 
> (For example) Makes sure that (if this is the last reference to the `Endpoint`) `~Endpoint` runs with the GIL released.
> 
> WDYT?
> 
> I think this change is worthwhile anyway so I can propose in a separate PR.

Sorry, I missed this reply earlier. I think this is a great idea, it might actually do what we were attempting of releasing the GIL at destruction time, and `__dealloc__`  didn't occur to me before so thanks for thinking of that! I agree a separate PR is probably good, if you want to do it please feel free to submit it, otherwise I can work on it next week as well. :slightly_smiling_face:

wence-:
> I agree a separate PR is probably good, if you want to do it please feel free to submit it, otherwise I can work on it next week as well.

https://github.com/rapidsai/ucxx/pull/132

wence- (approved):
Thanks Peter

pentschev:
Thanks for the review @wence- !

PR #124, opened 2023-11-13
Fix `skbuild` arguments in `build.sh`

pentschev:
The previous method was incorrect and did not work as expected, for example specifying `-g` would not result in a Cython debug build.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -219,9 +219,11 @@ fi
 if buildAll || hasArg ucxx; then
 
     cd ${REPODIR}/python/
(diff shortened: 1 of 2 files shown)


wence- (approved):
Doesn't look obviously wrong, but I don't really know how these builds work.

pentschev:
To be entirely honest, I'm not always sure about those build tools either, this time I [borrowed from cuDF](https://github.com/rapidsai/cudf/blob/afd7d189b83cbcccba783877f42bb153b5cf315e/build.sh#L335-L337). In any case, I've tested this to be working currently, so I'm gonna go ahead and merged it as is. Thanks @wence- !

PR #125, opened 2023-11-13
Add argument to enable/disable endpoint error handling in benchmarks

pentschev:
Some transports (e.g., `sm`) do not support endpoint error handling. To be able to benchmark them, add a new argument to benchmarks to enable or disable endpoint error handling.

Taken from https://github.com/rapidsai/ucx-py/pull/1007 .


wence- (approved):
Makes sense

PR #132, opened 2023-11-17
Release GIL when dropping C++ objects

wence-:
The Cython cdef classes wrapping UCXX C++ objects should drop the GIL when releasing their stored C++ objects, in case a callback that ends up running on another thread in the C++ dtor needs to acquire the GIL.

### DIFF
--- a/python/ucxx/_lib/libucxx.pyx
+++ b/python/ucxx/_lib/libucxx.pyx
@@ -232,7 +232,7 @@ cdef class UCXConfig():
 
     def __dealloc__(self):
         with nogil:
-            self._config.reset(nullptr)
+            self._config.reset()
 
     def get(self):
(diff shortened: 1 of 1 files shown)


python/ucxx/_lib/libucxx.pyx
```diff
@@ -230,6 +230,10 @@ cdef class UCXConfig():
         # in `__init__`.
         self._config = move(make_unique[Config](user_options))
 
+    def __dealloc__(self):
+        with nogil:
+            self._config.reset(nullptr)
```
pentschev, line 235:
```suggestion
            self._config.reset()
```

pentschev (approved):
Thanks @wence- for working on it. Left one minor request just to make everything consistent, but otherwise looks good!

PR #136, opened 2023-11-27
Set timeout for closing endpoints in Python

pentschev:
https://github.com/rapidsai/ucxx/pull/123 introduced timeouts to the generic callbacks, preventing failure to acquire lock due to GIL competition. However, those were not exposed to Python and at least one of the reasons it still timeouts is because of that, notice how the default `period=0` (never unblock) is used:

```cpp
Thread 1 (Thread 0x7f36d675f740 (LWP 155586) "pytest"):
#0  futex_wait_cancelable (private=<optimized out>, expected=0, futex_word=0x7fff45058e58) at ../sysdeps/nptl/futex-internal.h:183
#1  __pthread_cond_wait_common (abstime=0x0, clockid=0, mutex=0x7fff45058e08, cond=0x7fff45058e30) at pthread_cond_wait.c:508
#2  __pthread_cond_wait (cond=0x7fff45058e30, mutex=0x7fff45058e08) at pthread_cond_wait.c:647
#3  0x00007f36d43634d4 in std::condition_variable::wait<ucxx::utils::CallbackNotifier::wait(uint64_t)::<lambda()> > (__p=..., __lock=..., this=0x7fff45058e30) at /opt/conda/envs/test/x86_64-conda-linux-gnu/include/c++/11.4.0/condition_variable:103
#4  ucxx::utils::CallbackNotifier::wait (this=this@entry=0x7fff45058e00, period=period@entry=0) at /datasets/pentschev/src/ucxx-deadlock/cpp/src/utils/callback_notifier.cpp:66
#5  0x00007f36d43470e1 in ucxx::Endpoint::close (this=0x7f369c701a90, period=0, maxAttempts=1) at /datasets/pentschev/src/ucxx-deadlock/cpp/src/endpoint.cpp:171
#6  0x00007f36d4753381 in __pyx_pw_4ucxx_4_lib_7libucxx_11UCXEndpoint_9close(_object*, _object* const*, long, _object*) () from /opt/conda/envs/test/lib/python3.10/site-packages/ucxx/_lib/libucxx.cpython-310-x86_64-linux-gnu.so
```

This PR exposes those arguments to Python and specify a default for Python async API `Endpoint.abort()` to prevent such deadlocks from occurring.

### DIFF
--- a/python/ucxx/_lib_async/endpoint.py
+++ b/python/ucxx/_lib_async/endpoint.py
@@ -68,8 +68,9 @@ class Endpoint:
         Parameters
         ----------
         period: int
-            maximum period to wait for a generic pre/post progress thread
-            operation will wait for.
+            maximum period to wait (in ns) for internal endpoint operations
+            to complete, usually two operations (pre and post) are involved
+            thus the maximum perceived timeout should be multiplied by two.
         max_attempts: int
             maximum number of attempts to close endpoint, only applicable
             if worker is running a progress thread and `period > 0`.
@@ -81,7 +82,7 @@ class Endpoint:
         self._ep = None
         self._ctx = None
 
-    async def close(self, period=0, max_attempts=1):
+    async def close(self, period=int(1e9), max_attempts=1):
         """Close the endpoint cleanly.
         This will attempt to flush outgoing buffers before actually
         closing the underlying UCX endpoint.
@@ -93,8 +94,9 @@ class Endpoint:
         Parameters
         ----------
         period: int
-            maximum period to wait for a generic pre/post progress thread
-            operation will wait for.
+            maximum period to wait (in ns) for internal endpoint operations
+            to complete, usually two operations (pre and post) are involved
+            thus the maximum perceived timeout should be multiplied by two.
         max_attempts: int
             maximum number of attempts to close endpoint, only applicable
             if worker is running a progress thread and `period > 0`.
--- a/python/ucxx/_lib_async/endpoint.py
+++ b/python/ucxx/_lib_async/endpoint.py
@@ -77,12 +77,12 @@ class Endpoint:
         """
         if self._ep is not None:
             logger.debug("Endpoint.abort(): 0x%x" % self.uid)
-            # Wait for a maximum of 10s
+            # Wait for a maximum of `period` ns
             self._ep.close(period=period, max_attempts=max_attempts)
         self._ep = None
         self._ctx = None
 
-    async def close(self, period=10**9, max_attempts=1):
+    async def close(self, period=10**10, max_attempts=1):
         """Close the endpoint cleanly.
         This will attempt to flush outgoing buffers before actually
         closing the underlying UCX endpoint.
--- a/python/ucxx/_lib_async/endpoint.py
+++ b/python/ucxx/_lib_async/endpoint.py
@@ -51,26 +51,56 @@ class Endpoint:
         """Is this endpoint closed?"""
         return self._ep is None or not self._ep.is_alive()
 
-    def abort(self):
+    def abort(self, period=int(10e9), max_attempts=1):
         """Close the communication immediately and abruptly.
         Useful in destructors or generators' ``finally`` blocks.
 
+        Despite the attempt to close communication immediately, in some
+        circumstances, notably when the parent worker is running a progress
+        thread, a maximum timeout may be specified for which the close operation
+        will wait. This can be particularly important for cases where the progress
+        thread might be attempting to acquire the GIL while the current
+        thread owns that resource.
+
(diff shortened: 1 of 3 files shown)


wence-:
These hangs occur even with #132 as well?

python/ucxx/_lib_async/endpoint.py
```diff
@@ -51,26 +51,58 @@ def closed(self):
...
         if self._ep is not None:
             logger.debug("Endpoint.abort(): 0x%x" % self.uid)
-            self._ep.close()
+            # Wait for a maximum of 10s
+            self._ep.close(period=period, max_attempts=max_attempts)
         self._ep = None
         self._ctx = None
 
-    async def close(self):
+    async def close(self, period=int(1e9), max_attempts=1):
```
wence-, line 85:
```suggestion
    async def close(self, period=10**9, max_attempts=1):
```

WDYT?

python/ucxx/_lib_async/endpoint.py
```diff
@@ -51,26 +51,58 @@ def closed(self):
         """Is this endpoint closed?"""
         return self._ep is None or not self._ep.is_alive()
 
-    def abort(self):
+    def abort(self, period=int(10e9), max_attempts=1):
```
wence-, line 54:
```suggestion
    def abort(self, period=10**10, max_attempts=1):
```

wence- (approved):

pentschev:
> These hangs occur even with #132 as well?

Yes, but note that here the situation #132 is resolving for has not occurred yet. The issue that this PR is attempting to solve occurs when we call `UCXEndpoint.close()` which calls `ucxx::Endpoint::close()` _without_ destroying `ucxx::Endpoint`.

python/ucxx/_lib_async/endpoint.py
```diff
@@ -51,26 +51,58 @@ def closed(self):
...
         if self._ep is not None:
             logger.debug("Endpoint.abort(): 0x%x" % self.uid)
-            self._ep.close()
+            # Wait for a maximum of 10s
+            self._ep.close(period=period, max_attempts=max_attempts)
         self._ep = None
         self._ctx = None
 
-    async def close(self):
+    async def close(self, period=10**9, max_attempts=1):
```
pentschev, line 85:
```suggestion
    async def close(self, period=10**10, max_attempts=1):
```

python/ucxx/_lib_async/endpoint.py
```diff
@@ -51,26 +51,58 @@ def closed(self):
...
+            to complete, usually two operations (pre and post) are involved
+            thus the maximum perceived timeout should be multiplied by two.
+        max_attempts: int
+            maximum number of attempts to close endpoint, only applicable
+            if worker is running a progress thread and `period > 0`.
         """
         if self._ep is not None:
             logger.debug("Endpoint.abort(): 0x%x" % self.uid)
-            self._ep.close()
+            # Wait for a maximum of 10s
```
pentschev, line 80:
```suggestion
            # Wait for a maximum of `period` ns
```

python/ucxx/_lib_async/endpoint.py
pentschev: I think you meant to write `10**10` here too @wence- , but please correct me if I missed something.

wence-: I just copied the value that was here https://github.com/rapidsai/ucxx/pull/136/commits/1cef6d9b15162bea243a66e3a36f417d5f0a38a4, but 10s in both cases as a default seems fine.

pentschev: The value was `int(1e9)`, which is `10**10`. The other instance you actually wrote `10**10` but here you wrote `10**9` which I then assumed was accidental.

Also a side note: I started thinking about this 10s/1 attempt default and I'm not sure it makes sense. Perhaps a more sane default would be 3s/3 attempts. Perhaps it doesn't make sense to block for such a long period and giving up immediately, perhaps a shorter wait and multiple attempts will give it more room to unblock if something went wrong and retry. I'll leave it like this for now but will reconsider this in the future.

wence-: `1e9` is `10**9` (at least last time I looked :P) but the other one was `10e9` (which is `10**10`). But this is an aside. I too think it might make sense to have a retry setup, but not sure.

pentschev: > `1e9` is `10**9` (at least last time I looked :P) but the other one was `10e9` (which is `10**10`).

Ha! True, it was my typo, twice! I meant to write `10e9` here _and_ in the code where you just corrected me about. 🤦

pentschev:
Thanks @wence- for another review. 😄

Issue #138, opened 2023-12-01
Failure in `cuCtxGetCurrent` in CUDA 11.4.3 and RAPIDS 24.02


pentschev:
# Failure in Python `test_terminate_process_kill_timeout`

The failure described here has been occurring in CI and is deterministically happening and thus blocking builds in `branch-0.36` (RAPIDS 24.02).

## Clone or clean UCXX

First, make sure to have a _clean_ clone of UCXX in `branch-23.12`:

```
git clone https://github.com/rapidsai/ucxx ucxx-ctx-failure -b branch-0.36
cd ucxx-ctx-failure
```

The steps above can be skipped but _no cached files may exist_, so make sure to run `git clean -ffdx` if you already have a clone of `ucxx`.

## Setup

To reproduce (tested on a DGX-1 baremetal with driver `525.125.06`), start the container:

```bash
docker run --rm -it --pull always --gpus device=0 --network=host --volume $PWD:/repo --workdir /repo rapidsai/ci-conda:cuda11.4.3-centos7-py3.9
```

Inside the container:

```bash
./ci/build_cpp.sh
./ci/test_python.sh
```

## Failure

This will eventually cause the failure below:

<details>

```cpp
python/ucxx/_lib/tests/test_utils.py::test_terminate_process_kill_timeout[default] [1701464172.314996] [dgx13:58152:0] cuda_copy_iface.c:461  UCX  ERROR cuCtxGetCurrent(&cuda_context)() failed: initialization error
[1701464172.315044] [dgx13:58152:0]  cuda_ipc_iface.c:531  UCX  ERROR cuCtxGetCurrent(&cuda_context)() failed: initialization error
PASSED
python/ucxx/_lib/tests/test_utils.py::test_terminate_process_kill_timeout[fork] [1701464173.303491] [dgx13:54899:0]       event_set.c:176  UCX  ERROR epoll_ctl(event_fd=48, DEL, fd=72) failed: No such file or directory
[1701464173.303547] [dgx13:54899:0]           async.c:556  UCX  WARN  failed to remove async handler 0x55a01403e320 [id=72 ref 1] ???() : Input/output error
[1701464173.303586] [dgx13:54899:0]       event_set.c:176  UCX  ERROR epoll_ctl(event_fd=48, DEL, fd=73) failed: No such file or directory
[...]

PR #139, opened 2023-12-04
Ensure garbage is collected in Python tests

pentschev:
C++ destructors may not run on time if Python garbage collection doesn't run soon enough, this is problematic if we fork the process causing state from previous tests to leak into the new processes which may fail due to lack of state owned by the parent process (e.g., CUDA context). Additionally, forking after the process has started the worker progress thread and hasn't stopped is problematic as it will be a shared state in the new process.

Fixes https://github.com/rapidsai/ucxx/issues/138 .


python/ucxx/_lib/tests/conftest.py
```diff
@@ -0,0 +1,20 @@
...
+# For example, without `RAPIDS_NO_INITIALIZE=True`, `test_benchmark_cluster`
+# will succeed if running alone, but fails when all tests are run in batch.
+os.environ["RAPIDS_NO_INITIALIZE"] = "True"
+
+
+@pytest.fixture(autouse=True)
+def collect_garbage():
+    import gc
+
+    gc.collect()
```
wence-, line 20: I think the more idiomatic way of doing this is to define a teardown protocol (https://docs.pytest.org/en/7.1.x/_modules/_pytest/hookspec.html#pytest_runtest_teardown)
```
def pytest_runtest_teardown(item, nextitem):
    yield
    gc.collect()
```

(We can move the `import gc` to the top-level).

python/ucxx/_lib_async/tests/conftest.py
```diff
@@ -16,6 +16,13 @@
 os.environ["RAPIDS_NO_INITIALIZE"] = "True"
 
 
+@pytest.fixture(autouse=True)
+def collect_garbage():
+    import gc
+
+    gc.collect()
+
```
wence-, line 24: Similarly here, perhaps:

```
def pytest_runtest_teardown(item, nextitem):
    yield
    gc.collect()
```

wence- (approved):
Minor idiom suggestion, but good spot!

python/ucxx/_lib/tests/conftest.py
pentschev: Great suggestion, I forgot about pytest hooks. Done in https://github.com/rapidsai/ucxx/pull/139/commits/bb63249bd25a33632507a1fe1800874ca04c18cb .

wence-: Sorry, and I made a bork. The `yield` should not be there.

python/ucxx/_lib_async/tests/conftest.py
```diff
@@ -16,9 +17,9 @@
 os.environ["RAPIDS_NO_INITIALIZE"] = "True"
 
 
-@pytest.fixture(autouse=True)
-def collect_garbage():
-    import gc
+@pytest.hookimpl(hookwrapper=True)
+def pytest_runtest_teardown(item, nextitem):
+    yield
```
wence-, line 22:
```suggestion
```

python/ucxx/_lib/tests/conftest.py
```diff
@@ -0,0 +1,21 @@
...
+# context on the appropriate device to fail.
+# For example, without `RAPIDS_NO_INITIALIZE=True`, `test_benchmark_cluster`
+# will succeed if running alone, but fails when all tests are run in batch.
+os.environ["RAPIDS_NO_INITIALIZE"] = "True"
+
+
+@pytest.hookimpl(hookwrapper=True)
+def pytest_runtest_teardown(item, nextitem):
+    yield
+
```
pentschev, line 20:
```suggestion
```

python/ucxx/_lib_async/tests/conftest.py
```diff
@@ -16,6 +17,13 @@
 os.environ["RAPIDS_NO_INITIALIZE"] = "True"
 
 
+@pytest.hookimpl(hookwrapper=True)
+def pytest_runtest_teardown(item, nextitem):
+    yield
+
```
pentschev, line 23:
```suggestion
```

python/ucxx/_lib/tests/conftest.py
```diff
@@ -14,6 +14,6 @@
 os.environ["RAPIDS_NO_INITIALIZE"] = "True"
 
 
-@pytest.hookimpl(hookwrapper=True)
+@pytest.hookimpl()
```
wence-, line 17: I'm not sure you need this decorator at all?

pentschev: `@pytest.hookimpl()` is required otherwise it won't apply to tests, not automatically at least.

pentschev:
I think there's no reason not to get this merged even with #142 being in, though I think it would be useful to wait a couple of days before merging to see how nightly builds look like after #142.

python/ucxx/_lib/tests/conftest.py
wence-: Hmm, I was confused then. If I have:

```
# conftest.py
def pytest_runtest_teardown(item, nextitem):
    print("Running teardown")
```

```
# test.py

def test():
    print("Running test")
```

Then
```
pytest test.py -s -v
[...]
test.py::test Running test
PASSEDRunning teardown
```

The addition of the decorator doesn't make a difference.

I _think_ this is because the name `pytest_runtest_teardown` is already registered as a hook name, so when pytest collects the conftest.py it notices the hook and adds it to the known hooks to be called.

wence-:
> I think there's no reason not to get this merged even with #142 being in, though I think it would be useful to wait a couple of days before merging to see how nightly builds look like after #142.

Agreed

python/ucxx/_lib/tests/conftest.py
pentschev: Ok, I must have done something wrong then. I initially tested without the decorator and for some reason it didn't print, I tried this again now and it prints. Removed it now in https://github.com/rapidsai/ucxx/pull/139/commits/c79bd35a37aaaf68eb5a8de42fa63e3e26dd7c28 .

pentschev:
We only got branch-0.36 nightlies actually building about 4-5 days ago, but the original issue from #138 hasn't happened since so #142 must have fixed the issue. In any case as discussed previously there's no harm in merging this PR, so I'm going ahead with that now.

PR #140, opened 2023-12-04
Use `notify_one()` in `CallbackNotifier`

pentschev:
It is unclear why but for some reason `notify_all()` is causing futexes never to return in some situations. This occurs very frequently in CI and is also less frequently reproducible locally.

The typical stack trace for the blocked thread is shown below:

```cpp
Thread 6 (Thread 0x7f13ec84f700 (LWP 2823667) "pytest"):
#0  futex_wait (private=<optimized out>, expected=32765, futex_word=0x7ffd5186a874) at ../sysdeps/nptl/futex-internal.h:141
#1  futex_wait_simple (private=<optimized out>, expected=32765, futex_word=0x7ffd5186a874) at ../sysdeps/nptl/futex-internal.h:172
#2  __condvar_quiesce_and_switch_g1 (private=<optimized out>, g1index=<synthetic pointer>, wseq=<optimized out>, cond=0x7ffd5186a860) at pthread_cond_common.c:416
#3  __pthread_cond_broadcast (cond=0x7ffd5186a860) at pthread_cond_broadcast.c:73
#4  0x00007f140fe5f23c in ucxx::BaseDelayedSubmissionCollection<std::function<void ()> >::process() (this=0x560d0effafd0) at /repo/cpp/include/ucxx/delayed_submission.h:154
#5  0x00007f140fe5f399 in ucxx::DelayedSubmissionCollection::processPost (this=<optimized out>) at /repo/cpp/src/delayed_submission.cpp:84
#6  0x00007f140fe7ed71 in ucxx::WorkerProgressThread::progressUntilSync(std::function<bool ()>, bool const&, std::function<void (void*)>, void*, std::shared_ptr<ucxx::DelayedSubmissionCollection>) (progressFunction=..., stop=@0x560d0f6527f8: false, startCallback=..., startCallbackArg=<optimized out>, delayedSubmissionCollection=...) at /opt/conda/envs/test/x86_64-conda-linux-gnu/include/c++/11.4.0/bits/shared_ptr_base.h:1295
#7  0x00007f140fe7f3ee in std::__invoke_impl<void, void (*)(std::function<bool ()>, bool const&, std::function<void (void*)>, void*, std::shared_ptr<ucxx::DelayedSubmissionCollection>), std::function<bool ()>, std::reference_wrapper<bool>, std::function<void (void*)>, void*, std::shared_ptr<ucxx::DelayedSubmissionCollection> >(std::__invoke_other, void (*&&)(std::function<bool ()>, bool const&, std::function<void (void*)>, void*, std::shared_ptr<ucxx::DelayedSubmissionCollection>), std::function<bool ()>&&, std::reference_wrapper<bool>&&, std::function<void (void*)>&&, void*&&, std::shared_ptr<ucxx::DelayedSubmissionCollection>&&) (__f=<optimized out>, __f=<optimized out>) at /opt/conda/envs/test/x86_64-conda-linux-gnu/include/c++/11.4.0/bits/invoke.h:61
#8  std::__invoke<void (*)(std::function<bool ()>, bool const&, std::function<void (void*)>, void*, std::shared_ptr<ucxx::DelayedSubmissionCollection>), std::function<bool ()>, std::reference_wrapper<bool>, std::function<void (void*)>, void*, std::shared_ptr<ucxx::DelayedSubmissionCollection> >(void (*&&)(std::function<bool ()>, bool const&, std::function<void (void*)>, void*, std::shared_ptr<ucxx::DelayedSubmissionCollection>), std::function<bool ()>&&, std::reference_wrapper<bool>&&, std::function<void (void*)>&&, void*&&, std::shared_ptr<ucxx::DelayedSubmissionCollection>&&) (__fn=<optimized out>) at /opt/conda/envs/test/x86_64-conda-linux-gnu/include/c++/11.4.0/bits/invoke.h:96
#9  std::thread::_Invoker<std::tuple<void (*)(std::function<bool ()>, bool const&, std::function<void (void*)>, void*, std::shared_ptr<ucxx::DelayedSubmissionCollection>), std::function<bool ()>, std::reference_wrapper<bool>, std::function<void (void*)>, void*, std::shared_ptr<ucxx::DelayedSubmissionCollection> > >::_M_invoke<0ul, 1ul, 2ul, 3ul, 4ul, 5ul>(std::_Index_tuple<0ul, 1ul, 2ul, 3ul, 4ul, 5ul>) (this=<optimized out>) at /opt/conda/envs/test/x86_64-conda-linux-gnu/include/c++/11.4.0/bits/std_thread.h:259
#10 std::thread::_Invoker<std::tuple<void (*)(std::function<bool ()>, bool const&, std::function<void (void*)>, void*, std::shared_ptr<ucxx::DelayedSubmissionCollection>), std::function<bool ()>, std::reference_wrapper<bool>, std::function<void (void*)>, void*, std::shared_ptr<ucxx::DelayedSubmissionCollection> > >::operator()() (this=<optimized out>) at /opt/conda/envs/test/x86_64-conda-linux-gnu/include/c++/11.4.0/bits/std_thread.h:266
#11 std::thread::_State_impl<std::thread::_Invoker<std::tuple<void (*)(std::function<bool ()>, bool const&, std::function<void (void*)>, void*, std::shared_ptr<ucxx::DelayedSubmissionCollection>), std::function<bool ()>, std::reference_wrapper<bool>, std::function<void (void*)>, void*, std::shared_ptr<ucxx::DelayedSubmissionCollection> > > >::_M_run() (this=<optimized out>) at /opt/conda/envs/test/x86_64-conda-linux-gnu/include/c++/11.4.0/bits/std_thread.h:211
#12 0x00007f140f92fe95 in std::execute_native_thread_routine (__p=<optimized out>) at ../../../../../libstdc++-v3/src/c++11/thread.cc:104
#13 0x00007f1412647609 in start_thread (arg=<optimized out>) at pthread_create.c:477
#14 0x00007f1412412133 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

### DIFF
--- a/cpp/include/ucxx/utils/callback_notifier.h
+++ b/cpp/include/ucxx/utils/callback_notifier.h
@@ -20,15 +20,15 @@ class CallbackNotifier {
    * @brief Construct a thread-safe notification object
    *
    * Construct a thread-safe notification object which can signal
-   * release of some shared state with `set()` while other threads
-   * block on `wait()` until the shared state is released.
+   * release of some shared state with `set()` while a single thread
+   * blocks on `wait()` until the shared state is released.
    *
    * If libc is glibc and the version is older than 2.25, the
    * implementation uses a spinlock otherwise it uses a condition
    * variable.
    *
    * When C++-20 is the minimum supported version, it should use
-   * atomic.wait + notify_all.
+   * atomic.wait + notify_one.
    */
   CallbackNotifier() : _flag{false} {};
 
@@ -42,8 +42,8 @@ class CallbackNotifier {
   /**
    * @brief Notify waiting threads that we are done and they can proceed
    *
-   * Set the flag to true and notify others threads blocked by a call to `wait()`.
-   * See also `std::condition_variable::notify_all`.
+   * Set the flag to true and notify a single thread blocked by a call to `wait()`.
+   * See also `std::condition_variable::notify_one`.
    */
   void set();
 
--- a/cpp/src/utils/callback_notifier.cpp
+++ b/cpp/src/utils/callback_notifier.cpp
@@ -46,7 +46,7 @@ void CallbackNotifier::set()
       // ordering.
       _flag.store(true, std::memory_order_relaxed);
     }
-    _conditionVariable.notify_all();
+    _conditionVariable.notify_one();
   }
 }
 


pentschev:
I need to confirm that this is the only change that really (seems to have) fixed the bug, I have some other local changes I've been testing against for 8 hours and it didn't deadlock once, where before some of the runs would start deadlocking in <1 hour. This has been the latest change that actually resolved the problem, I just want to confirm the other changes have no part in it.

pentschev:
@wence- can you think of any reason `notify_all()` would be causing issues here? I don't think the way we currently use `CallbackNotifier` should make any difference on utilizing `notify_one` vs `notify_all` but I maybe ignoring some detail.

wence-:
I think switching to notify_one is fine. I think originally we had that, and then I suggested moving to notify_all in case we ever used it in single producer, multiple consumer mode. But we're not doing that.

pentschev:
> I think switching to notify_one is fine. I think originally we had that, and then I suggested moving to notify_all in case we ever used it in single producer, multiple consumer mode. But we're not doing that.

Yes, I agree. I'm wondering though if you have any ideas why `notify_all()` could be problematic the way we're using it, AFAIU they should be equivalent for a single consumer mode, so the fact we get stuck at `__pthread_cond_broadcast` seems very odd.

wence- (changes requested):
Can we adjust the docstrings in `callback_notifier.h` to this change. They currently say things about `notify_all` and "waiting threads", whereas now (which is fine), this becomes a mechanism for communicating between two threads.

pentschev:
> Can we adjust the docstrings in `callback_notifier.h` to this change. They currently say things about `notify_all` and "waiting threads", whereas now (which is fine), this becomes a mechanism for communicating between two threads.

Done in https://github.com/rapidsai/ucxx/pull/140/commits/54414d1aa01de13a288b87039f210f825a8511a9, would you mind taking another look?

wence- (approved):

wence-:
> Yes, I agree. I'm wondering though if you have any ideas why `notify_all()` could be problematic the way we're using it, AFAIU they should be equivalent for a single consumer mode, so the fact we get stuck at `__pthread_cond_broadcast` seems very odd.

My vague understanding is that pthread_cond_broadcast is one of the tricksiest parts of the pthread signalling code to get right, so :shrug:

pentschev:
Thanks @wence- for reviewing!

PR #141, opened 2023-12-05
Clear out callback data once we have run the callback

wence-:
Although the UCXX python layer is careful not to set up refcycles (such that everything is collected deterministically), the callbacks are a loophole that can break this guarantee.

It is easy to set up a callback that references an object that produces a reference cycle. To avoid this, once we are done, internally, with the callback function (after it is called) drop any references to it.

This reduces the ease of post-hoc debugging (since one the callback has run, we can't see what it was), but ensures that we don't accidentally have a refcycle that inhibits deterministic collection of UCXX objects.

### DIFF
--- a/python/ucxx/_lib/libucxx.pyx
+++ b/python/ucxx/_lib/libucxx.pyx
@@ -435,6 +435,7 @@ cdef void _generic_callback(void *args) with gil:
             *cb_data['cb_args'],
             **cb_data['cb_kwargs']
         )
+        cb_data.clear()
     except Exception as e:
         pass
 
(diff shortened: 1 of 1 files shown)


wence-:
Related to #139.

wence-:
I think this should be enough without #139, but let us put that in as well (it can't really harm things)

pentschev (approved):
I think this is incorrect, clearing `cb_data` will, AFAIU, clear the object in its entirety. A listener for example will reuse the same callback every time there's an incoming connection. Am I missing something?

wence-:
> I think this is incorrect, clearing `cb_data` will, AFAIU, clear the object in its entirety. A listener for example will reuse the same callback every time there's an incoming connection. Am I missing something?

No, you are right, this doesn't work :(.

pentschev:
Thanks anyway for trying to clean this @wence- , too bad it can't work as is, it would have been very much better.

PR #142, opened 2023-12-05
Avoid refcycle in listener create callback for deliver_endpoint=True

wence-:
Rather than storing a strong reference to the listener object we created, use a weakref.

### DIFF
--- a/python/ucxx/_lib/libucxx.pyx
+++ b/python/ucxx/_lib/libucxx.pyx
@@ -6,6 +6,7 @@ import asyncio
 import enum
 import functools
 import logging
+import weakref
 
 from cpython.buffer cimport PyBUF_FORMAT, PyBUF_ND, PyBUF_WRITABLE
 from cpython.ref cimport PyObject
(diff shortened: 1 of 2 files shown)


python/ucxx/_lib/libucxx.pyx
```diff
@@ -1302,6 +1303,7 @@ cdef class UCXListener():
         shared_ptr[Listener] _listener
         bint _enable_python_future
         dict _cb_data
+        object __weakref__
```
pentschev, line 1306: Not sure I understand the purpose of this, could you explain its purpose?

pentschev (approved):
Those changes look sensible to me. Thanks for resolving this, I certainly overlooked using a `weakref` here which is the more appropriate solution.

python/ucxx/_lib/libucxx.pyx
wence-: By default, `cdef` classes are not weakrefable (because they're just a struct, and so the `__weakref__` attribute doesn't exist). To enable it, we need this slot (https://cython.readthedocs.io/en/latest/src/userguide/extension_types.html#making-extension-types-weak-referenceable). This is kind of morally equivalent to defining a Python class with `__slots__`, at which point you must explicitly put `__weakref__` in slots to make the class weakrefable.

pentschev: Interesting, I wasn't aware of that. Thanks for the details.

pentschev:
Thanks @wence- for digging on this issue and finding a fix! 😄

PR #143, opened 2023-12-06
Use a weakref finalizer to stop notifier thread

wence-:
When creating our own application context with the python notifier thread enabled we previously had to arrange to stop it before dropping the reference to the context we created. The consequence of not doing so is that our application hangs in exit waiting for a thread to join that never will.

Applying the principle of least surprise, an object should clean up all of its resources when dropping out of scope. To ensure this change Application.stop_notifier_thread to a staticmethod and attach it as a weakref finalizer when starting the notifier thread. This way, we don't have to handle the case where there is no notifier thread, nor do we have to remember to call the stop function manually.

This additionally allows us to clean up the top-level reset function since there is no longer a need to stop the notifier "manually" during reset.

### DIFF
--- a/python/ucxx/_lib_async/application_context.py
+++ b/python/ucxx/_lib_async/application_context.py
@@ -1,6 +1,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2022-2023, NVIDIA CORPORATION & AFFILIATES.
 # SPDX-License-Identifier: BSD-3-Clause
 
+import functools
 import logging
 import os
 import threading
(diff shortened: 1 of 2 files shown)


pentschev:
This looks like a great solution, thanks for working on it @wence- . I see some tests are failing though because they explicitly call `ucxx.stop_notifier_thread()` so that should be fixed after removal, unfortunately I can't suggest changes to files that weren't touched via the GH interface.

pentschev:
> How do we want to handle stopping the notifier in distributed-ucxx? It seems like there it might make sense to still stop the notifier explicitly, rather than waiting for the global application context to go out of scope in atexit teardown.

Yeah, that was the only way I found to stop the notifier thread "on time" for distributed, if you have other ideas how to make that better they're certainly welcome. If you're worried about the top-level call exclusively, we should be ok removing that and then calling it directly from the `ApplicationContext` inside `distributed-ucxx`, i.e.: `ucxx.core._get_ctx().stop_notifier_thread()`.

wence-:
I figured out the problem. The test_benchmark_cluster test relies on an explicit call to `ucxx.stop_notifier_thread()`. The fixture that automatically resets ucxx (and hence stops the notifier) doesn't work for this setup because ucxx has been initialised in a subprocess.

wence-:
So I just dropped the removal of `stop_notifier_thread` everywhere.

wence-:
Hmm, just the one test run failed, but no obvious error.

pentschev:
> Hmm, just the one test run failed, but no obvious error.

Seems like this was an exit code `124` which means the `pytest` process timed out. Are you able to reproduce that locally? It seems like there may be some resource preventing the process from returning.

PR #146, opened 2023-12-12
Switch to scikit-build-core

vyasr:
Contributes to rapidsai/build-planning#2

PR #148, opened 2024-01-03
Update `dependencies.yaml` to support CUDA 12.*

pentschev:


conda/environments/all_cuda-120_arch-x86_64.yaml
ajschmidt8, line 1: Is there any reason why this file was removed? I would expect it to stick around due to this line in `dependencies.yaml`:

- https://github.com/rapidsai/ucxx/blob/c84c0f976c5b4b5b6d489366696013e77ed515b4/dependencies.yaml#L6

Typically this should be caught by pre-commit, but there is an issue with pre-commit identifying new files as described in this issue:

- https://github.com/pre-commit/pre-commit/issues/2789

quasiben (approved):

conda/environments/all_cuda-120_arch-x86_64.yaml
ajschmidt8: ok, i'll have to investigate this separately. i will approve for now.

ajschmidt8 (approved):

dependencies.yaml
```diff
@@ -115,7 +115,7 @@ dependencies:
               - cuda-version=11.8
               - cudatoolkit
           - matrix:
-              cuda: "12.0"
+              cuda: "12.*"
             packages:
               - cuda-version=12.0
               - cuda-cudart-dev
```
bdice, line 121: This isn't right. We shouldn't be installing `cuda-version=12.0` for all `cuda 12.*` inputs.

conda/environments/all_cuda-120_arch-x86_64.yaml
bdice: Running `pre-commit run --all-files` generated this file again for me. I don't think it was supposed to be removed.

pentschev: Indeed the file showed up locally now, see https://github.com/rapidsai/ucxx/pull/146#pullrequestreview-1802934101. It should be addressed in there.

PR #150, opened 2024-01-03
Fix CUDA dependency issues from PR #148.

bdice:
Following up on an issue with `dependencies.yaml` from #148. See https://github.com/rapidsai/ucxx/pull/148/files#r1440832477.

cc: @ajschmidt8 @pentschev


pentschev (approved):
LGTM, thanks @bdice .

ajschmidt8 (approved):

dependencies.yaml
```diff
@@ -98,26 +98,31 @@ dependencies:
...
+      - output_types: conda
+        matrices:
+          - matrix:
+              cuda: "11.*"
+            packages:
               - cudatoolkit
           - matrix:
               cuda: "12.*"
             packages:
-              - cuda-version=12.0
```
jakirkham, line 120: Should this be dropped entirely?

dependencies.yaml
```diff
@@ -98,26 +98,31 @@ dependencies:
...
+              - cuda-version=12.0
+      - output_types: conda
+        matrices:
+          - matrix:
+              cuda: "11.*"
+            packages:
               - cudatoolkit
           - matrix:
               cuda: "12.*"
             packages:
```
jakirkham, line 125: Think we would want to just relax the constraint

```suggestion
            packages:
              - cuda-version=12
```

jakirkham (commented):
Thanks Bradley! 🙏

Had some questions below

dependencies.yaml
bdice: The `cuda` matrix entry is set to `major.minor` by our CI images / workflows. We want to make sure that we install the same `major.minor` that is requested.

dependencies.yaml
bdice: No, we need `cuda-version` to be pinned.

jakirkham: Ok am struggling to square this comment with the comment below

If we think we want a pin, why are we not adding a new one (either with my suggestion or a different one)?

dependencies.yaml
```diff
@@ -98,26 +98,31 @@ dependencies:
...
               - cuda-version=11.5
-              - cudatoolkit
           - matrix:
               cuda: "11.8"
             packages:
               - cuda-version=11.8
+          - matrix:
+              cuda: "12.0"
+            packages:
+              - cuda-version=12.0
```
bdice, line 116: @jakirkham I moved this pin upward from below. I can explain the goal here: I split the `cudatoolkit` dependency list into _two_ lists. The first list pins `cuda-version=major.minor` based on the `cuda` matrix entry. The second list adds the relevant CUDA package names for each _major_ version: `cudatoolkit` for 11.* and `cuda-cudart-dev` for 12.*. This uses the new globbing feature from `rapids-dependency-file-generator` 1.8.0, and allows us to de-duplicate the previous code, which listed `cudatoolkit` under every CUDA 11 minor version rather than only once for all CUDA 11.* versions. Does that make more sense?

dependencies.yaml
bdice: See my comment on the moved pinning above: https://github.com/rapidsai/ucxx/pull/150/files#r1440878433

jakirkham: That isn't related to this question

Put another way, what happens to this environment when CUDA 13 exists?

bdice: Values of the `cuda` matrix entry must match to an entry in each `specific:` matrix set. The rules for dependency file generation are documented here: https://github.com/rapidsai/dependency-file-generator#how-dependency-lists-are-merged.

If you try to run this with a `cuda` value that isn't in the matrix set above `[11.2, 11.4, 11.5, 11.8, 12.0]` you get an error. In the near future we will be adding support for 12.2 in accordance with https://github.com/rapidsai/build-planning/issues/6, and we'll need to add an entry for `12.2` to have `cuda-version=12.2` in the first list above when we do that.

```
$ rapids-dependency-file-generator --output conda --file_key all --matrix "cuda=12.1;arch=x86_64;py=3.10"
...
ValueError: No matching matrix found in 'cudatoolkit' for: {'cuda': '12.1', 'arch': 'x86_64', 'py': '3.10'}
```

jakirkham: Ok thanks Bradley! 🙏

jakirkham (approved):

jakirkham:
Looks like there was a test failure [on CI]( https://github.com/rapidsai/ucxx/actions/runs/7402765958/job/20143664050?pr=150#step:7:1089 ):

```python
FAILED python/ucxx/_lib/tests/test_server_client.py::test_server_client[thread-0-stream] - RuntimeError: Process did not exit cleanly (exit code: 1)
```

pentschev:
Had to rerun to get it through, I'm still ironing out some of those tests that occasionally fail.

PR #153, opened 2024-01-08
Fix segmentation fault on canceled requests that are not awaited

pentschev:
Calling `InflightRequests::cancelAll()` does not guarantee the requests to be canceled immediately, just like any other request we must check their status before releasing the `ucxx::Request` object. To prevent the user from releasing a request that has not yet completed and may still call the completion callback we must make sure we still keep a reference
to it until its status is set. With this change we now ensure both `ucxx::Endpoint` and `ucxx::Worker` will only release references after those requests that have issued cancelation complete.

It is also important that all requests have a valid status before destroying the object, thus we should cancel a request if all references to the Cython `UCXRequest` object have been dropped but the request has not completed yet.

Additionally use `std::unique_ptr<InflightRequests>` in `ucxx::Worker` and reenable `test_ucxx_unreachable`.

### DIFF
--- a/cpp/include/ucxx/inflight_requests.h
+++ b/cpp/include/ucxx/inflight_requests.h
@@ -134,7 +134,7 @@ class InflightRequests {
    *
    * @returns The count of requests that are in process of cancelation.
    */
-  size_t getCancelingCount();
+  size_t getCancelingSize();
 };
 
 }  // namespace ucxx
--- a/cpp/include/ucxx/inflight_requests.h
+++ b/cpp/include/ucxx/inflight_requests.h
@@ -15,17 +15,24 @@ class Request;
 
 typedef std::map<const Request* const, std::shared_ptr<Request>> InflightRequestsMap;
 typedef std::unique_ptr<InflightRequestsMap> InflightRequestsMapPtr;
-typedef std::pair<InflightRequestsMapPtr, InflightRequestsMapPtr> InflightRequestsMapPtrPair;
+typedef struct TrackedRequests {
+  InflightRequestsMapPtr _inflight;
+  InflightRequestsMapPtr _canceling;
+
+  TrackedRequests()
+    : _inflight(std::make_unique<InflightRequestsMap>()),
+      _canceling(std::make_unique<InflightRequestsMap>())
+  {
+  }
+} TrackedRequests;
+typedef std::unique_ptr<TrackedRequests> TrackedRequestsPtr;
 
 class InflightRequests {
  private:
-  InflightRequestsMapPtr _inflightRequests{
-    std::make_unique<InflightRequestsMap>()};  ///< Container storing pointers to all inflight
-                                               ///< requests known to the owner of this object
-  InflightRequestsMapPtr _cancelingRequests{
-    std::make_unique<InflightRequestsMap>()};  ///< Container storing pointers to all requests
-                                               ///< known to the owner of this object in
-                                               ///< process of cancelation
+  TrackedRequestsPtr _trackedRequests{
+    std::make_unique<TrackedRequests>()};  ///< Container storing pointers to all inflight
+                                           ///< and in cancelation process requests known to
+                                           ///< the owner of this object
   std::mutex _mutex{};  ///< Mutex to control access to inflight requests container
   std::mutex
     _cancelMutex{};  ///< Mutex to allow cancelation and prevent removing requests simultaneously
@@ -77,10 +84,10 @@ class InflightRequests {
    * Merge containers of inflight requests obtained from `InflightRequests::release()` of
    * another object with the internal containers.
    *
-   * @param[in] inflightRequestsMapPair containers of inflight requests to merge with the
-   *                                    internal containers.
+   * @param[in] trackedRequestsPtr containers of tracked inflight requests to merge with the
+   *                               internal tracked inflight requests.
    */
-  void merge(InflightRequestsMapPtrPair inflightRequestsMapPtrPair);
+  void merge(TrackedRequestsPtr trackedRequests);
 
   /**
    * @brief Remove an inflight request from the internal container.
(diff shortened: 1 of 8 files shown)


cpp/include/ucxx/inflight_requests.h
```diff
@@ -7,23 +7,38 @@
...
 #include <mutex>
+#include <utility>
 
 namespace ucxx {
 
 class Request;
 
 typedef std::map<const Request* const, std::shared_ptr<Request>> InflightRequestsMap;
 typedef std::unique_ptr<InflightRequestsMap> InflightRequestsMapPtr;
+typedef std::pair<InflightRequestsMapPtr, InflightRequestsMapPtr> InflightRequestsMapPtrPair;
```
wence-, line 18: **suggestion**: Although "a name is not type safety", this type seems very easy to get confused with. It's a pair of maps, the first of which is (IIUC) "new" requests that we should record as being in flight, while the second is "existing" requests that we should record for cancellation.

This seems easy to misuse (the types of the two things don't differ) and `.first` and `.second` don't give the reader context to know what is wrong. One must "just know" that the first entry is new and the second to cancel.

How about (no type safety, but name safety):

```
struct ChangedRequests {
   InflightRequestsMapPtr new_;
   InflightRequestsMapPtr to_cancel_;
}
```

cpp/src/inflight_requests.cpp
```diff
@@ -21,11 +21,16 @@ void InflightRequests::insert(std::shared_ptr<Request> request)
...
 
-void InflightRequests::merge(InflightRequestsMapPtr inflightRequestsMap)
+void InflightRequests::merge(InflightRequestsMapPtrPair inflightRequestsMapPtrPair)
 {
-  std::lock_guard<std::mutex> lock(_mutex);
-
-  _inflightRequests->merge(*inflightRequestsMap);
+  {
+    std::lock_guard<std::mutex> lock(_mutex);
+    _inflightRequests->merge(*(inflightRequestsMapPtrPair.first));
```
wence-, line 28: **question:** Now that these are `unique_ptr`s, should this be `std::move(inflightrequestsmapptrpair.first)` rather than `*...`?

cpp/src/inflight_requests.cpp
```diff
@@ -67,6 +72,38 @@ void InflightRequests::remove(const Request* const request)
...
+size_t InflightRequests::dropCanceled()
+{
+  size_t removed = 0;
+
+  {
+    std::scoped_lock lock{_cancelMutex};
+    for (auto it = _cancelingRequests->begin(); it != _cancelingRequests->end();) {
+      auto request = it->second;
+      if (request != nullptr && request->getStatus() != UCS_INPROGRESS) {
+        it = _cancelingRequests->erase(it);
```
wence-, line 84: If `getStatus` is not `UCS_OK` (and not `UCS_INPROGRESS`) do we need to throw an exception, or is that handled elsewhere?

cpp/src/inflight_requests.cpp
```diff
@@ -21,11 +21,16 @@ void InflightRequests::insert(std::shared_ptr<Request> request)
...
+void InflightRequests::merge(InflightRequestsMapPtrPair inflightRequestsMapPtrPair)
 {
-  std::lock_guard<std::mutex> lock(_mutex);
-
-  _inflightRequests->merge(*inflightRequestsMap);
+  {
+    std::lock_guard<std::mutex> lock(_mutex);
+    _inflightRequests->merge(*(inflightRequestsMapPtrPair.first));
+  }
+  {
```
wence-, line 30: **question:** What are the preconditions on the two maps to this function?

Is it the case that every to-cancel request must have been already been observed? Is it allowed to have entries in the "new" requests that are also in the "to-cancel" requests?

pentschev: > Is it the case that every to-cancel request must have been already been observed?

Yes, a request will only be moved to the `_trackedRequests->_canceling` if it was previously in `_trackedRequests->_inflight` and `cancelAll()` was called, see [this block](https://github.com/rapidsai/ucxx/pull/153/files#diff-5f8b26cd809dc146d0f147fe15bee78cf7a1d9ac5898f09e4fa733f34710fc51R109-R133).

> Is it allowed to have entries in the "new" requests that are also in the "to-cancel" requests?

No, the only place where `_trackedRequests->_canceling` is populated is the `cancelAll()` block linked above, thus a request cannot go into "canceling" state right away. The exception is when merging requests, the "canceling requests" will still be "canceling requests" when we merge `InflightRequests` from an `Endpoint` into that of the `Worker`, but in that case they were already being canceled so it's not as if they were immediately registered in a "canceling" state when `Endpoint::registerInflightRequest()` was first called.

cpp/include/ucxx/inflight_requests.h
pentschev: Agreed, done in https://github.com/rapidsai/ucxx/pull/153/commits/2d78accd21d576bd6f426d0bb579ed3179fc370e .

cpp/src/inflight_requests.cpp
pentschev: No, `merge` takes an object to merge from and not a pointer to the object. Move would be appropriate if we were instead purely replacing `_trackedRequest->_inflight` with it.

cpp/src/inflight_requests.cpp
pentschev: No, this is all handled by `ucxx::Request`, here we're only tracking what to do with its internal reference, preventing the object from going out-of-scope and causing a segfault when the UCP request callback is executed if the user released it too early.

cpp/src/inflight_requests.cpp
wence-: OK, thanks!

wence- (approved):
Thanks Peter!

pentschev:
Thanks for reviewing this one too, @wence- !

PR #156, opened 2024-01-10
Improve log messages

pentschev:
Logging is a bit unorganized, having different formats and not always including useful information (such as the address of the object being logged). This change attempts to make logging a bit more descriptive and consistent where possible.

### DIFF
--- a/cpp/python/src/future.cpp
+++ b/cpp/python/src/future.cpp
@@ -60,13 +60,18 @@ static PyObject* get_asyncio_future_object()
   }
 
   asyncio_module = PyImport_Import(asyncio_str);
-  if (PyErr_Occurred()) ucxx_trace_req("Python error here");
+  if (PyErr_Occurred()) {
+    ucxx_trace_req("Error importing asyncio");
+    PyErr_Print();
+  }
   if (PyErr_Occurred()) PyErr_Print();
   if (asyncio_module == NULL) goto finish;
 
   asyncio_future_object = PyObject_GetAttr(asyncio_module, future_str);
-  if (PyErr_Occurred()) ucxx_trace_req("Python error here");
-  if (PyErr_Occurred()) PyErr_Print();
+  if (PyErr_Occurred()) {
+    ucxx_trace_req("Error getting asyncio.Future method");
+    PyErr_Print();
+  }
   Py_DECREF(asyncio_module);
   if (asyncio_future_object == NULL) { goto finish; }
 
@@ -98,8 +103,10 @@ PyObject* create_python_future()
   }
 
   result = PyObject_CallFunctionObjArgs(future_object, NULL);
-  if (PyErr_Occurred()) ucxx_trace_req("Python error here");
-  if (PyErr_Occurred()) PyErr_Print();
+  if (PyErr_Occurred()) {
+    ucxx_trace_req("Error creating asyncio.Future");
+    PyErr_Print();
+  }
 
 finish:
   PyGILState_Release(state);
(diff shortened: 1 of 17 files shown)


cpp/python/src/notifier.cpp
```diff
@@ -23,16 +23,17 @@ Notifier::~Notifier() {}
 
 void Notifier::scheduleFutureNotify(std::shared_ptr<::ucxx::Future> future, ucs_status_t status)
 {
-  ucxx_trace_req(
-    "Notifier::scheduleFutureNotify(): future: %p, handle: %p", future.get(), future->getHandle());
+  ucxx_trace_req("ucxx::python::Notifier::scheduleFutureNotify, future: %p, handle: %p",
```
wence-, line 26: Note to switch to `std::source_location::current().function_name()` once using C++20. Although that does seem to include the signature, which might not be ideal :(.

wence- (approved):

cpp/python/src/notifier.cpp
```diff
@@ -23,16 +23,17 @@ Notifier::~Notifier() {}
...
+                 future->getHandle());
   auto p = std::make_pair(future, status);
   {
     std::lock_guard<std::mutex> lock(_notifierThreadMutex);
     _notifierThreadFutureStatus.push_back(p);
     _notifierThreadFutureStatusReady = true;
   }
   _notifierThreadConditionVariable.notify_one();
-  ucxx_trace_req("Notifier::scheduleFutureNotify() notified: future: %p, handle: %p",
+  ucxx_trace_req("ucxx::python::Notifier::scheduleFutureNotify, notified: future: %p, handle: %p",
```
charlesbluca, line 36:
```suggestion
  ucxx_trace_req("ucxx::python::Notifier::scheduleFutureNotify, notified future: %p, handle: %p",
```

cpp/python/src/future.cpp
```diff
@@ -60,13 +60,18 @@ static PyObject* get_asyncio_future_object()
   }
 
   asyncio_module = PyImport_Import(asyncio_str);
-  if (PyErr_Occurred()) ucxx_trace_req("Python error here");
+  if (PyErr_Occurred()) {
+    ucxx_trace_req("Error importing asyncio");
```
charlesbluca, line 64: Would it make sense to include the function name here?


```suggestion
    ucxx_trace_req("ucxx::python::get_asyncio_future_object, error importing asyncio");
```

Same question for other added logging here

cpp/python/src/python_future_task_collector.cpp
```diff
@@ -34,7 +34,8 @@ void PythonFutureTaskCollector::collect()
     std::lock_guard<std::mutex> lock(_mutex);
     for (auto& handle : _toCollect)
       Py_XDECREF(handle);
-    ucxx_trace("Collected %lu PythonFutureTasks", _toCollect.size());
+    ucxx_trace("PythonFutureTaskCollector::collect, collected %lu PythonFutureTasks",
```
charlesbluca, line 37: Might be missing some nuance of how this plugs into the Python/Cython API, would we want to fully qualify this?


```suggestion
    ucxx_trace("ucxx::python::PythonFutureTaskCollector::collect, collected %lu PythonFutureTasks",
```

cpp/python/src/notifier.cpp
pentschev: Yeah, exactly, plus we're still on C++17 for now at least. 🙁 

Not the perfect solution but we can use `__func__` for at least part of this, done in https://github.com/rapidsai/ucxx/pull/156/commits/124efaedeee0b1e1b0c32d7424c7272a06423b92 .

cpp/python/src/notifier.cpp
pentschev: Good catch, fixed in https://github.com/rapidsai/ucxx/pull/156/commits/2f4317c5b0fd200669cfbc821d54c48dde98a263 .

cpp/python/src/future.cpp
pentschev: I had intentionally omitted those here because messages in those cases are unique enough they can't be misinterpreted. But yeah, I agree now we should do that to keep things more consistent. Fixed in https://github.com/rapidsai/ucxx/pull/156/commits/f9fa937fc23b34e9aa179b8727fe93881ee763e1 .

cpp/python/src/python_future_task_collector.cpp
pentschev: Same as above, but fixed it the same way in https://github.com/rapidsai/ucxx/pull/156/commits/f9fa937fc23b34e9aa179b8727fe93881ee763e1 as well.

charlesbluca (approved):

pentschev:
@wence- @charlesbluca thank you both for the reviews! 😄

PR #157, opened 2024-01-10
Remove invalid install locations

vyasr:
UCXX has no generated headers in the build directory, so these incorrect includes result in a build directory that is not directly usable as a dependency.


pentschev (approved):
LGTM, thanks @vyasr !

PR #158, opened 2024-01-10
Replace Python getter functions with properties in sync API

pentschev:
To make `libucxx.pyx` more consistent and Pythonic, replace various getter functions with `@property`.

### DIFF
--- a/python/ucxx/_lib/libucxx.pyx
+++ b/python/ucxx/_lib/libucxx.pyx
@@ -1207,7 +1207,7 @@ cdef class UCXEndpoint():
             handle = self._endpoint.get().getHandle()
 
         return int(<uintptr_t>handle)
-    
+
     @property
     def ucxx_ptr(self):
         cdef Endpoint* endpoint
@@ -1216,7 +1216,7 @@ cdef class UCXEndpoint():
             endpoint = self._endpoint.get()
 
         return int(<uintptr_t>endpoint)
-            
+
     @property
     def worker_handle(self):
         cdef ucp_worker_h handle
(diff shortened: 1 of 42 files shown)


python/ucxx/_lib/libucxx.pyx
```diff
@@ -527,9 +548,27 @@ cdef class UCXWorker():
...
+
+    @property
+    def enable_python_future(self):
+        return self._enable_python_future
+
+    def get_address(self):
+        warnings.warn(
+            "UCXWorker.get_address() is deprecated and will soon be removed, "
+            "use the UCXWorker.address property instead",
+            DeprecationWarning,
```
wence-, line 567: Just to note that if we want these to be visible to the user, they should be `FutureWarning` not `DeprecationWarning` (see https://docs.rapids.ai/api/cudf/stable/developer_guide/contributing_guide/#deprecating-and-removing-code and https://docs.python.org/3/library/warnings.html#warning-categories)

wence- (approved):
Minor suggestion to change `DeprecationWarning` to `FutureWarning`, otherwise all great AFAICT!

python/ucxx/_lib/libucxx.pyx
pentschev: Good point, fixed in https://github.com/rapidsai/ucxx/pull/158/commits/62d00b4829d96bea4542a0de8cc8934072c2ee6b

charlesbluca (approved):

pentschev:
Thanks @charlesbluca and @wence- !

PR #159, opened 2024-01-10
refactor CUDA versions in dependencies.yaml

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/7.

Proposes splitting the `cuda-version` dependency in `dependencies.yaml` out to its own thing, separate from the bits of the CUDA Toolkit this project needs.

### Benefits of this change

* prevents accidental inclusion of multiple `cuda-version` version in environments
* reduces update effort (via enabling more use of globs like `"12.*"`)
* improves the chance that errors like "`conda` recipe is missing a dependency" are caught in CI

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -9,7 +9,8 @@ files:
       - build_cpp
       - build_python
       - checks
-      - cudatoolkit
+      - cuda
+      - cuda_version
       - dev
       - py_version
       - run_python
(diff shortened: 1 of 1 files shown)


pentschev (approved):

jakirkham (approved):
Thanks James! 🙏

PR #160, opened 2024-01-11
Do not strip shared library in debug mode with scikit-build

pentschev:
To ensure debug symbols remain after building `libucxx.pyx` Cython library in debug mode with scikit-build `SKBUILD_INSTALL_STRIP=false` must be defined.


pentschev:
@wence- would you mind approving if you have no objections?

wence- (approved):

vyasr:
@pentschev if you think this is a worthwhile change, could you make similar modifications to other build scripts in RAPIDS?

jakirkham:
Maybe we can just file an issue on [rapidsai/build-planning]( https://github.com/rapidsai/build-planning ) and evaluate on a project-by-project basis

vyasr:
Agreed. @jakirkham would you mind filing that issue?

PR #165, opened 2024-01-16
Prevent pointer conversion within `@classmethod` and improve Cython typing

pentschev:
Converting `shared_ptr`->`void*`->`uintptr_t` is a very cumbersome and error prone solution. By instantiating the object within the `@classmethod` scope we can thus prevent calling the constructor again where the conversion is necessary. The downside is we now must set member attributes also in the scope of `@classmethod`, but that should be a better tradeoff to the pointer conversion.

Improve static typing in Cython where missing types were missing (exceptions are `cb_func` as it seems typing functions is not possible/trivial with Cython), as well as making `UCXXTag`/`UCXXTagMask` Cython classes so that they can be used as Cython types.


python/ucxx/_lib/tag.pyx
```diff
@@ -0,0 +1,24 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024, NVIDIA CORPORATION & AFFILIATES.
+# SPDX-License-Identifier: BSD-3-Clause
+
+
+from libc.stdint cimport uint64_t
+
+
+cdef class UCXXTag:
+    cdef:
+        readonly uint64_t value
```
charlesbluca, line 10: Do we need this indentation? From the [Cython docs](https://cython.readthedocs.io/en/stable/src/userguide/extension_types.html#static-attributes), I would've thought we could just do


```suggestion
    cdef readonly uint64_t value
```

python/ucxx/_lib/tag.pyx
```diff
@@ -0,0 +1,24 @@
...
+
+    def __cinit__(self, uint64_t tag) -> None:
+        self.value = tag
+
+
+cdef class UCXXTagMask:
+    cdef:
+        readonly uint64_t value
```
pentschev, line 18:
```suggestion
    cdef readonly uint64_t value
```

python/ucxx/_lib/tag.pyx
pentschev: In this case here we don't need it because it's a single variable, it gets more useful when you have multiple variables.

charlesbluca: That makes sense to me - in that case, I'll leave this up to your preference, just wanted to make sure I wasn't missing something 🙂

python/ucxx/_lib/libucxx.pyx
```diff
@@ -420,22 +403,31 @@ cdef class UCXAddress():
 
     # For old UCX-Py API compatibility
     @classmethod
-    def from_worker(cls, UCXWorker worker):
+    def from_worker(cls, UCXWorker worker) -> UCXAddress:
+        cdef UCXAddress address = UCXAddress.__new__(UCXAddress)
+
+        warnings.warn(
+            "UCXAddress.from_worker() is deprecated and will soon be removed, "
+            "use UCXWorker.create_from_worker() instead",
```
charlesbluca, line 411:
```suggestion
            "use UCXAddress.create_from_worker() instead",
```

pentschev: Good catch, the warning is wrong and your suggestion is what it should read. Accessing through `UCXWorker.address` is merely a convenience and they're 100% equivalent, in fact `UCXWorker.address` calls `UCXAddress.create_from_worker()`, as you've probably seen already.

charlesbluca (approved):

pentschev:
Thanks for the reviews and suggestions here @charlesbluca !

PR #166, opened 2024-01-16
Add support for functions `ucp_get_nbx` and `ucp_put_nbx`

mdemoret-nv:
This PR is a work in progress.

This PR adds support for 2 new UCX functions: `ucp_get_nbx` and `ucp_put_nbx`. The implementation follows the pattern established by the tagged send/receive functions.

There is still a lot of commented out code and no tests have been added. The new functions have only been tested locally in a feature branch for MRC.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -77,6 +77,7 @@ rapids_cmake_support_conda_env(conda_env MODIFY_PREFIX_PATH)
 # * compiler options ------------------------------------------------------------------------------
 rapids_find_package(
   ucx REQUIRED
+  GLOBAL_TARGETS ucx::ucp ucx::ucs ucx::uct
   BUILD_EXPORT_SET ucxx-exports
   INSTALL_EXPORT_SET ucxx-exports
 )
@@ -119,6 +120,7 @@ add_library(
   src/request.cpp
   src/request_am.cpp
   src/request_helper.cpp
+  src/request_mem.cpp
   src/request_stream.cpp
   src/request_tag.cpp
   src/request_tag_multi.cpp
@@ -149,9 +151,9 @@ target_compile_options(
 # Specify include paths for the current target and dependents
 target_include_directories(
   ucxx
-  PUBLIC "$<BUILD_INTERFACE:${UCXX_SOURCE_DIR}/include>"
+  PUBLIC "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>"
          "$<BUILD_INTERFACE:${UCXX_GENERATED_INCLUDE_DIR}/include>"
-  PRIVATE "$<BUILD_INTERFACE:${UCXX_SOURCE_DIR}/src>"
+  PRIVATE "$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/src>"
   INTERFACE "$<INSTALL_INTERFACE:include>"
 )
 
--- a/cpp/include/ucxx/constructors.h
+++ b/cpp/include/ucxx/constructors.h
@@ -20,6 +20,7 @@ class Listener;
 class Notifier;
 class Request;
 class RequestAm;
+class RequestMem;
 class RequestStream;
 class RequestTag;
 class RequestTagMulti;
@@ -73,6 +74,16 @@ std::shared_ptr<RequestStream> createRequestStream(std::shared_ptr<Endpoint> end
                                                    void* buffer,
                                                    size_t length,
                                                    const bool enablePythonFuture);
+
+std::shared_ptr<RequestMem> createRequestMem(std::shared_ptr<Endpoint> endpoint,
+                                             bool send,
+                                             void* buffer,
+                                             size_t length,
+                                             uint64_t remote_addr,
+                                             ucp_rkey_h rkey,
+                                             const bool enablePythonFuture,
+                                             RequestCallbackUserFunction callbackFunction,
+                                             RequestCallbackUserData callbackData);
 
 std::shared_ptr<RequestTag> createRequestTag(std::shared_ptr<Component> endpointOrWorker,
                                              bool send,
--- a/cpp/src/request_mem.cpp
+++ b/cpp/src/request_mem.cpp
@@ -1,0 +1,293 @@
+/**
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2023, NVIDIA CORPORATION & AFFILIATES.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+#include <cstdio>
+#include <memory>
+#include <string>
+
+#include <ucp/api/ucp.h>
+
+#include <ucxx/buffer.h>
+#include <ucxx/delayed_submission.h>
+#include <ucxx/request_mem.h>
+
+namespace ucxx {
+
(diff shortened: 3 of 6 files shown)


cpp/CMakeLists.txt
```diff
@@ -77,6 +77,7 @@ rapids_cmake_support_conda_env(conda_env MODIFY_PREFIX_PATH)
 # * compiler options ------------------------------------------------------------------------------
 rapids_find_package(
   ucx REQUIRED
+  GLOBAL_TARGETS ucx::ucp ucx::ucs ucx::uct
```
pentschev, line 80: I think such changes shouldn't be necessary, you mentioned offline they were required to build from inside another project. We'll have our CMake experts look at what would need to change here when the PR is ready for review.

cpp/include/ucxx/constructors.h
```diff
@@ -74,6 +75,16 @@ std::shared_ptr<RequestStream> createRequestStream(std::shared_ptr<Endpoint> end
                                                    size_t length,
                                                    const bool enablePythonFuture);
 
+std::shared_ptr<RequestMem> createRequestMem(std::shared_ptr<Endpoint> endpoint,
+                                             bool send,
+                                             void* buffer,
+                                             size_t length,
+                                             uint64_t remote_addr,
+                                             ucp_rkey_h rkey,
```
pentschev, line 83: As discussed offline, the changes were made on top of an older branch before https://github.com/rapidsai/ucxx/pull/121 was merged (on December 12), which reorganizes how data is passed to constructors and thus prevents adding several different arguments to them as is done [here](https://github.com/rapidsai/ucxx/pull/166/files#diff-99229522298ec9b3fb9e931133be93d2189a06e0cea3a180df26c5b0433556e4R54-R58), we now instead have [“data classes”](https://github.com/rapidsai/ucxx/blob/branch-0.36/cpp/include/ucxx/request_data.h) that include all operation-specific arguments and makes the API cleaner and less error-prone (i.e., with the user accidentally swapping arguments), as you can see in the [newest constructors](https://github.com/rapidsai/ucxx/blob/branch-0.36/cpp/include/ucxx/request_tag.h#L83). We will want to have this updated to the new style.

pentschev (commented):
In a first pass, things look like they're going in the right direction, I've left a couple of comments. It would be great if you could also add some tests.

cpp/include/ucxx/request_flush.h
```diff
@@ -0,0 +1,126 @@
...
+   * @param[in] enablePythonFuture  whether a python future should be created and
+   *                                subsequently notified.
+   * @param[in] callbackFunction    user-defined callback function to call upon completion.
+   * @param[in] callbackData        user-defined data to pass to the `callbackFunction`.
+   *
+   * @returns The `shared_ptr<ucxx::RequestFlush>` object
+   */
+  friend std::shared_ptr<RequestFlush> createRequestFlush(
+    std::shared_ptr<Component> endpointOrWorker,
+    const std::variant<data::Flush> requestData,
```
mdemoret-nv, line 91: How come you are using a variant with a single type? Is this to follow a pattern?

cpp/src/request_mem.cpp
```diff
@@ -0,0 +1,196 @@
...
+      [&endpoint, &enablePythonFuture, &callbackFunction, &callbackData](data::MemPut memPut) {
+        return std::shared_ptr<RequestMem>(new RequestMem(
+          endpoint, memPut, "memPut", enablePythonFuture, callbackFunction, callbackData));
+      },
+      [&endpoint, &enablePythonFuture, &callbackFunction, &callbackData](data::MemGet memGet) {
+        return std::shared_ptr<RequestMem>(new RequestMem(
+          endpoint, memGet, "memGet", enablePythonFuture, callbackFunction, callbackData));
+      },
+    },
+    requestData);
```
mdemoret-nv, line 35: Because you are doing a `std::visit` inside of the constructor, couldnt this be simplified to just:
```cpp
std::shared_ptr<RequestMem> req = std::shared_ptr<RequestMem>(new RequestMem(
          endpoint, requestData, isPut ? "memPut" : "memGet", enablePythonFuture, callbackFunction, callbackData));
```
And even the `isPut ? "memPut" : "memGet"`` could be removed and set in the dispatch inside of the constructor since those are the only two options. Seems a bit redundant to have a `std::visit` followed by another `std::visit`

cpp/src/request_mem.cpp
```diff
@@ -0,0 +1,196 @@
...
+               [this](data::MemGet memGet) {
+                 if (_endpoint == nullptr)
+                   throw ucxx::Error("A valid endpoint is required to receive memory messages.");
+               },
+               [](auto) { throw std::runtime_error("Unreachable"); },
+             },
+             requestData);
+
+  _callback     = callbackFunction;
+  _callbackData = callbackData;
```
mdemoret-nv, line 68: Why defer setting these? Couldnt you just add `_callback(std::move(callbackFunction))` after the base constructor?

cpp/src/request_mem.cpp
```diff
@@ -0,0 +1,196 @@
...
+                 [this](data::MemGet memGet) {
+                   if (_worker->getHandle() == nullptr) {
+                     ucxx_warn("Endpoint was closed before message could be received");
+                     Request::callback(this, UCS_ERR_CANCELED);
+                     return true;
+                   }
+                   return false;
+                 },
+                 [](auto) -> decltype(terminate) { throw std::runtime_error("Unreachable"); },
+               },
```
mdemoret-nv, line 146: Both of these branches are the same

cpp/src/request_mem.cpp
pentschev: From a simplistic point-of-view, you're right. However, what we're attempting to solve with `std::variant`/`std::visit` is to provide strong-typing, and thus prevent things like the caller from mistakenly shuffling arguments. I agree this adds some overhead that is unnecessary from a functional standpoint, but a "safe" interface is generally preferable. An alternative would be adding a type for `isPut` that is not a `bool` which would be fine for this two-case scenario, but that is another type of complexity we would need to add so there would be a tradeoff as well that I feel would be more confusing, even though it could slightly decrease code length.

cpp/src/request_mem.cpp
pentschev: The reason is this is a member of the base class, which can't be part of the constructor initializer list and thus we must add those to the base class constructor. This is the same in other `Request*` types which are on my plans to be done, but I haven't got there yet.

cpp/src/request_mem.cpp
pentschev: They're not, the warning message are different. The general purpose is anyway to ensure we treat all supported types with their own branches, and `throw` on unsupported types, thus providing strong typing.

cpp/src/request_mem.cpp
pentschev: I started https://github.com/rapidsai/ucxx/pull/194 to do that, leaving as draft until this PR and https://github.com/rapidsai/ucxx/pull/192 are merged because both use this pattern.

pentschev (approved):

pentschev:
Thanks @mdemoret-nv for working on this!

PR #167, opened 2024-01-17
add build wheel script and accompanying version info

### LINKED ISSUE #145: Build and publish wheels for UCXX
We're getting close to making the move from UCX-Py to UCXX, PRs such as https://github.com/rapidsai/raft/pull/1983 and https://github.com/rapidsai/cuml/pull/5697 are integrating UCXX to run tests in CI. The plan is, time permitting, to have RAPIDS projects running UCXX tests in CI for 24.02 and possibly switching to UCXX as the default in 24.04 (still pending approval). However, we cannot make that move without also publishing wheels.

@vyasr would you be able to work on wheels for UCXX sometime before 24.02? I'll be OOO until early January, so if you don't have the bandwidth to work on it I may need your help then, unless @charlesbluca or @wence- have some bandwidth, more importantly, want to give it a go at this. 🙂

msarahan:
This PR adds wheel builds for `ucxx`. A follow-up PR will add wheel builds for `distributed-ucxx`.

Closes #145


python/CMakeLists.txt
```diff
@@ -31,6 +31,7 @@ include(rapids-cython-core)
 if(NOT ucxx_FOUND)
   set(BUILD_TESTS OFF)
   set(BUILD_BENCHMARKS OFF)
+  set(UCXX_ENABLE_PYTHON ON)
```
pentschev, line 34: I think this won't have any effect here as this flag is part of the `cpp/CMakeLists.txt` to build CPython code from `cpp/python/`.

cpp/CMakeLists.txt
```diff
@@ -157,17 +161,15 @@ target_compile_definitions(
...
-    target_compile_definitions(ucxx PUBLIC UCXX_ENABLE_RMM)
+    target_link_libraries(ucxx PUBLIC rmm::rmm)
+    target_compile_definitions(ucxx PUBLIC UCXX_ENABLE_RMM "SPDLOG_ACTIVE_LEVEL=SPDLOG_LEVEL_${RMM_LOGGING_LEVEL}")
 endif()
 
 # Define spdlog level
-target_compile_definitions(ucxx PUBLIC "SPDLOG_ACTIVE_LEVEL=SPDLOG_LEVEL_${RMM_LOGGING_LEVEL}")
 
 # Specify the target module library dependencies
-target_link_libraries(
```
pentschev, line 167: So we remove RMM because it compiles from the header-only bits?

python/CMakeLists.txt
vyasr: This flag will have an effect for wheel builds, which trigger a build of the C++ from the Python build.

cpp/CMakeLists.txt
vyasr: We don't remove it, I just linked it conditionally in the block above on line 164.

pentschev: Ah yes, missed that. Thanks!

python/CMakeLists.txt
pentschev: > This flag will have an effect for wheel builds, which trigger a build of the C++ from the Python build

Ok, this makes sense.

> or more generally in the case when the Python package build cannot find a pre-built libucxx

I'm not sure about what this means though, what if a pre-built libucxx isn't found? Is that essentially then saying "build libucxx too"?

vyasr: Yes, that's right. That's how all our packages are set up. That way in theory someone could take the source and do a pip install and it would just work. They don't have to know to build the cpp lib first. Of course, in most cases we work with we want to build the cpp lib first and find it.

pentschev: Oh I see, that's pretty clever. Thanks for the details!

ci/build_wheel.sh
```diff
@@ -0,0 +1,46 @@
...
+# Add -cuXX to package name
+sed -r -i "s/rmm(.*)\"/rmm${PACKAGE_CUDA_SUFFIX}\1${alpha_spec}\"/g" ${pyproject_file}
+
+SKBUILD_CMAKE_ARGS="-DUCXX_ENABLE_RMM=ON" \
+    python -m pip wheel "${package_dir}"/ -w "${package_dir}"/dist -vvv --no-deps --disable-pip-version-check
+
+python -m auditwheel repair -w ${package_dir}/final_dist ${package_dir}/dist/*
+
+RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
+RAPIDS_PY_WHEEL_NAME="ucxx_${AUDITWHEEL_POLICY}_${RAPIDS_PY_CUDA_SUFFIX}" rapids-upload-wheels-to-s3 ${package_dir}/final_dist
```
vyasr, line 46:
```suggestion
RAPIDS_PY_WHEEL_NAME="ucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-upload-wheels-to-s3 ${package_dir}/final_dist
```

ci/build_wheel.sh
```diff
@@ -0,0 +1,46 @@
...
+# easiest way to do so is to augment the spec with a constraint containing a
+# min alpha version that doesn't affect the version bounds but does allow usage
+# of alpha versions for that dependency without --pre
+alpha_spec=''
+if ! rapids-is-release-build; then
+    alpha_spec=',>=0.0.0a0'
+fi
+
+# Add -cuXX to package name
+sed -r -i "s/rmm(.*)\"/rmm${PACKAGE_CUDA_SUFFIX}\1${alpha_spec}\"/g" ${pyproject_file}
```
vyasr, line 38:
```suggestion
sed -r -i "s/rmm(.*)\"/rmm${PACKAGE_CUDA_SUFFIX}\1${alpha_spec}\"/g" ${pyproject_file}
sed -r -i "s/cudf(.*)\"/cudf${PACKAGE_CUDA_SUFFIX}\1${alpha_spec}\"/g" ${pyproject_file}
```

pentschev:
@vyasr do we have a guideline on what we normally test/should test for wheels builds? Are other projects running the full test set or only a subset of everything that runs for conda? For now I've only added subset of the more common uses as reference.

ci/test_utils.sh
```diff
@@ -20,5 +20,13 @@ print_system_stats() {
 
 print_ucx_config() {
   rapids-logger "UCX Version and Build Configuration"
-  ucx_info -v
+
+  set +e
+  which ucx_info > /dev/null
+  set -e
+  if [ $? -eq 0 ]; then
```
charlesbluca, line 27: Does it make sense to move these set commands to the wheel test script? Assuming that `set -e` is throwing off the `$?` check immediately below.

Alternatively, could probably go back to the check we were using in https://github.com/rapidsai/ucxx/pull/167/commits/5f8b0c9b42ac080c3296c164268c9c2fc6aa17c8, storing the value of `which ucx_info` in a variable:

```diff
diff --git a/ci/test_utils.sh b/ci/test_utils.sh
index df443a2..ac8c4f4 100755
--- a/ci/test_utils.sh
+++ b/ci/test_utils.sh
@@ -22,11 +22,11 @@ print_ucx_config() {
   rapids-logger "UCX Version and Build Configuration"
 
   set +e
-  which ucx_info > /dev/null
+  UCX_INFO_PATH=$(which ucx_info > /dev/null)
   set -e
-  if [ $? -eq 0 ]; then
-    ucx_info -v
-  else
+  if [ $UCX_INFO_PATH == "" ]; then
     echo "ucx_info not found"
+  else
+    ucx_info -v
   fi
 }
```

pentschev: > Does it make sense to move these set commands to the wheel test script? Assuming that `set -e` is throwing off the `$?` check immediately below.

I'm doing it like this now because I don't know exactly what's included in the wheels. The ideal scenario is that `ucx_info` would be available via the wheels install as well, so I'm not yet sure whether we're missing UCX entirely or if it's just `ucx_info` and want to see first if we can get tests running. I'm guessing @vyasr will be able to tell us what we can or can't do with the wheels sometime later today.

> Alternatively, could probably go back to the check we were using in [5f8b0c9](https://github.com/rapidsai/ucxx/commit/5f8b0c9b42ac080c3296c164268c9c2fc6aa17c8), storing the value of `which ucx_info` in a variable:

There's actually a much simpler solution that I overlooked initially, see https://github.com/rapidsai/ucxx/pull/167/commits/1da02e6014066c958cdd9822648d15d8dfe4b8d8 .

vyasr:
@pentschev our standard approach is to run the full test suite on x86 machines but only run a small smoke test on arm due to the relative sparsity of arm runners. For example, here's cudf's test script: https://github.com/rapidsai/cudf/blob/branch-24.02/ci/test_wheel_cudf.sh#L25.

pentschev:
> @pentschev our standard approach is to run the full test suite on x86 machines but only run a small smoke test on arm due to the relative sparsity of arm runners. For example, here's cudf's test script: https://github.com/rapidsai/cudf/blob/branch-24.02/ci/test_wheel_cudf.sh#L25.

But you don't run C++ (`libcudf`) tests then, is that right? And is there a reason for not running them (runtime costs, complexity, building, ...)?

vyasr:
> But you don't run C++ (`libcudf`) tests then, is that right? And is there a reason for not running them (runtime costs, complexity, building, ...)?

- We don't want to build the C++ tests in wheel jobs, nor do we want to bundle the test executables into the wheel.
- It's redundant to existing C++ test jobs since there shouldn't be differences in the C++ (up to linking), so it's unnecessarily expensive from a CI perspective.
- From a conceptual standpoint you'd effectively be testing internals since wheels don't expose the C++ API.

ci/test_common.sh
```diff
@@ -0,0 +1,155 @@
...
+    exit $LAST_STATUS
+  fi
+}
+
+
+#################################### Python ####################################
+run_py_tests() {
+  CMD_LINE="timeout 2m pytest -vs python/ucxx/_lib/tests/"
+  log_command "${CMD_LINE}"
+  timeout 2m pytest -vs python/ucxx/_lib/tests/
```
charlesbluca, line 95:
```suggestion
  timeout 2m python -m pytest -vs python/ucxx/_lib/tests/
```

ci/test_common.sh
```diff
@@ -0,0 +1,155 @@
...
+  if [ ${LAST_STATUS} -ne 0 ]; then
+    echo "Failure running benchmark client after ${MAX_ATTEMPTS} attempts"
+    exit $LAST_STATUS
+  fi
+}
+
+
+#################################### Python ####################################
+run_py_tests() {
+  CMD_LINE="timeout 2m pytest -vs python/ucxx/_lib/tests/"
```
charlesbluca, line 93:
```suggestion
  CMD_LINE="timeout 2m python -m pytest -vs python/ucxx/_lib/tests/"
```

ci/test_common.sh
```diff
@@ -0,0 +1,155 @@
...
+  timeout 2m pytest -vs python/ucxx/_lib/tests/
+}
+
+run_py_tests_async() {
+  PROGRESS_MODE=$1
+  ENABLE_DELAYED_SUBMISSION=$2
+  ENABLE_PYTHON_FUTURE=$3
+  SKIP=$4
+
+  CMD_LINE="UCXPY_PROGRESS_MODE=${PROGRESS_MODE} UCXPY_ENABLE_DELAYED_SUBMISSION=${ENABLE_DELAYED_SUBMISSION} UCXPY_ENABLE_PYTHON_FUTURE=${ENABLE_PYTHON_FUTURE} timeout 20m pytest -vs python/ucxx/_lib_async/tests/ --durations=50"
```
charlesbluca, line 104:
```suggestion
  CMD_LINE="UCXPY_PROGRESS_MODE=${PROGRESS_MODE} UCXPY_ENABLE_DELAYED_SUBMISSION=${ENABLE_DELAYED_SUBMISSION} UCXPY_ENABLE_PYTHON_FUTURE=${ENABLE_PYTHON_FUTURE} timeout 20m python -m pytest -vs python/ucxx/_lib_async/tests/ --durations=50"
```

ci/test_common.sh
```diff
@@ -0,0 +1,155 @@
...
+
+  if [ $SKIP -ne 0 ]; then
+    echo -e "\e[1;33mSkipping unstable test: ${CMD_LINE}\e[0m"
+  else
+    log_command "${CMD_LINE}"
+    UCXPY_PROGRESS_MODE=${PROGRESS_MODE} UCXPY_ENABLE_DELAYED_SUBMISSION=${ENABLE_DELAYED_SUBMISSION} UCXPY_ENABLE_PYTHON_FUTURE=${ENABLE_PYTHON_FUTURE} timeout 20m pytest -vs python/ucxx/_lib_async/tests/ --durations=50
```
charlesbluca, line 110:
```suggestion
    UCXPY_PROGRESS_MODE=${PROGRESS_MODE} UCXPY_ENABLE_DELAYED_SUBMISSION=${ENABLE_DELAYED_SUBMISSION} UCXPY_ENABLE_PYTHON_FUTURE=${ENABLE_PYTHON_FUTURE} timeout 20m python -m pytest -vs python/ucxx/_lib_async/tests/ --durations=50
```

ci/test_common.sh
```diff
@@ -0,0 +1,155 @@
...
+  UCX_KEEPALIVE_INTERVAL=1ms UCXPY_ENABLE_DELAYED_SUBMISSION=${ENABLE_DELAYED_SUBMISSION} UCXPY_ENABLE_PYTHON_FUTURE=${ENABLE_PYTHON_FUTURE} timeout 2m python -m ucxx.benchmarks.send_recv --backend ${BACKEND} -o cupy --reuse-alloc -n 8MiB --n-buffers $N_BUFFERS --progress-mode ${PROGRESS_MODE} ${ASYNCIO_WAIT}
+}
+
+################################## Distributed #################################
+run_distributed_ucxx_tests() {
+  PROGRESS_MODE=$1
+  ENABLE_DELAYED_SUBMISSION=$2
+  ENABLE_PYTHON_FUTURE=$3
+
+  CMD_LINE="UCXPY_PROGRESS_MODE=${PROGRESS_MODE} UCXPY_ENABLE_DELAYED_SUBMISSION=${ENABLE_DELAYED_SUBMISSION} UCXPY_ENABLE_PYTHON_FUTURE=${ENABLE_PYTHON_FUTURE} timeout 10m pytest -vs python/distributed-ucxx/distributed_ucxx/tests/"
```
charlesbluca, line 148:
```suggestion
  CMD_LINE="UCXPY_PROGRESS_MODE=${PROGRESS_MODE} UCXPY_ENABLE_DELAYED_SUBMISSION=${ENABLE_DELAYED_SUBMISSION} UCXPY_ENABLE_PYTHON_FUTURE=${ENABLE_PYTHON_FUTURE} timeout 10m python -m pytest -vs python/distributed-ucxx/distributed_ucxx/tests/"
```

ci/test_common.sh
```diff
@@ -0,0 +1,155 @@
...
+
+  # Workaround for https://github.com/rapidsai/ucxx/issues/15
+  # CMD_LINE="UCX_KEEPALIVE_INTERVAL=1ms ${CMD_LINE}"
+
+  log_command "${CMD_LINE}"
+  UCXPY_PROGRESS_MODE=${PROGRESS_MODE} UCXPY_ENABLE_DELAYED_SUBMISSION=${ENABLE_DELAYED_SUBMISSION} UCXPY_ENABLE_PYTHON_FUTURE=${ENABLE_PYTHON_FUTURE} timeout 10m pytest -vs python/distributed-ucxx/distributed_ucxx/tests/
```
charlesbluca, line 154:
```suggestion
  UCXPY_PROGRESS_MODE=${PROGRESS_MODE} UCXPY_ENABLE_DELAYED_SUBMISSION=${ENABLE_DELAYED_SUBMISSION} UCXPY_ENABLE_PYTHON_FUTURE=${ENABLE_PYTHON_FUTURE} timeout 10m python -m pytest -vs python/distributed-ucxx/distributed_ucxx/tests/
```

pentschev:
I'm able to reproduce the segmentation faults locally and they only occur on tests using CUDA (cuDF, CuPy). UCX debug logs point to failure in finding `libuct_cuda`:

```
[1705599331.745637] [dgx14:14674:0]          module.c:259  UCX  DEBUG dlopen('/pyenv/versions/3.9.18/lib/python3.9/site-packages/ucxx/_lib/../../ucxx_cu11.libs/ucx/libuct_cuda.so.0.0.0', mode=0x1) failed: /pyenv/versions/3.9.18/lib/python3.9/site-packages/ucxx/_lib/../../ucxx_cu11.libs/ucx/libuct_cuda.so.0.0.0: cannot open shared object file: No such file or directory
```

I know we had similar issues on UCX-Py in the past, maybe you already know what we might be missing here @vyasr ?

vyasr:
> I'm able to reproduce the segmentation faults locally and they only occur on tests using CUDA (cuDF, CuPy). UCX debug logs point to failure in finding `libuct_cuda`:
> 
> ```
> [1705599331.745637] [dgx14:14674:0]          module.c:259  UCX  DEBUG dlopen('/pyenv/versions/3.9.18/lib/python3.9/site-packages/ucxx/_lib/../../ucxx_cu11.libs/ucx/libuct_cuda.so.0.0.0', mode=0x1) failed: /pyenv/versions/3.9.18/lib/python3.9/site-packages/ucxx/_lib/../../ucxx_cu11.libs/ucx/libuct_cuda.so.0.0.0: cannot open shared object file: No such file or directory
> ```
> 
> I know we had similar issues on UCX-Py in the past, maybe you already know what we might be missing here @vyasr ?

Yeah I was waiting for this issue to come up. We'll need to copy over the logic from ucx-py for bundling ucx itself.

vyasr:
@pentschev basically we need everything from here on down: https://github.com/rapidsai/ucx-py/blob/branch-0.36/ci/build_wheel.sh#L49

pentschev:
Thanks @vyasr , let me try that.

bdice:
@msarahan @pentschev @vyasr I have merged #169 into `branch-0.36`, and resolved dependency conflicts with this PR. I believe this PR's dependencies should be all good for ucxx wheels. We decided to split up this work and open a follow-up PR to add `distributed-ucxx` so those CI jobs are commented out for now, as @pentschev requested. I have rerun the failing jobs a couple times and I think the remaining failures should be investigated.

vyasr:
A large number of the outstanding errors seem to just be UCX requiring a higher ulimit, so can try bumping that and see what it buys us.

pentschev (approved):
All tests have passed now, this time I didn't have to rerun anything so I'm fairly confident that changes here are correct, thus I'm approving this.

@bdice @msarahan @vyasr thanks so much for your work here to get this through! 😄

raydouglass (changes requested):
This needs to be retargeted for `branch-0.37` then Ops can review again.

.github/workflows/build.yaml
```diff
@@ -43,3 +43,41 @@ jobs:
       branch: ${{ inputs.branch }}
       date: ${{ inputs.date }}
       sha: ${{ inputs.sha }}
+  wheel-build-ucxx:
+    secrets: inherit
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@branch-24.02
```
raydouglass, line 48: All of the workflow branches need to be updated to `branch-24.04`

conda/recipes/ucxx/meta.yaml
```diff
@@ -271,8 +271,7 @@ outputs:
         - tomli
       run:
         - python * *_cpython
-        - dask >=2023.9.2
-        - distributed >=2023.9.2
+        - rapids-dask-dependency =24.02
```
raydouglass, line 274:
```suggestion
        - rapids-dask-dependency =24.04
```

raydouglass (approved):

bdice:
@vyasr requested to review this before merging. I have added a `DO NOT MERGE` label until then. @vyasr, please remove that label once you approve. I have begun work on `distributed-ucxx` wheels in #171 so if we can punt on some of your suggested changes, I'm happy to address them in that second PR.

bdice:
This PR is incomplete -- I uncovered an issue in #171 where `ucxx` would not import without `rmm`, but `rmm` wasn't listed in the dependencies. We didn't observe this in testing the wheels because we have `cudf` as a `ucxx` testing dependency, which itself requires `rmm`. I would suggest that we get CI passing in #171 and merge those changes back into this PR, then review the whole thing (wheels for ucxx and distributed-ucxx) together.

vyasr:
I temporarily put this effort on pause since 1) wheels were pushed to 0.37 and 2) I was focusing on issues in https://github.com/rapidsai/raft/pull/1983/. I'll get back to this PR as soon as issues I uncovered in the raft/ucxx integration PR are fully fixed upstream.

vyasr:
We're going to wait on this PR until we can get the dependencies refactorings completed upstream.

pentschev:
> We're going to wait on this PR until we can get the dependencies refactorings completed upstream.

Which refactorings are you referring to? I'm assuming you're talking about something else and not the CMake as it won't make its way in for a while.

VERSION
```diff
@@ -0,0 +1 @@
+0.36.00
```
bdice, line 1:
```suggestion
0.37.00
```

bdice:
The dependencies refactor is #174, and it subsumes a lot of the diff of PRs 167 / 171, which will reduce PR conflicts. That one is almost ready to merge.

pentschev:
> The dependencies refactor is #174, and it subsumes a lot of the diff of PRs 167 / 171, which will reduce PR conflicts. That one is almost ready to merge.

I'm catching up today, thanks for all the work you all put here while I was gone!

Since #174 is now merged, where does that leave us? Are there any current blockers? Anything I can do to help pushing this further?

bdice:
@msarahan @vyasr I think this might be ready for review. Can you take a look? (@msarahan won't be able to approve since he opened the PR, but that's fine -- the contents have changed substantially since the early stages.)

ci/test_wheel_distributed_ucxx.sh
```diff
@@ -0,0 +1,39 @@
...
+
+# Install previously built ucxx wheel
+RAPIDS_PY_WHEEL_NAME="ucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 ./local-ucxx-dep
+python -m pip install ./local-ucxx-dep/ucxx*.whl
+
+# echo to expand wildcard before adding `[extra]` requires for pip
+python -m pip install $(echo ./dist/${PROJECT_NAME}*.whl)[test]
+
+# TODO: We need distributed installed in developer mode to provide test utils,
+# we still need to match to the `rapids-dask-dependency` version.
```
bdice, line 23: @pentschev Can you explain a bit more about this requirement, and how you would resolve the TODO here? It looks like this is a requirement to run distributed-ucxx tests, but it isn't available in the `distributed` package? Having a dependency on installing `distributed` from source (as opposed to something available in the wheel) might make it hard to reproduce CI failures in a local development environment.

ci/test_wheel_distributed_ucxx.sh
```diff
@@ -0,0 +1,39 @@
...
+rapids-logger "Install Distributed in developer mode"
+git clone https://github.com/dask/distributed /tmp/distributed
+python -m pip install -e /tmp/distributed
+
+# Run smoke tests for aarch64 pull requests
+# if [[ "$(arch)" == "aarch64" && "${RAPIDS_BUILD_TYPE}" == "pull-request" ]]; then
+#     python ci/wheel_smoke_test.py
+# else
+#     python -m pytest ./python/${PROJECT_NAME}/tests -k 'not test_sparse_pca_inputs' -n 4 --ignore=python/cuml/tests/dask && python -m pytest ./python/${PROJECT_NAME}/tests -k 'test_sparse_pca_inputs' && python -m pytest ./python/cuml/tests/dask
+# fi
```
bdice, line 33: It looks like we're running the full test suite on aarch64 and not just smoke tests. We run smoke tests instead of a full test suite for ARM wheels testing of pull requests for some RAPIDS libraries because we want to minimize ARM runner usage. We _do_ run the full test suite on ARM for conda packages and in the nightly builds for wheels, so we still have coverage (just not in PRs for wheels).

If we intend to run the full test suite on ARM for PRs, we should remove this commented section. Otherwise we should implement a minimal set of "smoke tests" in `ci/wheel_smoke_test.py`. This could be as simple as ensuring the package can be imported.

ci/test_wheel_ucxx.sh
```diff
@@ -0,0 +1,32 @@
...
+source "$(dirname "$0")/test_common.sh"
+
+mkdir -p ./dist
+RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
+RAPIDS_PY_WHEEL_NAME="${PROJECT_NAME}_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 ./dist
+
+# echo to expand wildcard before adding `[extra]` requires for pip
+python -m pip install $(echo ./dist/${PROJECT_NAME}*.whl)[test]
+
+# Run smoke tests for aarch64 pull requests
```
bdice, line 18: Same question about smoke tests. Implement them, or remove the commented section.

ci/test_wheel_distributed_ucxx.sh
```diff
@@ -19,6 +19,12 @@ python -m pip install ./local-ucxx-dep/ucxx*.whl
 # echo to expand wildcard before adding `[extra]` requires for pip
 python -m pip install $(echo ./dist/${PROJECT_NAME}*.whl)[test]
 
+# TODO: We need distributed installed in developer mode to provide test utils,
+# we still need to match to the `rapids-dask-dependency` version.
+rapids-logger "Install Distributed in developer mode"
+git clone https://github.com/dask/distributed /tmp/distributed
+python -m pip install -e /tmp/distributed
```
pentschev, line 26: @vyasr we don't need to address this immediately, but do you think it would work if we had optional dependencies for tests in https://github.com/rapidsai/rapids-dask-dependency/blob/ac821e6a3e396340f65fe79dc834f5b711d3b0cb/pip/rapids-dask-dependency/pyproject.toml#L14-L17 , where we essentially `pip install -e` the packages? I don't know if that's possible -- and I think it isn't -- but would like to know if you have any ideas how this could be done. We would also need something similar for conda packages, I currently [`pip install -e` in conda CI as well](https://github.com/rapidsai/ucxx/blob/94a19fde0e718ab3af18e1afc5ba79988dee9fdd/ci/test_python.sh#L99-L100) which isn't a good long-term solution.

bdice (commented):
Just a few comments to address, which need some input from others.

ci/test_wheel_distributed_ucxx.sh
bdice: Let's discuss here instead: https://github.com/rapidsai/ucxx/pull/167#discussion_r1470179504

ci/test_wheel_distributed_ucxx.sh
bdice: I had [a question](https://github.com/rapidsai/ucxx/pull/167#discussion_r1470174337) about this as well. Let's discuss in this thread.

ci/test_wheel_distributed_ucxx.sh
pentschev: Just wrote https://github.com/rapidsai/ucxx/pull/167#discussion_r1470179504 about this. Yes, we need some testing dependencies from distributed, specifically https://github.com/rapidsai/ucxx/blob/94a19fde0e718ab3af18e1afc5ba79988dee9fdd/python/distributed-ucxx/distributed_ucxx/tests/test_ucxx.py#L173 . I don't know of a good solution for this problem either though and I very much don't want to copy-paste the code in here because that will eventually cause us to get out-of-sync with Distributed code and essentially mean we're not testing it properly anymore.

ci/test_wheel_distributed_ucxx.sh
pentschev: Done in https://github.com/rapidsai/ucxx/pull/167/commits/3ef814b2ca7340d9fa4b6fda3f4624ed2d7cd323 .

ci/test_wheel_ucxx.sh
pentschev: Also done in https://github.com/rapidsai/ucxx/pull/167/commits/3ef814b2ca7340d9fa4b6fda3f4624ed2d7cd323 .

ci/test_wheel_distributed_ucxx.sh
pentschev: Also adding here what I wrote in the other thread:
> Yes, we need some testing dependencies from distributed, specifically https://github.com/rapidsai/ucxx/blob/94a19fde0e718ab3af18e1afc5ba79988dee9fdd/python/distributed-ucxx/distributed_ucxx/tests/test_ucxx.py#L173
> I don't know of a good solution for this problem either though and I very much don't want to copy-paste the code in here because that will eventually cause us to get out-of-sync with Distributed code and essentially mean we're not testing it properly anymore.

ci/test_wheel_distributed_ucxx.sh
```diff
@@ -0,0 +1,39 @@
...
+python -m pip install $(echo ./dist/${PROJECT_NAME}*.whl)[test]
+
+# TODO: We need distributed installed in developer mode to provide test utils,
+# we still need to match to the `rapids-dask-dependency` version.
+rapids-logger "Install Distributed in developer mode"
+git clone https://github.com/dask/distributed /tmp/distributed
+python -m pip install -e /tmp/distributed
+
+print_ucx_config
+
```
vyasr, line 29:
```suggestion
```

ci/test_wheel_ucxx.sh
```diff
@@ -0,0 +1,31 @@
...
+
+mkdir -p ./dist
+RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
+RAPIDS_PY_WHEEL_NAME="${PROJECT_NAME}_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 ./dist
+
+# echo to expand wildcard before adding `[extra]` requires for pip
+python -m pip install $(echo ./dist/${PROJECT_NAME}*.whl)[test]
+
+print_ucx_config
+
```
vyasr, line 19:
```suggestion
```

ci/test_wheel_distributed_ucxx.sh
```diff
@@ -25,8 +25,6 @@ rapids-logger "Install Distributed in developer mode"
 git clone https://github.com/dask/distributed /tmp/distributed
 python -m pip install -e /tmp/distributed
 
-print_ucx_config
```
pentschev: @vyasr  looks ok for wheels, but can you remind why `ucx_info` isn't available in the wheel package ? This is very often required for any debugging, so I think we should always include it.

ci/test_wheel_distributed_ucxx.sh
pentschev: You're right that they intentionally do not ship because they are only used in that particular test file. I very much wanted to avoid having to modify things there, the file is reasonably large with many functions that would need to move and we would have to bear the burden of properly documenting/annotating those functions that I'm not even familiar with internals, and eventually we would also need to bear the burden of some maintenance, which we currently don't have the bandwidth to. If we can't find a reasonably simple solution to install Distributed in dev mode, I think I'd prefer to limit or skip that test for now.

msarahan (commented):
I'm amazed and horrified by the library manipulation that has to be done, but needs must. I'm glad that the three of you are so knowledgeable. I can't approve this, but I don't have anything left to add that would improve this.

vyasr:
> I'm amazed and horrified by the library manipulation that has to be done, but needs must. I'm glad that the three of you are so knowledgeable. I can't approve this, but I don't have anything left to add that would improve this.

Yeah it's pretty horrific. At least this time around I already knew what was needed, the first time was an adventure to figure out why cuml dask tests were seg faulting :sweat_smile: Especially since we don't have MG CI for wheel tests so that was a bit of a last-minute discovery for wheels.

ci/test_wheel_distributed_ucxx.sh
vyasr: There isn't an easy way to bundle arbitrary executables into the wheel. It can be done, but it'll require more of the type of complexity that we're already doing for bundling ucx libraries, along with some extra bits to wrap that in an installable Python entry point. I don't think we want to gate this PR on figuring that out.

pentschev: We definitely don't want that in this PR, but this may be something we should consider looking into sometime in the near future. Any serious-ish use of UCX will eventually need that information for some debugging.

ci/build_wheel_distributed_ucxx.sh
```diff
@@ -0,0 +1,12 @@
...
+
+set -euo pipefail
+
+package_dir="python/distributed-ucxx"
+
+export RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
+
+./ci/build_wheel.sh distributed-ucxx ${package_dir}
+
+RAPIDS_PY_WHEEL_NAME="distributed_ucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-upload-wheels-to-s3 ${package_dir}/dist
```
vyasr, line 12: We can move both line 8 and line 12 into build_wheel.sh (and remove the export), then use `package_name` when setting `RAPIDS_PY_WHEEL_NAME`. The only catch will be replacing the hyphen in the name with an underscore, which we have an example of in line 24 of build_wheel.sh (the line setting `__git_commit__`)

ci/build_wheel_ucxx.sh
```diff
@@ -0,0 +1,12 @@
...
+
+set -euo pipefail
+
+package_dir="python"
+
+export RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
+
+./ci/build_wheel.sh ucxx ${package_dir}
+
+RAPIDS_PY_WHEEL_NAME="ucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-upload-wheels-to-s3 ${package_dir}/final_dist
```
vyasr, line 12: Same comment as in build_wheel_distributed_ucxx.sh

ci/test_common.sh
```diff
@@ -0,0 +1,155 @@
+#!/bin/bash
+
+# SPDX-FileCopyrightText: Copyright (c) 2022-2024, NVIDIA CORPORATION & AFFILIATES.
+# SPDX-License-Identifier: BSD-3-Clause
+
+set -euo pipefail
+
+source "$(dirname "$0")/test_utils.sh"
```
vyasr, line 8: Can we just pull test_utils.sh into this file, or just put all this new functionality directly into test_utils.sh? AFAICT there's no reason for them to be separate and we always need both.

ci/test_utils.sh
```diff
@@ -20,5 +20,8 @@ print_system_stats() {
 
 print_ucx_config() {
   rapids-logger "UCX Version and Build Configuration"
+
+  set +e
   ucx_info -v
+  set -e
```
vyasr, line 26: Are these change relevant? Moreover, are they still necessary now that the wheel test scripts are no longer calling `print_ucx_config`?

ci/test_wheel_distributed_ucxx.sh
vyasr: I'd rather limit or skip. This logic:
> I very much don't want to copy-paste the code in here because that will eventually cause us to get out-of-sync with Distributed code and essentially mean we're not testing it properly anymore.

feels directly contradictory to 
> we would have to bear the burden of properly documenting/annotating those functions that I'm not even familiar with internals, and eventually we would also need to bear the burden of some maintenance, which we currently don't have the bandwidth to

If we don't want to bear the maintenance burden then I'd much rather not try and take on some partial maintenance burden via a hacky installation of distributed here that forces us to stay partially in sync.

ci/wheel_smoke_test_distributed_ucxx.py
```diff
@@ -0,0 +1,65 @@
+import asyncio
+
+import pytest
+
+from distributed.comm import connect, listen
+from distributed.protocol import to_serialize
+
+import ucxx
+
+import distributed_ucxx  # noqa: E402
```
vyasr, line 10:
```suggestion
import distributed_ucxx
```

ci/wheel_smoke_test_distributed_ucxx.py
```diff
@@ -0,0 +1,65 @@
...
+@pytest.mark.parametrize(
+    "g",
+    [
+        lambda cudf: cudf.Series([1, 2, 3]),
+        lambda cudf: cudf.DataFrame({"a": [1, 2, None], "b": [1.0, 2.0, None]}),
+    ],
+)
+@gen_test()
+async def test_ping_pong_cudf(ucxx_loop, g):
+    # if this test appears after cupy an import error arises
```
vyasr, line 46: I assume typo?
```suggestion
    # if this test appears after cudf an import error arises
```

vyasr: Hmm but this is also potentially a bit concerning. I see Python 3.7 in your traceback. Have we tested this on a few different machines with different Python versions? Could we remove this check and try running in CI? I'm wondering if this problem arises due to compatibility issues with the CXX ABI. It's probably not actionable, but it's worth double-checking in case it is.

ci/wheel_smoke_test_distributed_ucxx.py
vyasr, line 1: This file is missing a copyright.

ci/wheel_smoke_test_ucxx.py
```diff
@@ -0,0 +1,116 @@
+# SPDX-FileCopyrightText: Copyright (c) 2022-2023, NVIDIA CORPORATION & AFFILIATES.
```
vyasr, line 1:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2022-2024, NVIDIA CORPORATION & AFFILIATES.
```

cpp/CMakeLists.txt
```diff
@@ -171,17 +175,15 @@ target_compile_definitions(
 
 # Enable RMM if necessary
 if(UCXX_ENABLE_RMM)
-    target_compile_definitions(ucxx PUBLIC UCXX_ENABLE_RMM)
+    target_link_libraries(ucxx PUBLIC rmm::rmm)
+    target_compile_definitions(ucxx PUBLIC UCXX_ENABLE_RMM "SPDLOG_ACTIVE_LEVEL=SPDLOG_LEVEL_${RMM_LOGGING_LEVEL}")
 endif()
 
 # Define spdlog level
```
vyasr, line 182: Whoops this was my fault. Please move this comment to line 179.

vyasr (changes requested):
This is very close now, we should be able to merge soon.

ci/build_wheel.sh
```diff
@@ -0,0 +1,143 @@
...
+    alpha_spec=',>=0.0.0a0'
+fi
+
+if [[ ${package_name} == "distributed-ucxx" ]]; then
+    sed -r -i "s/rapids-dask-dependency==(.*)\"/rapids-dask-dependency==\1${alpha_spec}\"/g" ${pyproject_file}
+    sed -r -i "s/\"ucxx(.*)\"/\"ucxx${PACKAGE_CUDA_SUFFIX}\1${alpha_spec}\"/g" ${pyproject_file}
+
+    cd "${package_dir}"
+
+    python -m pip wheel . -w dist -vvv --no-deps --disable-pip-version-check
```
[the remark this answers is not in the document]
pentschev: I think this is wrong, in line 39 above we `cd "${package_dir}"` and now we attempt to build `"${package_dir}"` instead of `.`, am I missing something?

ci/test_wheel_distributed_ucxx.sh
pentschev: I don't see where the contradiction as the statements are talking about two different aspects:
1. Copy-pasting those function _in distributed-ucxx_ means they will eventually become out-of-sync with the actual implementation from Distributed, which is the source-of-truth;
2. The maintenance burden I'm referring to is the maintenance burden of those functions _in Distributed_ as we will now have "some authorship" because we're requesting they live as part of distributed packages and will have to document and properly type-annotate them, etc.

IOW, we still want to use them here _for testing purposes only_ but I'd very much avoid touching much of Distributed code if we don't absolutely need to.

ci/wheel_smoke_test_distributed_ucxx.py
pentschev: This is just a copy of https://github.com/rapidsai/ucxx/blob/e492dca5b3fe1b0da095c12327b1e1f9c92f262d/python/distributed-ucxx/distributed_ucxx/tests/test_ucxx.py#L178-L215 with fewer parameters, the comment is not relevant here as this test is running alone here so I've removed it inhttps://github.com/rapidsai/ucxx/pull/167/commits/2107f160a6c7530e39dfdcc7d03b5bb124791f5c .

ci/test_common.sh
pentschev: Done in https://github.com/rapidsai/ucxx/pull/167/commits/9e6ce880d5101076be031161f1fbeed3a4337e33

ci/wheel_smoke_test_distributed_ucxx.py
pentschev: Fixed in https://github.com/rapidsai/ucxx/pull/167/commits/3d43a5caef83f5c345bb62910759268c500b38a3

ci/wheel_smoke_test_distributed_ucxx.py
pentschev: Removed in https://github.com/rapidsai/ucxx/pull/167/commits/d6af7c692fa20db8706fbe26c8ad7b647bd3ddbc

ci/wheel_smoke_test_ucxx.py
pentschev: Fixed in https://github.com/rapidsai/ucxx/pull/167/commits/3d43a5caef83f5c345bb62910759268c500b38a3

cpp/CMakeLists.txt
pentschev: Done in https://github.com/rapidsai/ucxx/pull/167/commits/1b3977f3f7b73a75271c3c540113d142548bc985

pentschev (approved):
I think I've addressed all your comments @vyasr , please have another look.

ci/build_wheel.sh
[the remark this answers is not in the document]
vyasr: Previously we were doing `cd $package_dir; build .`, and I'm suggesting that we do `build $package_dir`. In both cases we build `$package_dir`. Does that make sense? Perhaps you missed that I deleted the cd?

ci/build_wheel_distributed_ucxx.sh
vyasr: Still to-do

ci/build_wheel_ucxx.sh
vyasr: Still to-do

ci/test_wheel_distributed_ucxx.sh
vyasr: Right, but right now we implicitly incur part of the maintenance burden of 2 because if dask decides to change the behavior of that function in some way we'll see it in our test. We're not maintaining the function, but we are effectively signing up to keep track of an implementation detail of dask's tests. Conversely, if it was public there would be a higher maintenance burden (because we'd have to maintain docs/annotations/etc as you say) but we would also be less likely to be broken without warning (of course, given dask's low stability it could still happen, but it reduces the odds). The current approach also makes it harder for devs to understand why the tests fail if they don't know that they need to download the source of distributed to access this functionality.

In any case, I don't think there is a good solution for installing the editable version from source. Every solution I can think of is equally hacky, so for now I think we can leave this as is.

vyasr (changes requested):
A couple of minor leftover items then we should be good to go.

ci/build_wheel.sh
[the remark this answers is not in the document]
pentschev: Oops, I definitely missed that, sorry about it. I've applied the change now.

ci/build_wheel_distributed_ucxx.sh
pentschev: Grr, I hate that GH has discrepancies on what's shown in the "Files changed" tab (suggested changes) and general comments that do not suggest a change (like this one), I always end up missing stuff.

Now done in https://github.com/rapidsai/ucxx/pull/167/commits/2a8495df31ad1c563a7eba0967f39783006e7cc6

ci/build_wheel_ucxx.sh
pentschev: Also done in https://github.com/rapidsai/ucxx/pull/167/commits/2a8495df31ad1c563a7eba0967f39783006e7cc6

ci/test_utils.sh
pentschev: You're right, we don't need it anymore. Removed in https://github.com/rapidsai/ucxx/pull/167/commits/035c45352f9dd23c717bc5c66811be8652b2a591

ci/test_wheel_distributed_ucxx.sh
vyasr: I think the split you proposed for getting extra coverage in CI. Let's make that change in a follow-up PR where you apply it to both conda and pip testing.

vyasr (approved):

bdice (approved):
LGTM! Thanks for all the teamwork on this!

PR #169, opened 2024-01-18
Fix pip dependencies

trxcllnt:
* Move conda-only dependencies out of `pyproject` and `requirements` sections in `dependencies.yaml`
* Add `rmm`, `cudf`, and `cupy` matrices


dependencies.yaml
```diff
@@ -180,11 +198,55 @@ dependencies:
...
+              - cupy-cuda12x -f https://pip.cupy.dev/aarch64 # TODO: Verify that this works.
+          # All CUDA 11 + x86_64 versions
+          - matrix: {cuda: "11.*", arch: x86_64}
+            packages:
+              - cupy-cuda11x>=12.0.0
+          # All CUDA 11 + aarch64 versions
+          - matrix: {cuda: "11.*", arch: aarch64}
+            packages:
+              - cupy-cuda11x -f https://pip.cupy.dev/aarch64 # TODO: Verify that this works.
+          - {matrix: null, packages: [cupy-cuda11x>=12.0.0]}
```
bdice, line 231: cupy has ARM wheels on PyPI now and no special index is needed. We can simplify this.

FYI, I've already applied this change in several of the PRs testing CUDA 12.2, at the same time that I updated and simplified the matrices for other RAPIDS packages.

```suggestion
      - output_types: [requirements, pyproject]
        matrices:
          - matrix: {cuda: "12.*"}
            packages:
              - cupy-cuda12x>=12.0.0
          - matrix: {cuda: "11.*"}
            packages:
              - cupy-cuda11x>=12.0.0
          - {matrix: null, packages: [cupy-cuda11x>=12.0.0]}
```

python/pyproject.toml
```diff
@@ -41,7 +41,7 @@ classifiers = [
 test = [
     "cloudpickle",
     "cudf==24.2.*",
-    "cupy",
+    "cupy-cuda11x>=12.0.0",
```
bdice, line 44: We're inconsistent in how we handle cudf/cupy fallback. This is a problem across basically all of RAPIDS, so it'd be good for us to unify our approaches.

If we need `cupy-cuda11x` here, we would probably want `cudf-cu11` as well. For all RAPIDS packages we've been falling back to the "conda name" but we've been falling back to the "CUDA 11 pip name" for cupy. We should do one of:
- use `cudf` and `cupy`
- use `cudf-cu11` and `cupy-cuda11x`

trxcllnt: I'm using the "conda name" here because it works automatically when building the repos from source.

bdice: Gotcha. I wondered about that. We can leave this as-is, then.

bdice (approved):

pentschev (approved):

raydouglass (approved):

PR #161, opened 2024-01-11
Support CUDA 12.2

jameslamb:
## Description

* switches to CUDA 12.2.2 for building conda packages and wheels
* adds new tests running against CUDA 12.2.2

### Notes for Reviewers

This is part of ongoing work to build and test packages against CUDA 12.2.2 across all of RAPIDS.

For more details see:

* https://github.com/rapidsai/build-planning/issues/7
* https://github.com/rapidsai/shared-workflows/pull/166

Planning a second round of PRs to revert these references back to a proper `branch-24.{nn}` release branch of `shared-workflows` once https://github.com/rapidsai/shared-workflows/pull/166 is merged.

*(created with `rapids-reviser`)*

PR #171, opened 2024-01-19
Build wheels for distributed-ucxx

bdice:
This PR follows #167 and adds wheel builds for `distributed-ucxx`.

Note: This should be rebased following #167's merge.

PR #164, opened 2024-01-12
Add Doxygen build and fix all warnings

pentschev:
The C++ code was documented for some time but Doxygen build process was not included. This change now introduces Doxygen builds and fixes all documentation warnings.

PR #174, opened 2024-01-22
handle more RAPIDS version formats in update-version.sh, refactor dependencies.yaml

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/13.

Updates `update-version.sh` to correctly handle RAPIDS dependencies like `cudf-cu12==24.2.*`.

This also pulls in some dependency refactoring originally added in #161, which allows greater use of dependencies.yaml globs (and therefore less maintenance effort to support new CUDA versions).

### How I tested this

The portability of this updated `sed` command was tested here: https://github.com/rapidsai/cudf/pull/14825#issuecomment-1904735849.

In this repo, I ran the following:

```shell
./ci/release/update-version.sh '0.36.00'
git diff

./ci/release/update-version.sh '0.37.00
git diff
```

Confirmed that that first `git diff` changed all the things I expected, and that second one showed 0 changes.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -12,10 +12,13 @@ files:
       - cuda
       - cuda_version
       - dev
+      - docs
       - py_version
       - run_python
       - test_cpp
       - test_python
+      - depends_on_cupy
+      - depends_on_cudf
   test_cpp:
     output: none
     includes:
@@ -27,10 +30,17 @@ files:
       - cuda_version
       - py_version
       - test_python
+      - depends_on_cupy
+      - depends_on_cudf
   checks:
     output: none
     includes:
       - checks
+      - py_version
+  docs:
+    output: none
+    includes:
+      - docs
       - py_version
   py_build:
     output: pyproject
@@ -54,6 +64,8 @@ files:
       key: test
     includes:
       - test_python
+      - depends_on_cupy
+      - depends_on_cudf
 channels:
   - rapidsai
   - rapidsai-nightly
@@ -76,18 +88,30 @@ dependencies:
           - spdlog>=1.12.0,<1.13
   build_python:
     common:
+      - output_types: [conda]
+        packages:
+          - &rmm_conda rmm==24.4.*
       - output_types: [conda, requirements, pyproject]
         packages:
           - *cmake_ver
           - cython>=3.0.0
           - ninja
-          - rmm==24.4.*
       - output_types: conda
         packages:
           - scikit-build-core>=0.7.0
       - output_types: [requirements, pyproject]
         packages:
           - scikit-build-core[pyproject]>=0.7.0
+    specific:
+      - output_types: [requirements, pyproject]
+        matrices:
+          - matrix: {cuda: "12.*"}
+            packages:
+              - rmm-cu12==24.4.*
+          - matrix: {cuda: "11.*"}
+            packages:
+              - rmm-cu11==24.4.*
+          - {matrix: null, packages: [*rmm_conda]}
   checks:
     common:
       - output_types: [conda, requirements]
@@ -144,6 +168,13 @@ dependencies:
           - autoconf
           # UCXX Build
           - pkg-config
+          # Docs Build
+          - &doxygen doxygen=1.10.0 # pre-commit hook needs a specific version.
+  docs:
+    common:
+      - output_types: [conda]
+        packages:
+          - *doxygen
   py_version:
     specific:
       - output_types: conda
@@ -180,11 +211,45 @@ dependencies:
       - output_types: [conda, requirements, pyproject]
         packages:
           - cloudpickle
-          - cudf==24.4.*
-          - cupy
           - dask
           - distributed
           - numba>=0.57.1
           - pytest
           - pytest-asyncio
           - pytest-rerunfailures
+  depends_on_cupy:
+    common:
+      - output_types: conda
+        packages:
+          - cupy>=12.0.0
+    specific:
+      - output_types: [requirements, pyproject]
+        matrices:
+          - matrix: {cuda: "12.*"}
+            packages:
+              - cupy-cuda12x>=12.0.0
+          - matrix: {cuda: "11.*"}
+            packages:
+              - cupy-cuda11x>=12.0.0
+          - {matrix: null, packages: [cupy-cuda11x>=12.0.0]}
+  depends_on_cudf:
+    common:
+      - output_types: conda
+        packages:
+          - &cudf_conda cudf==24.4.*
+      - output_types: requirements
+        packages:
+          # pip recognizes the index as a global option for the requirements.txt file
+          # This index is needed for rmm, cubinlinker, ptxcompiler.
+          - --extra-index-url=https://pypi.nvidia.com
+          - --extra-index-url=https://pypi.anaconda.org/rapidsai-wheels-nightly/simple
+    specific:
+      - output_types: [requirements, pyproject]
+        matrices:
+          - matrix: {cuda: "12.*"}
+            packages:
+              - cudf-cu12==24.4.*
+          - matrix: {cuda: "11.*"}
+            packages:
+              - cudf-cu11==24.4.*
+          - {matrix: null, packages: [*cudf_conda]}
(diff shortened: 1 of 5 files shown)


dependencies.yaml
```diff
@@ -76,18 +88,30 @@ dependencies:
           - spdlog>=1.12.0,<1.13
   build_python:
     common:
+      - output_types: [conda]
+        packages:
+          - &rmm_conda rmm==24.4.*
```
bdice, line 93: This should be in a list `depends_on_rmm`. I refactored this and a few other things in https://github.com/rapidsai/ucxx/pull/171/files#diff-5475a6d76de4c506ee92cf6f941bba30a6a07b5881cfea531d42a6ec035095a6 but I'd like this PR to go in first, and reduce the diff of #171. Maybe just compare the two dependencies files and port over the changes you like into this PR?

bdice: I backported this in b0142c2.

bdice:
I included some changes that we know are needed for #167 / #171. The changes in b0142c2 will reduce merge conflicts as we continue with multiple work streams (adding CUDA 12.2, fixing branch updates, adding wheels).

jameslamb:
> I included some changes that we know are needed for https://github.com/rapidsai/ucxx/pull/167 / https://github.com/rapidsai/ucxx/pull/171

These all look good to me, thanks @bdice

dependencies.yaml
```diff
@@ -29,43 +33,64 @@ files:
...
     includes:
       - checks
       - py_version
-  docs:
-    output: none
-    includes:
-      - docs
-      - py_version
-  py_build:
+  py_build_ucxx:
```
bdice, line 45: Accidentally removed docs key.
```suggestion
  docs:
    output: none
    includes:
      - docs
      - py_version
  py_build_ucxx:
```

AyodeAwe (approved):

jakirkham:
It looks like some conflicts crept in here. Maybe these files need to be regenerated with [`rapids-dependency-file-generator`]( https://github.com/rapidsai/dependency-file-generator )

jakirkham:
Have resolved conflicts with PR ( https://github.com/rapidsai/ucxx/pull/164 )

jakirkham (approved):

jakirkham (changes requested):
As we are moving to `rapids-dask-dependency` in this PR, think we should update the Conda recipes as well. In particular am noticing these lines here

https://github.com/rapidsai/ucxx/blob/2249bed6ca09550abb92a6ea9496b0a024697fdd/conda/recipes/ucxx/meta.yaml#L274-L275

jameslamb:
> As we are moving to rapids-dask-dependency in this PR, think we should update the Conda recipes as well. In particular am noticing these lines here

@jakirkham I just did that in https://github.com/rapidsai/ucxx/pull/174/commits/20a86a403469486e9dcf0e781431d52447d2a96b

I don't see any other uses of `dask` or `distributed` that need to be replaced.

jameslamb:
whoops that needs to be a RAPIDS version not UCX's version

```text
The reported errors are:
- Encountered problems while solving:
-   - nothing provides rapids-dask-dependency 0.37.* needed by distributed-ucxx-0.37.00a-py3.10_240125_g20a86a4_18
```

([build link](https://github.com/rapidsai/ucxx/actions/runs/7656935655/job/20866474251?pr=174))

Will fix that momentarily.

conda/recipes/ucxx/meta.yaml
```diff
@@ -271,8 +271,7 @@ outputs:
         - tomli
       run:
         - python * *_cpython
-        - dask >=2023.9.2
-        - distributed >=2023.9.2
+        - rapids-dask-dependency ={{ minor_version }}
```
jakirkham, line 274: UCXX is versioned differently than the rest of RAPIDS. So think we want something like this. Should also fix the error on CI

```suggestion
        - rapids-dask-dependency =24.4
```

jameslamb: yeah my mistake, thanks for that. I think that's the right fix based on the `ci/release/update-version.sh` change in this PR.

jakirkham (approved):
Thanks James! 🙏

bdice:
I noticed a small issue, `rapids-dask-dependency` was pinned as `=24.4` while `update-version.sh` would pin it as `=24.04`. Then I re-ran `ci/release/update-version.sh 0.37.00` and realized that there were a bunch of problems from the docs PR #164 not aligning with `update-version.sh`, and PRs targeting `branch-0.36` that were not updated for `branch-0.37`. I pushed e06b20c to fix these issues. Sorry for the late change - I know we want to get this merged soon, but this felt sufficiently related to fix in this PR.

jameslamb:
> Sorry for the late change - I know we want to get this merged soon, but this felt sufficiently related to fix in this PR.

Totally fine, thanks for that! I hadn't noticed it.

jakirkham:
Thanks Bradley! 🙏

Had missed that

cpp/doxygen/Doxyfile
```diff
@@ -38,7 +38,7 @@ PROJECT_NAME           = libucxx
 # could be handy for archiving the generated documentation or if some version
 # control system is used.
 
-PROJECT_NUMBER         = 0.36.00
+PROJECT_NUMBER         = 0.37.00
```
jakirkham, line 41: For context, this this is the line that needs C++ review

python/pyproject.toml
```diff
@@ -41,9 +42,7 @@ classifiers = [
 test = [
     "cloudpickle",
     "cudf==24.4.*",
-    "cupy>=12.0.0",
-    "dask",
-    "distributed",
```
jameslamb, line 46: I think maybe something was missed here? I'd expect to see `dask` and `distributed` drop out and be replaced by `rapids-dask-dependency` in this block.

jakirkham (approved):

python/pyproject.toml
bdice: It turns out that the dask bits weren't needed for pure ucxx tests (only distributed-ucxx). The pinning appears in the dependencies of distributed-ucxx. I think we're good here.

dependencies.yaml
```diff
@@ -190,48 +213,78 @@ dependencies:
...
       - output_types: [conda, requirements, pyproject]
         packages:
           - cloudpickle
-          - dask
-          - distributed
-          - numba>=0.57.1
+          - *numba
           - pytest
           - pytest-asyncio
           - pytest-rerunfailures
```
jakirkham, line 246:
```suggestion
          - pytest-rerunfailures
          - rapids-dask-dependency==24.4.*
```

python/pyproject.toml
jameslamb: Or does `ucxx` genuinely not need Dask for its tests?

Sorry not that familiar with this library 😬

jakirkham: Hmm...ok. If we do want this here, added a suggestion below. Though no strong feelings either way. Happy to go with whatever was agreed to before

https://github.com/rapidsai/ucxx/pull/174#discussion_r1466735120

dependencies.yaml
bdice: I think this is not needed - see above. https://github.com/rapidsai/ucxx/pull/174#discussion_r1466735058

python/pyproject.toml
jameslamb: > I believe that the dask bits aren't needed for pure ucxx tests

Ok got it, thanks.

jakirkham: FWIW do see lines like this

So think we may want to include this in the end

https://github.com/rapidsai/ucxx/blob/2249bed6ca09550abb92a6ea9496b0a024697fdd/python/ucxx/utils.py#L29-L30

https://github.com/rapidsai/ucxx/blob/2249bed6ca09550abb92a6ea9496b0a024697fdd/python/ucxx/_lib_async/tests/test_send_recv_two_workers.py#L25

bdice: Okay, this might have been a misunderstanding on my part. Let's merge the suggestion below and keep it as a test dependency.

jakirkham (approved):

jakirkham:
Thanks all! 🙏

PR #183, opened 2024-01-31
Run tests requiring Distributed internals separately

pentschev:
Some tests require Distributed internals and installed in developer mode. Since we have no good way to install it with `rapids-dask-dependency`, we now run those separately from the rest.

### DIFF
--- a/ci/test_common.sh
+++ b/ci/test_common.sh
@@ -167,9 +167,19 @@ run_distributed_ucxx_tests() {
 
   CMD_LINE="UCXPY_PROGRESS_MODE=${PROGRESS_MODE} UCXPY_ENABLE_DELAYED_SUBMISSION=${ENABLE_DELAYED_SUBMISSION} UCXPY_ENABLE_PYTHON_FUTURE=${ENABLE_PYTHON_FUTURE} timeout 10m python -m pytest -vs python/distributed-ucxx/distributed_ucxx/tests/"
 
-  # Workaround for https://github.com/rapidsai/ucxx/issues/15
-  # CMD_LINE="UCX_KEEPALIVE_INTERVAL=1ms ${CMD_LINE}"
-
   log_command "${CMD_LINE}"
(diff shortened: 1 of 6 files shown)


vyasr (approved):
Thanks!

raydouglass (approved):

ci/test_python.sh
charlesbluca, line 1: I see there's still a Distributed source install in this file preceding the standard test runs - is this intentional, or should that be removed?

pentschev: No, that was my mistake, good catch! Fixed in https://github.com/rapidsai/ucxx/pull/183/commits/78e01dd76ea134f6c0a8be88282c4cb3e8e04e8c

charlesbluca (approved):
Thanks @pentschev!

pentschev:
Thanks for reviews @charlesbluca @raydouglass @vyasr !

ci/test_common.sh
```diff
@@ -160,16 +160,40 @@ run_py_benchmark() {
...
+  # developer mode. This isn't a great solution but it's what we can currently do
+  # to run non-public API tests in CI.
+
+  rapids-logger "Install Distributed in developer mode"
+  git clone https://github.com/dask/distributed /tmp/distributed
+  pip install -e /tmp/distributed
+  # `pip install -e` removes files under `distributed` but not the directory, later
+  # causing failures to import modules.
+  PYTHON_ENV_PATH=${CONDA_PREFIX:-/pyenv}
+  rm -rf $(find ${PYTHON_ENV_PATH} -type d -iname "site-packages")/distributed
```
pentschev, line 174: I had to do this as explained in the comment above. I think this only happens if you happen to run something that `import distributed` first and creates pycache and stuff. I've always had this problem when developing locally and I do not have a better solution. Just pinging here in case you know of a cleaner way @vyasr .

pentschev: Sorry, I had issued a /merge here already. If you have any ideas I'm happy to address it in a follow-up PR.

PR #190, opened 2024-02-09
Add `ucxx::MemoryHandle` and `ucxx::RemoteKey` C++ classes

pentschev:
Add `ucxx::MemoryHandle` and `ucxx::RemoteKey` C++ classes, wrapping UCP memory handles (`ucp_mem_h`) and remote keys (`ucp_rkey_h`), respectively.

These new classes are required to provide basic support for RMA (Remote Memory Access) operations being implemented in #166 .

### DIFF
--- a/cpp/include/ucxx/context.h
+++ b/cpp/include/ucxx/context.h
@@ -188,6 +188,13 @@ class Context : public Component {
    * `ucxx::MemoryHandle` and will not be destroyed until all `ucxx::MemoryHandle`
    * objects are destroyed first.
    *
+   * The allocation requires a `size` and a `buffer`. The actual size of the allocation may
+   * be larger than requested, and can later be found calling the `getSize()` method. The
+   * `buffer` provided may be either a `nullptr`, in which case UCP will allocate a new
+   * memory region for it, or an already existing allocation, in which case UCP will only
+   * map it for RMA and it's the caller's responsibility to keep `buffer` alive until this
+   * object is destroyed.
+   *
    * @code{.cpp}
    * // `context` is `std::shared_ptr<ucxx::Context>`
    * // Allocate a 128-byte buffer with UCP.
--- a/cpp/include/ucxx/context.h
+++ b/cpp/include/ucxx/context.h
@@ -17,6 +17,7 @@
 
 namespace ucxx {
 
+class MemoryHandle;
 class Worker;
 
 /**
@@ -178,6 +179,37 @@ class Context : public Component {
    */
   std::shared_ptr<Worker> createWorker(const bool enableDelayedSubmission = false,
                                        const bool enableFuture            = false);
+
+  /**
+   * @brief Create a new `std::shared_ptr<ucxx::memoryHandle>`.
+   *
+   * Create a new `std::shared_ptr<ucxx::MemoryHandle>` as a child of the current
+   * `ucxx::Context`.  The `ucxx::Context` will retain ownership of the underlying
+   * `ucxx::MemoryHandle` and will not be destroyed until all `ucxx::MemoryHandle`
+   * objects are destroyed first.
+   *
+   * @code{.cpp}
+   * // `context` is `std::shared_ptr<ucxx::Context>`
+   * // Allocate a 128-byte buffer with UCP.
+   * auto memoryHandle = context->createMemoryHandle(128, nullptr);
+   *
+   * // Map an existing 128-byte buffer with UCP.
+   * size_t allocationSize = 128;
+   * auto buffer = new uint8_t[allocationSize];
+   * auto memoryHandleFromBuffer = context->createMemoryHandle(
+   *    allocationSize * sizeof(*buffer), reinterpret_cast<void*>(buffer)
+   * );
+   * @endcode
+   *
+   * @throws ucxx::Error if either `ucp_mem_map` or `ucp_mem_query` fail.
+   *
+   * @param[in] size    the minimum size of the memory allocation
+   * @param[in] buffer  the pointer to an existing allocation or `nullptr` to allocate a
+   *                    new memory region.
+   *
+   * @returns The `shared_ptr<ucxx::MemoryHandle>` object
+   */
+  std::shared_ptr<MemoryHandle> createMemoryHandle(const size_t size, void* buffer);
 };
 
 }  // namespace ucxx
--- a/cpp/include/ucxx/memory_handle.h
+++ b/cpp/include/ucxx/memory_handle.h
@@ -113,7 +113,7 @@ class MemoryHandle : public Component {
    *
    * Lifetime of the `ucp_mem_h` handle is managed by the `ucxx::MemoryHandle` object and
    * its ownership is non-transferrable. Once the `ucxx::MemoryHandle` is destroyed the
-   * memory is unampped and the handle is not valid anymore, it is the user's responsibility
+   * memory is unmapped and the handle is not valid anymore, it is the user's responsibility
    * to ensure the owner's lifetime while using the handle.
    *
    * @code{.cpp}
--- a/cpp/include/ucxx/memory_handle.h
+++ b/cpp/include/ucxx/memory_handle.h
@@ -67,7 +67,8 @@ class MemoryHandle : public Component {
    * be larger than requested, and can later be found calling the `getSize()` method. The
    * `buffer` provided may be either a `nullptr`, in which case UCP will allocate a new
    * memory region for it, or an already existing allocation, in which case UCP will only
-   * map it for RMA.
+   * map it for RMA and it's the caller's responsibility to keep `buffer` alive until this
+   * object is destroyed.
    *
    * @code{.cpp}
    * // `context` is `std::shared_ptr<ucxx::Context>`
--- a/cpp/include/ucxx/memory_handle.h
+++ b/cpp/include/ucxx/memory_handle.h
@@ -63,12 +63,14 @@ class MemoryHandle : public Component {
    * The constructor for a `shared_ptr<ucxx::MemoryHandle>` object, mapping a memory buffer
    * with UCP to provide RMA (Remote Memory Access) to.
    *
-   * The allocation requires a `size` and a `buffer`. The actual size of the allocation may
-   * be larger than requested, and can later be found calling the `getSize()` method. The
-   * `buffer` provided may be either a `nullptr`, in which case UCP will allocate a new
-   * memory region for it, or an already existing allocation, in which case UCP will only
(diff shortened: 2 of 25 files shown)


cpp/include/ucxx/context.h
```diff
@@ -178,6 +179,37 @@ class Context : public Component {
...
+   * // `context` is `std::shared_ptr<ucxx::Context>`
+   * // Allocate a 128-byte buffer with UCP.
+   * auto memoryHandle = context->createMemoryHandle(128, nullptr);
+   *
+   * // Map an existing 128-byte buffer with UCP.
+   * size_t allocationSize = 128;
+   * auto buffer = new uint8_t[allocationSize];
+   * auto memoryHandleFromBuffer = context->createMemoryHandle(
+   *    allocationSize * sizeof(*buffer), reinterpret_cast<void*>(buffer)
+   * );
```
wence-, line 201: *suggestion*: In this case, the user is responsible for keeping the buffer live until the memory handle is disposed of. Perhaps add a note to that effect.

cpp/include/ucxx/memory_handle.h
```diff
@@ -0,0 +1,159 @@
...
+   * @brief Constructor for `shared_ptr<ucxx::MemoryHandle>`.
+   *
+   * The constructor for a `shared_ptr<ucxx::MemoryHandle>` object, mapping a memory buffer
+   * with UCP to provide RMA (Remote Memory Access) to.
+   *
+   * The allocation requires a `size` and a `buffer`. The actual size of the allocation may
+   * be larger than requested, and can later be found calling the `getSize()` method. The
+   * `buffer` provided may be either a `nullptr`, in which case UCP will allocate a new
+   * memory region for it, or an already existing allocation, in which case UCP will only
+   * map it for RMA.
```
wence-, line 70: *question*: If a buffer is provided, can it be the case that `getSize()` will return a value other than `size`?

cpp/include/ucxx/memory_handle.h
```diff
@@ -0,0 +1,159 @@
...
+   * //    context, allocationSize * sizeof(*buffer), reinterpret_cast<void*>(buffer)
+   * // );
+   * @endcode
+   *
+   * @throws ucxx::Error if either `ucp_mem_map` or `ucp_mem_query` fail.
+   *
+   * @param[in] context parent context where to map memory.
+   * @param[in] size    the minimum size of the memory allocation
+   * @param[in] buffer  the pointer to an existing allocation or `nullptr` to allocate a
+   *                    new memory region.
```
wence-, line 98: *suggestion*: Mention that if a buffer is provided, it is the user's responsibility to keep it alive until the memory handle is disposed of.

cpp/include/ucxx/memory_handle.h
```diff
@@ -0,0 +1,159 @@
...
+
+  ~MemoryHandle();
+
+  /**
+   * @brief Get the underlying `ucp_mem_h` handle.
+   *
+   * Lifetime of the `ucp_mem_h` handle is managed by the `ucxx::MemoryHandle` object and
+   * its ownership is non-transferrable. Once the `ucxx::MemoryHandle` is destroyed the
+   * memory is unampped and the handle is not valid anymore, it is the user's responsibility
+   * to ensure the owner's lifetime while using the handle.
```
wence-, line 114: *question*: owner of what?

cpp/include/ucxx/remote_key.h
```diff
@@ -0,0 +1,212 @@
...
+ * in the form of `ucp_rkey_h` (remote key, or rkey for short). This class encapsulates
+ * that object and provides methods to simplify its handling, both locally and remotely
+ * including (de-)serialization for transfers over the wire and reconstruction of the
+ * object on the remote process.
+ */
+class RemoteKey : public Component {
+ private:
+  ucp_rkey_h _remoteKey{nullptr};  ///< The unpacked remote key.
+  void* _packedRemoteKey{
+    nullptr};                      ///< The packed (local) remote key that may be unpacked (remote).
```
wence-, line 34: *question*: I think I can't quite parse this docstring. Is this the key that we generated locally, packed to send to the remote?

cpp/include/ucxx/remote_key.h
```diff
@@ -0,0 +1,212 @@
...
+   * possibly transferred over-the-wire and reconstruct the object to allow remote access.
+   *
+   * @code{.cpp}
+   * // remoteKey is `std::shared_ptr<ucxx::RemoteKey>`
+   * auto serializedRemoteKey = remoteKey->serialize();
+   * @endcode
+   *
+   * @throws std::runtime_error if checksum of the serialized object fails.
+   *
+   * @returns The serialized remote key.
```
wence-, line 89:
```suggestion
   * @returns The deserialized remote key.
```

cpp/include/ucxx/remote_key.h
```diff
@@ -0,0 +1,212 @@
...
+  friend std::shared_ptr<RemoteKey> createRemoteKeyFromSerialized(
+    std::shared_ptr<Endpoint> endpoint, SerializedRemoteKey serializedRemoteKey);
+
+  ~RemoteKey();
+
+  /**
+   * @brief Get the underlying `ucp_rkey_h` handle.
+   *
+   * Lifetime of the `ucp_rkey_h` handle is managed by the `ucxx::RemoteKey` object and
+   * its ownership is non-transferrable. Once the `ucxx::RemoteKey` is destroyed the address
```
wence-, line 152: *question*: Does this suggest the `RemoteKey` should not be a `Component` (and hence we shouldn't hand out `shared_ptr`s to it)? Or do I misunderstand?

cpp/src/memory_handle.cpp
```diff
@@ -0,0 +1,81 @@
...
+  if (buffer == nullptr) {
+    params = {.field_mask  = UCP_MEM_MAP_PARAM_FIELD_FLAGS | UCP_MEM_MAP_PARAM_FIELD_LENGTH,
+              .length      = size,
+              .flags       = UCP_MEM_MAP_NONBLOCK | UCP_MEM_MAP_ALLOCATE,
+              .memory_type = UCS_MEMORY_TYPE_HOST};
+  } else {
+    params = {.field_mask  = UCP_MEM_MAP_PARAM_FIELD_ADDRESS | UCP_MEM_MAP_PARAM_FIELD_LENGTH,
+              .address     = buffer,
+              .length      = size,
+              .memory_type = UCS_MEMORY_TYPE_HOST};
```
wence-, line 33: *question*: Memory handles only work with host memory? Or we're only supporting host memory right now?

cpp/src/remote_key.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+  auto serializedString = ss.str();
+
+  // Hash data to provide some degree of confidence on received data.
+  std::stringstream ssHash;
+  std::hash<std::string> hasher;
+  SerializedRemoteKeyHash hash = hasher(serializedString);
+  ssHash.write(reinterpret_cast<char const*>(&hash), sizeof(hash));
+
+  return ssHash.str() + serializedString;
```
wence-, line 108: *question*: I suspect this isn't really performance critical, but why not just `memcpy` the bytes into an allocated buffer rather than using a stringstream?

cpp/src/remote_key.cpp
```diff
@@ -0,0 +1,141 @@
...
+  std::stringstream ss;
+
+  ss.write(reinterpret_cast<char const*>(&_packedRemoteKeySize), sizeof(_packedRemoteKeySize));
+  ss.write(reinterpret_cast<char const*>(_packedRemoteKey), _packedRemoteKeySize);
+  ss.write(reinterpret_cast<char const*>(&_memoryBaseAddress), sizeof(_memoryBaseAddress));
+  ss.write(reinterpret_cast<char const*>(&_memorySize), sizeof(_memorySize));
+
+  auto serializedString = ss.str();
+
+  // Hash data to provide some degree of confidence on received data.
```
wence-, line 102: Is this hashing working around bugginess in UCX, or is it "belt and braces"?

wence- (changes requested):
Mostly some requests for clarifications in docstrings. Overall I think this looks good.

pentschev (commented):
Thanks for the review @wence- , I have replied to your comments. Please have a look when you get another chance.

cpp/include/ucxx/remote_key.h
wence-: Ah, thanks.

cpp/include/ucxx/remote_key.h
wence-: Perhaps "The packed `ucp_rkey_h` key, suitable for transfer to a remote" ?

pentschev: Agreed, this is a better phrasing. Done via https://github.com/rapidsai/ucxx/pull/190/commits/a700b11efa50673b152c4742edc5fbc562d43bc4

wence- (approved):
Just one outstanding question about the required size constraints on a user-provided `buffer`, but again, I think (if it needs it at all) this is just a doc clarification.

Thanks!

cpp/include/ucxx/memory_handle.h
pentschev: No, it's only the case when you're handing over the allocation to UCP (i.e., `buffer=nullptr`), but I see why the confusion, thus I pushed https://github.com/rapidsai/ucxx/pull/190/commits/13233f02e9b54a0555f46771e3e46378ccdfe2b0 in an attempt to clarify it. Please let me know if it's still confusing.

wence-: Looks great, thanks!

cpp/include/ucxx/memory_handle.h
```diff
@@ -0,0 +1,162 @@
...
+                                                          void* buffer);
+
+  ~MemoryHandle();
+
+  /**
+   * @brief Get the underlying `ucp_mem_h` handle.
+   *
+   * Lifetime of the `ucp_mem_h` handle is managed by the `ucxx::MemoryHandle` object and
+   * its ownership is non-transferrable. Once the `ucxx::MemoryHandle` is destroyed the
+   * memory is unampped and the handle is not valid anymore, it is the user's responsibility
```
bdice, line 116:
```suggestion
   * memory is unmapped and the handle is not valid anymore, it is the user's responsibility
```

bdice (approved):

cpp/include/ucxx/memory_handle.h
pentschev: Thanks for catching that, a bit disappointed codespell didn't catch it though.

robertmaynard (approved):

PR #194, opened 2024-02-19
Add `ucxx::Request` callback to constructor arguments

pentschev:
Callbacks had to be assigned during the constructor because the member attributes are part of the base class. Now the base class exposes those arguments via the constructor and they can then be passed directly to the base class' constructor.


cpp/src/request.cpp
```diff
@@ -210,7 +214,8 @@ void Request::setStatus(ucs_status_t status)
 
     if (_status != UCS_INPROGRESS)
       ucxx_error(
-        "ucxx::Request: %p, setStatus called with status: %d (%s) but status: %d (%s) was already "
+        "ucxx::Request: %p, setStatus called with status: %d (%s) but status: %d (%s) was "
+        "already "
         "set",
```
wence-, line 219:
```suggestion
        "already set",
```

wence- (approved):

pentschev:
Thanks @wence- for reviewing!

PR #196, opened 2024-02-22
Run `distributed-ucxx` CI tests in separate workflow

pentschev:
Most of the unstable tests are tests from `distributed-ucxx`, they run much quicker than `ucxx` Python tests. When the workflow fails we currently need to rerun all `ucxx` tests which take ~80% of the time, this change should help in reducing CI resources usage.


wence- (approved):

ci/test_python_distributed.sh
```diff
@@ -0,0 +1,49 @@
...
+set -euo pipefail
+
+source "$(dirname "$0")/test_common.sh"
+
+rapids-logger "Create test conda environment"
+. /opt/conda/etc/profile.d/conda.sh
+
+rapids-dependency-file-generator \
+  --output conda \
+  --file_key test_python \
```
bdice, line 15: @pentschev The errors might just be an incorrect file key?
```suggestion
  --file_key test_python_distributed \
```

AyodeAwe (approved):

ci/build_wheel.sh
```diff
@@ -34,6 +34,16 @@ if ! rapids-is-release-build; then
...
+sed -r -i "s/rmm(.*)\"/rmm${PACKAGE_CUDA_SUFFIX}\1${alpha_spec}\"/g" ${pyproject_file}
+sed -r -i "s/cudf(.*)\"/cudf${PACKAGE_CUDA_SUFFIX}\1${alpha_spec}\"/g" ${pyproject_file}
+
+# Update cupy package name (different suffix from RAPIDS)
+if [[ $PACKAGE_CUDA_SUFFIX == "-cu12" ]]; then
+    sed -i "s/cupy-cuda11x/cupy-cuda12x/g" ${pyproject_file}
+fi
+
 if [[ ${package_name} == "distributed-ucxx" ]]; then
     sed -r -i "s/rapids-dask-dependency==(.*)\"/rapids-dask-dependency==\1${alpha_spec}\"/g" ${pyproject_file}
```
bdice, line 48: This is now run earlier in the script.
```suggestion
```

bdice (changes requested):

ci/build_wheel.sh
pentschev: Good catch, fixed!

bdice (approved):

PR #198, opened 2024-02-26
Make (de-)allocation in `ucxx::Address` consistent

pentschev:
Previously, `ucxx::Address` was allocating with a handle with `new` and deallocating with `std::free` which is inconsistent. This change uses the `new`/`delete` pair to make it consistent and safer.


pentschev:
@mdemoret-nv as per your observation last week.

wence- (approved):

pentschev:
Thanks @wence- .

PR #206, opened 2024-03-13
Unset Python Endpoint close callback upon destruction

pentschev:
In some cases, particularly with `distributed-ucxx`, the C++ Endpoint outlives the Python object that owns a close callback registered with `set_close_callback()`, making it an invalid reference and causing segfaults. To prevent that it is necessary to remove the close callback, preventing it from being called after it's not valid anymore.


python/distributed-ucxx/distributed_ucxx/tests/test_ucxx.py
```diff
@@ -356,7 +356,7 @@ async def test_cuda_context(
                     )
 
 
-@gen_test()
+@gen_test(timeout=60)
```
pentschev, line 359: With some local experimentation it looks like teardown failures are in essence a `gen_test` timeout, but it doesn't show up as such. I tried increasing the timeout here but one of the tests nevertheless failed. Locally though I see this test completing in under 10 seconds or in 60 seconds with this change even though it ultimately passes after the full 60 seconds. With that I'm trying to understand whether this is a problem with UCXX or if this is something with the Distributed test suite (e.g., `gen_test`) as I would expect it to fail every time if it times out.

python/distributed-ucxx/distributed_ucxx/ucxx.py
```diff
@@ -281,6 +281,9 @@ def __init__(  # type: ignore[no-untyped-def]
 
         logger.debug("UCX.__init__ %s", self)
 
+    def __del__(self) -> None:
+        self.abort()
```
wence-, line 285: *question*: Generally speaking we can't rely on `__del__` to run as a finalizer. I think what you're wanting to do here is ensure that `self._ep.abort()` is called when the object is gc'd. Can you do that by calling `weakref.finalize(self, self._ep.abort)` when the inner endpoint is created? I think that won't create a cycle so should be safe. Downside of that is that the `_ep` property will not be set to `None`. However, since this is only called in teardown, the object `self` no longer exists anyway.

python/distributed-ucxx/distributed_ucxx/utils_test.py
```diff
@@ -55,13 +55,19 @@ def ucxx_exception_handler(event_loop, context):
...
 @pytest.fixture(scope="function")
-def ucxx_loop():
+def ucxx_loop(request):
     """Allows UCX to cancel progress tasks before closing event loop.
 
     When UCX tasks are not completed in time (e.g., by unexpected Endpoint
     closure), clean up tasks before closing the event loop to prevent unwanted
     errors from being raised.
+
+    Additionally add a `ignore_alive_references` marker that will override
```
wence-, line 65:
```suggestion
    Additionally add an `ignore_alive_references` marker that will override
```

wence- (approved):
Minor tidying suggestions

python/distributed-ucxx/distributed_ucxx/ucxx.py
pentschev: I tried that exact approach yesterday and it seems to abort endpoints too early, causing some of the tests to fail. I didn't spend too much time trying to figure out why that happens, but for some reason it's not straightforward as you suggested, unfortunately.

I agree `__del__` can't be relied as a finalizer generally, so I added it for some extra peace-of-mind, knowing it will call `self._ep.abort()` under-the-hood to remove the endpoint close callback before it is ultimately destroyed.

wence-: That _suggests_ though I don't know why exactly, that the endpoint is being aborted before (for example) something that relies on the endpoint. One scenario I could envisage that happening in, I think, is if the endpoint is part of a ref-cycle. The cyclic GC will then break the cycle _somewhere_ and then collect in some order, which could result in an endpoint being collected (and therefore aborted) _before_ somethat relies on it. Maybe https://cython.readthedocs.io/en/latest/src/userguide/extension_types.html#disabling-cycle-breaking-tp-clear is relevant.

wence-: My thought was just that pure-python classes that don't hold "external-style" data are not subject to the same problem. But if the ref-cycle exists in Python + some cython classes then if the GC decides to break the cycle using one of the cython classes, since they manage "foreign" memory which it doesn't know about, it can't keep things alive long enough in the right order. One might be able to fix this by annotating all the cython classes with `@no_gc_clear`, but I don't think we should do it here.

pentschev: Yeah, makes sense. I think we should do some investigation into this in the future, the problem is this is just too time consuming, but it needs to be done as this is likely related to the issues we see at teardown.

pentschev:
I just reverted the commit increasing timeout for `test_transpose` as that doesn't have any effect. If there are no objections I'll merge it shortly after CI passes.

pentschev:
Thanks for taking the time for reviewing @wence- !

python/distributed-ucxx/distributed_ucxx/ucxx.py
wence-: https://github.com/rapidsai/ucxx/issues/209

PR #212, opened 2024-03-16
Mark some `distributed-ucxx` tests as flaky and rerun

pentschev:
Some tests fail during teardown which probably indicates there are some resources that leak in certain situations. While this should definitely be resolved we weren't able to do so yet, thus to prevent os many CI failures we now mark those tests flaky if they fail at teardown (when UCX is reset).


wence- (approved):
I think this is OK.

PR #225, opened 2024-04-26
Implement the buffer protocol directly to stop requiring numpy's C API

vyasr:
Currently ucxx makes use of numpy's C API exactly once, in order to expose host buffers to Python. It does so by leveraging the fact that numpy arrays can be constructed directly from a pointer using the C API. However, we can produce essentially an equivalent result by using numpy's Python object on a representation of the same pointer in terms of a custom object that implements the buffer protocol. This adds about 20 lines of Cython code in exchange for dropping numpy as a build dependency entirely, which is a nice maintenance benefit (see also https://github.com/rapidsai/build-planning/issues/41).


vyasr:
CC @seberg

python/ucxx/_lib/libucxx.pyx
```diff
@@ -40,15 +41,45 @@ include "tag.pyx"
 logger = logging.getLogger("ucx")
 
 
-np.import_array()
-
+cdef class HostBufferAdapter:
```
jakirkham, line 44: As we already have [`Array`]( https://github.com/rapidsai/ucxx/blob/bdd3ff95a6646f642543095bccee3c7b0db3a985/python/ucxx/_lib/arr.pyx#L73 ), could we just use that? Or if `Array` is missing something, could we add whatever we need to it?

If possible, would like to move towards one object that does this job for us consistently

jakirkham (commented):
Thanks Vyas! 🙏

Had a question below

python/ucxx/_lib/libucxx.pyx
vyasr: I don't think that makes sense here. As we discussed in the meeting, Array and objects like it are designed to be consumers of various protocols, but they shouldn't support raw pointers. Any object that we want to be consumable by Array needs to export one of the supported protocols. In this case, we're trying to construct a numpy array from a C++ object, the HostBuffer. Unless we want to implement the buffer protocol on the HostBuffer object directly (that requires changing ucxx's C++ code, but is an option), the easiest thing we can do is produce a wrapper around it that does export the protocol. 

The more appropriate usage of Array here would be to use it in place of the numpy array that we're returning, since that that point we're using numpy as a consumer of the buffer protocol and the Array should be able to do that too. That would be an API-breaking change though, and not one that buys us anything.

jakirkham: > Unless we want to implement the buffer protocol on the HostBuffer object directly (that requires changing ucxx's C++ code, but is an option)

Yes this is what I'm suggesting

vyasr: But then how would you construct the Array from the HostBuffer?

seberg (commented):
Overall looks good to me, it isn't like any of this is public API, beyond that there is an object that owns the data.

Some silly nitpicks, but the `free` logic should be fixed.

python/ucxx/_lib/libucxx.pyx
```diff
@@ -40,15 +41,45 @@ include "tag.pyx"
...
+        buffer.suboffsets = NULL
 
-cdef ptr_to_ndarray(void* ptr, np.npy_intp N):
-    cdef np.ndarray[np.uint8_t, ndim=1, mode="c"] arr = (
-        np.PyArray_SimpleNewFromData(1, &N, np.NPY_UINT8, <np.uint8_t*>ptr)
-    )
-    PyArray_ENABLEFLAGS(arr, NPY_ARRAY_OWNDATA)
-    return arr
+    def __releasebuffer__(self, Py_buffer *buffer):
+        free(self._ptr)
```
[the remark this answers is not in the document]
vyasr: Good call on tying the lifetime to this object instead, I agree that makes more sense.

> you could still implement it, since it disallows some limited downstream usage so if we need it later having it blocks that, but it doesn't matter really

Could you explain this statement? For now I've changed the implementation to a no-op, but it sounds like you're saying the mere existence of this method changes the behavior in some potential meaningful (and potentially desirable?) way.

python/ucxx/_lib/libucxx.pyx
```diff
@@ -40,15 +41,45 @@ include "tag.pyx"
...
+        return obj
+
+    def __getbuffer__(self, Py_buffer *buffer, int flags):
+        buffer.buf = self._ptr
+        buffer.format = 'c'
+        buffer.internal = NULL
+        buffer.itemsize = self._itemsize
+        buffer.len = self._size * self._itemsize
+        buffer.ndim = 1
+        buffer.obj = self
```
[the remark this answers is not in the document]
vyasr: Fair point, we'd be at the mercy of Cython setting up appropriate handling of errors to properly decrement the ref count here otherwise you'd like. I'll reorder.

python/ucxx/_lib/libucxx.pyx
```diff
@@ -40,15 +41,45 @@ include "tag.pyx"
...
+
+        This factory takes ownership of the input host_buffer's data, so
+        attempting to use the input after this function is called will result
+        in undefined behavior.
+        """
+        cdef HostBufferAdapter obj = HostBufferAdapter.__new__(HostBufferAdapter)
+        obj._size = host_buffer.getSize()
+        obj._ptr = host_buffer.release()
+        obj._shape = [obj._size]
+        obj._itemsize = sizeof(uint8_t)
```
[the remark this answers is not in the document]
vyasr: Yeah I decided to avoid the magic number here but I agree that it's equivalent.

python/ucxx/_lib/libucxx.pyx
vyasr: If I'm understanding his statements correctly, @seberg and I are basically in agreement here. @jakirkham could you elaborate on how you envisioned this working and whether the above discussion clears up any confusion?

python/ucxx/_lib/libucxx.pyx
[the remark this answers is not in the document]
seberg: CPython has some the parsing functions `PyArg_ParseTuple("%s", buffer)` which only work if a buffer-like doesn't implement `__releasebuffer__` (because that usage pattern cannot call the release buffer function).
So implementing it as a no-op disallows those contexts and may make adding it later easier.

I noticed that cupy also does that for it's pinned memory, but not sure they had a specific reason.

python/ucxx/_lib/libucxx.pyx
seberg: Right, I think what makes this object distinct and useful is that it owns its data and `Array` does not.  That seems like a useful split.
That probably means it won't be a helper for exporting the buffer protocol elsewhere, but that is probably good as the code here is simple and doesn't have to deal with complexities like shapes, strides, dtypes, ….

python/ucxx/_lib/libucxx.pyx
```diff
@@ -40,15 +41,48 @@ include "tag.pyx"
...
+        obj._strides = [obj._itemsize]
+        return obj
+
+    def __getbuffer__(self, Py_buffer *buffer, int flags):
+        buffer.buf = self._ptr
+        buffer.format = 'c'
```
seberg, line 70:
```suggestion
        buffer.format = 'B'
```
randomly noticed the tests were failing, `c` translates to a Python bytes/NumPy string dtype of length 1. `B` translates to the unsigned byte it was before.

vyasr: Ah so this is a signed vs unsigned discrepancy?

python/ucxx/_lib/libucxx.pyx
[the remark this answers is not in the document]
vyasr: I see, I've seen those functions but I didn't realize that the existence of this method would block that. Good context, thanks.

seberg (approved):
LGTM, and I think it would be fine to put this in as is just to get rid of linking NumPy.  It is effectively private right now anyway, so not worth too much API bikeshed.
(If someone prefers to make that obvious with a leading `_` that seems fine.)

If I want to bring things up:
* I am not sure that `HostBufferAdapter` is the best name, but "adapter" is the only current function, so it seems good.
* If you care about overheads (probably no need to), there might be some lower hanging fruits like:
  * Make shape/stride the scalars they are (and dereference them) and even if not assign from scalar rather than list (but maybe cython optimizes both away!).
  * Using `@cython.freelist` could reduce overhead a bit.

Really just bringing that up!  I don't think either matters (API effectively private and array creation probably takes much longer anyway).

pentschev (approved):
This looks good, thanks @vyasr for addressing it. I think we can now [remove the NumPy package from the `CMakeLists.txt`](https://github.com/rapidsai/ucxx/blob/faa22fb061b28b3df10f66f9bda0023eda9fd367/python/ucxx/_lib/CMakeLists.txt#L16-L22), or is there any NumPy dependency left I'm overlooking?

pentschev (approved):
LGTM, thanks again @vyasr !

PR #226, opened 2024-04-29
Use UCX wheels

vyasr:
This PR updates UCX to stop bundling UCX libraries, instead preferring to get them from the new UCX wheels that we are publishing from https://github.com/rapidsai/ucx-wheels.

### DIFF
--- a/ci/build_wheel.sh
+++ b/ci/build_wheel.sh
@@ -18,8 +18,6 @@ commit=$(git rev-parse HEAD)
 
 RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
 
-libucx_wheelhouse=$(RAPIDS_PY_WHEEL_NAME="${RAPIDS_PY_CUDA_SUFFIX}" rapids-get-pr-wheel-artifact ucx-wheels 1 cpp)
-
 # This is the version of the suffix with a preceding hyphen. It's used
 # everywhere except in the final wheel name.
 PACKAGE_CUDA_SUFFIX="-${RAPIDS_PY_CUDA_SUFFIX}"
@@ -59,7 +57,7 @@ if [[ ${package_name} == "distributed-ucxx" ]]; then
     RAPIDS_PY_WHEEL_NAME="distributed_ucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-upload-wheels-to-s3 ${package_dir}/dist
 elif [[ ${package_name} == "ucxx" ]]; then
     SKBUILD_CMAKE_ARGS="-DUCXX_ENABLE_RMM=ON" \
-        python -m pip wheel "${package_dir}"/ -w "${package_dir}"/dist -vvv --no-deps --disable-pip-version-check --find-links ${libucx_wheelhouse}
+        python -m pip wheel "${package_dir}"/ -w "${package_dir}"/dist -vvv --no-deps --disable-pip-version-check
 
     python -m auditwheel repair -w ${package_dir}/final_dist --exclude "libucp.so.0" ${package_dir}/dist/*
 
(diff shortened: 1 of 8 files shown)


dependencies.yaml
```diff
@@ -123,6 +125,12 @@ dependencies:
           - librmm==24.6.*
           - ninja
           - spdlog>=1.12.0,<1.13
+  ucx_wheel:
+    common:
+      - output_types: [requirements, pyproject]
+        packages:
+          # TODO: Figure out versioning
+          - libucx==1.14.1
```
pentschev, line 133:
```suggestion
          - libucx>=1.15.0
```

python/pyproject.toml
```diff
@@ -6,6 +6,7 @@ build-backend = "scikit_build_core.build"
 requires = [
     "cmake>=3.26.4",
     "cython>=3.0.0",
+    "libucx==1.14.1",
```
pentschev, line 9:
```suggestion
    "libucx>=1.15.0",
```

python/pyproject.toml
```diff
@@ -22,6 +23,7 @@ authors = [
 license = { text = "BSD-3-Clause" }
 requires-python = ">=3.9"
 dependencies = [
+    "libucx==1.14.1",
```
pentschev, line 26:
```suggestion
    "libucx>=1.15.0",
```

python/ucxx/__init__.py
```diff
@@ -6,6 +6,17 @@
...
 
+# If libucx was installed as a wheel, we must request it to load the library symbols.
+# Otherwise, we assume that the library was installed in a system path that ld can find.
+try:
+    import libucx
+except ModuleNotFoundError:
+    pass
+else:
+    libucx.load_library()
+    del libucx
```
pentschev, line 17: This is a bit unexpected to my eyes. Why won't the libraries be found when loading the shared libraries with wheels?

pentschev (changes requested):
Thanks @vyasr , left a few comments and some minor change requests as per our offline discussion.

ci/build_wheel.sh
```diff
@@ -6,6 +6,10 @@ set -euo pipefail
 package_name=$1
 package_dir=$2
 
+# Clear out system ucx files to ensure that we're getting ucx from the wheel.
+rm -rf /usr/lib64/ucx
+rm -rf /usr/lib64/libuc*
+
```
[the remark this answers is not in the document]
vyasr: I added this because [UCX is installed in the wheel containers](https://github.com/rapidsai/ci-imgs/blob/main/ci-wheel.Dockerfile#L140) and I wanted to make sure that we weren't accidentally depending on it. I would prefer to leave this here until we update the CI containers to be safe, but we need to update other RAPIDS repos to use the new UCX wheels before we can update the containers.

python/ucxx/__init__.py
vyasr: Yes, this is something I'm introducing in the new wheels. The idea is that wheels don't install to a standard location, so the loader doesn't know where those libraries are. The only way you could make them be found is by doing something like setting the RPATH. RPATH modification is fragile, though, especially across package boundaries. It relies on specific file layouts etc that we cannot guarantee. The approach here basically makes it so that each Python package is responsible for loading the libraries that it bundles (basically via dlopen) so that they're visible to future libraries. That way, we don't need to deal with cross-package RPATH hacking; instead, each package is responsible for loading its own libraries and consumers just call `load_library` like this.

vyasr: I pointed @pentschev to some resources offline, but tl;dr yes this is a potential problem, but this is about as good of a solution as we're going to get if we go the dynamic linking route. The only way to completely avoid these issues is with static linking, which we're trying to move away from in RAPIDS in general due to issues like binary size considerations (and also that isn't a viable solution for UCX anyway due to its extensive usage of dlopen and the way the library itself is written requiring a dynamic cudart).

python/pyproject.toml
vyasr: No worries, always good to check.

raydouglass (approved):

pentschev (approved):
LGTM, thanks @vyasr !

PR #228, opened 2024-05-06
remove 'tomli' dependency

jameslamb:
Proposes removing the build-time dependency on `tomli` for wheels and conda packages.

It doesn't appear to be used anywhere here.

```shell
git grep tomli
```

## Notes for Reviewers

I originally noticed something similar in `ucx-py` (https://github.com/rapidsai/ucx-py/pull/1042), then went searching for similar cases across RAPIDS.

That dependency was added for `distributed-ucxx` back in #60. I'm not sure why, but I suspect it was related to the use of `versioneer` in this project at the time. Reference: https://github.com/python-versioneer/python-versioneer/issues/338#issuecomment-1381170813

This project doesn't use `versioneer` any more (#114). I strongly suspect that the dependency on `tomli` can be removed.


wence-:
Hmm, I think you are right, and that the toml parsing that _must_ be required by the build backend just installs tomli in the build env for python < 3.11

wence- (approved):
Seems probable.

jameslamb:
Yep, I think so! `scikit-build-core` will pull in `tomli` if it needs it: https://github.com/scikit-build/scikit-build-core/blob/f6ed5a28fc85e621b03d984011d17def888ee0db/pyproject.toml#L41

jakirkham (approved):

raydouglass (approved):

jakirkham:
Thanks all! 🙏

PR #238, opened 2024-06-24
Add several fixes primarily targeting worker progress thread

pentschev:
# Add several fixes primarily targeting worker progress thread

Several changes are added to improve stability of worker progress thread. Details of changes are below.

## Support generic callback cancelation

Once registered it was not possible to cancel generic callbacks, which could cause the callback to be called after the timeout had expired and the owner/resources were not alive anymore. This change allows canceling a generic callback that couldn't be completed successfully, avoiding access to invalid memory of objects that are not alive anymore.

## Remove `ErrorCallbackData`

The `ErrorCallbackData` structure and `ucxx::Endpoint::errorCallback` static method used by `ucxx::Endpoint` were removed in favor of using a friend function `endpointErrorCallback` that receives a pointer to `ucxx::Endpoint` itself, thus simplifying ownership of all resources where only `ucxx::Endpoint` now is responsible to retain ownership of all resources.

### Add missing locks

All accesses to any of internal `_trackedRequests` resources are now protected by locks. This has been the source of various issues in the past, thus switching to a safer approach seems wise.

During `merge()` both merger and mergee have resource locks now to prevent races.

### Use regular instances instead of pointer for `TrackedRequests`

To simplify `ucxx::InflightRequests`/`TrackedRequests` instead of using pointers to `InflightRequestsMap` objects, switch to using simple instances.

### Replace `std::ref` with `std::shared_ptr`

Using `std::ref` was causing undefined behavior in the state of `_stop`, switching to `std::shared_ptr` has resolved the problem.

### Set progress thread ID before progress loop stats

Set the thread ID with parent before starting the progress loop, so that checking whether the a callback is running in the progress thread is always correct, otherwise the parent may be unable to determine the progress thread has already started.

### Use simple instance instead of pointer

New members, such as `isRunning()` and `stop()`, were introduced so a simple instance of `WorkerProgressThread` may be used instead of smart pointers. With this change the owner of the instance may query whether the thread indeed stopped, as opposed to just letting it go out-of-scope while the thread was not joined.

## Improvement to `CallbackNotifier`

Instead of blocking continuously for longer periods of time, it's now possible to periodically signal the worker again to cause it to wakeup (when in blocking mode). Currently defaults to 100ms.


wence- (commented):
A few comments, but haven't looked at everything yet

cpp/include/ucxx/delayed_submission.h
```diff
@@ -40,28 +43,33 @@ template <typename T>
...
   std::string _name{"undefined"};  ///< The human-readable name of the collection, used for logging
-  bool _enabled{true};  ///< Whether the resource required to process the collection is enabled.
-  std::vector<T> _collection{};  ///< The collection.
-  std::mutex _mutex{};           ///< Mutex to provide access to `_collection`.
+  bool _enabled{true};    ///< Whether the resource required to process the collection is enabled.
+  ItemIdType _itemId{0};  ///< The item ID counter, used to allow cancelation.
+  std::deque<std::pair<ItemIdType, T>> _collection{};  ///< The collection.
+  std::set<ItemIdType> _canceled{};                    ///< IDs of canceled items.
+  std::mutex _mutex{};          ///< Mutex to provide access to `_collection`.
+  std::mutex _canceledMutex{};  ///< Mutex to provide access to `_canceled`.
```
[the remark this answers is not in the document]
pentschev: Good catch, I intended to remove it but ended up forgetting, done in https://github.com/rapidsai/ucxx/pull/238/commits/e1f123d1dcfa00a0e0c9443a76f93056944fed68.

cpp/include/ucxx/delayed_submission.h
```diff
@@ -122,20 +136,47 @@ class BaseDelayedSubmissionCollection {
...
-      ucxx_trace_req("Submitting %lu %s callbacks", itemsToProcess.size(), _name.c_str());
-      for (auto& item : itemsToProcess)
-        processItem(item);
+    for (auto i = 0; i < toProcess; ++i) {
+      std::pair<ItemIdType, T> item;
+      {
+        std::lock_guard<std::mutex> lock(_mutex);
+        item = std::move(_collection.front());
+        _collection.pop_front();
+        if (_canceled.erase(item.first)) continue;
```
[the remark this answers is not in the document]
pentschev: That's right, but as you noted we're not making use of it currently, I intended to remove it but forgot to do so. We might need one if we address the `TODO` in `cancel()`, but I think that can be left for a follow-up PR as it's not entirely clear it's necessary.

cpp/include/ucxx/delayed_submission.h
```diff
@@ -122,20 +136,47 @@ class BaseDelayedSubmissionCollection {
...
     }
 
-    if (itemsToProcess.size() > 0) {
-      ucxx_trace_req("Submitting %lu %s callbacks", itemsToProcess.size(), _name.c_str());
-      for (auto& item : itemsToProcess)
-        processItem(item);
+    for (auto i = 0; i < toProcess; ++i) {
+      std::pair<ItemIdType, T> item;
+      {
+        std::lock_guard<std::mutex> lock(_mutex);
```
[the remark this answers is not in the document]
pentschev: The worry is it may take too long, particularly critical if something takes too long on the UCX end to complete. Most of the tasks we use are supposed to be non-blocking, but I've encountered weird issues in very specific scenarios that may occur with one transport but not another. The problem if this occurs is we block the main thread if it's trying to schedule/cancel another request. It may be a bit overreaching and perhaps we could be able to indeed do as you say, but I'd prefer to do it carefully in the future, I'd hate to fix the segfault just to introduce a new potential deadlock here.

pentschev (commented):
Thanks @wence- , I've addressed/responded to your review. Please have a look next week. 🙂

cpp/include/ucxx/endpoint.h
```diff
@@ -83,10 +53,15 @@ class Endpoint : public Component {
...
                                       ///< closed, used for logging purposes only
   bool _endpointErrorHandling{true};  ///< Whether the endpoint enables error handling
-  std::unique_ptr<ErrorCallbackData> _callbackData{
-    nullptr};  ///< Data struct to pass to endpoint error handling callback
-  std::shared_ptr<InflightRequests> _inflightRequests{
-    std::make_shared<InflightRequests>()};  ///< The inflight requests
+  std::unique_ptr<InflightRequests> _inflightRequests{
+    std::make_unique<InflightRequests>()};  ///< The inflight requests
+  std::mutex _mutex{std::mutex()};          ///< Mutex used to prevent race conditions with
+                                            ///< `ucxx::Endpoint::setCloseCallback()`.
```
wence-, line 59: *nit*: Document what this mutex protects?

cpp/include/ucxx/worker_progress_thread.h
```diff
@@ -47,8 +47,8 @@ typedef void* ProgressThreadStartCallbackArg;
  */
 class WorkerProgressThread {
  private:
-  std::thread _thread{};     ///< Thread object
-  bool _stop{false};         ///< Signal to stop on next iteration
+  std::thread _thread{};                                       ///< Thread object
+  std::shared_ptr<bool> _stop{std::make_shared<bool>(false)};  ///< Signal to stop on next iteration
```
wence-, line 51: Why does this now need to be a shared ptr? Is it that because it was passed around by reference between threads previously, it might have gone out of scope on thread A when read on thread B?

cpp/src/inflight_requests.cpp
```diff
@@ -12,31 +12,42 @@ namespace ucxx {
 
 InflightRequests::~InflightRequests() { cancelAll(); }
 
-size_t InflightRequests::size() { return _trackedRequests->_inflight->size(); }
+size_t InflightRequests::size()
+{
+  std::scoped_lock localLock{_mutex};
```
wence-, line 17: This mutex is protecting `_trackedRequests` being rewritten out from under us, and the lock on `->_mutex` is protecting concurrent modification?

wence- (approved):
Minor niggles. There are lots of locks, but AFAICT, they looked right to me.

cpp/include/ucxx/endpoint.h
pentschev: Does https://github.com/rapidsai/ucxx/pull/238/commits/8c3148dd72f5ca7d599b9949c385a77d783edaa8 suits your expectations here?

cpp/include/ucxx/worker_progress_thread.h
pentschev: Initially I thought so, but while debugging I observed `std::ref` containing garbage briefly after starting the thread. I couldn't figure what was happening, the parent was still alive and `std::shared_ptr` resolved the issue so I just decided not to spend much more time on it.

cpp/src/inflight_requests.cpp
pentschev: `InflightRequests::_mutex` (used with what I called `localLock`) is a reentrant lock to prevent two threads from attempting to modify resources (`_trackedRequests`) simultaneously. Reentrancy is required for cases where `cancelAll()` ends up triggering `remove()` (if the object's refcount goes to 0), but we still require exclusive access to modify the actual request trackers separately (inflight/canceling requests) depending on the operation.

wence-: Thanks!

cpp/include/ucxx/worker_progress_thread.h
wence-: Fair.

wence-:
Thanks Peter!

PR #240, opened 2024-06-28
Increase maximum UCX runtime pin

pentschev:
Increase maximum UCX runtime pin to support UCX up to `v1.17.x`, but still build against `v1.15.0` until we gain confidence with newer UCX.

Additionally disable protov2 by default as CUDA async/managed memory is not yet supported and leads to poor performance.

### DIFF
--- a/python/ucxx/__init__.py
+++ b/python/ucxx/__init__.py
@@ -95,6 +95,11 @@ if "UCX_MAX_RNDV_RAILS" not in os.environ and get_ucx_version() >= (1, 12, 0):
     logger.info("Setting UCX_MAX_RNDV_RAILS=1")
     os.environ["UCX_MAX_RNDV_RAILS"] = "1"
 
+if "UCX_PROTO_ENABLE" not in os.environ:
+    # UCX protov2 still doesn't support CUDA async/managed memory
+    logger.info("Setting UCX_PROTO_ENABLE=n")
+    os.environ["UCX_PROTO_ENABLE"] = "n"
+
 
 from ._version import __git_commit__, __version__
 
--- a/ci/test_python.sh
+++ b/ci/test_python.sh
@@ -56,4 +56,4 @@ for nbuf in 1 8; do
 done
 
 rapids-logger "C++ future -> Python future notifier example"
-python -m ucxx.examples.python_future_task_example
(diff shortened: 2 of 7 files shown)


msarahan:
Seems like the (private) cudf testing API may have changed:

```
FAILED python/ucxx/_lib_async/tests/test_custom_send_recv.py::test_send_recv_cudf[<lambda>0] - ImportError: cannot import name 'assert_eq' from 'cudf.testing._utils' (/opt/conda/envs/test/lib/python3.11/site-packages/cudf/testing/_utils.py)
```

pentschev:
> Seems like the (private) cudf testing API may have changed:
> 
> ```
> FAILED python/ucxx/_lib_async/tests/test_custom_send_recv.py::test_send_recv_cudf[<lambda>0] - ImportError: cannot import name 'assert_eq' from 'cudf.testing._utils' (/opt/conda/envs/test/lib/python3.11/site-packages/cudf/testing/_utils.py)
> ```

This was fixed in https://github.com/rapidsai/ucxx/pull/239 but I had an out-of-date branch. Merged upstream now and it should be fixed.

python/ucxx/__init__.py
```diff
@@ -95,6 +95,11 @@ def _is_mig_device(handle):
     logger.info("Setting UCX_MAX_RNDV_RAILS=1")
     os.environ["UCX_MAX_RNDV_RAILS"] = "1"
 
+if "UCX_PROTO_ENABLE" not in os.environ:
+    # UCX protov2 still doesn't support CUDA async/managed memory
+    logger.info("Setting UCX_PROTO_ENABLE=n")
+    os.environ["UCX_PROTO_ENABLE"] = "n"
```
wence-, line 101: Do we document, like ucx-py, where the UCXX settings diverge from the UCX defaults? If so, maybe do that for this one too?

wence- (approved):

jameslamb (approved):

python/ucxx/__init__.py
pentschev: We do not have docs here yet primarily because I didn't want to maintain two copies of it, so I'm planning to copy/move up-to-date docs from UCX-Py when we're about to make the switch.

pentschev:
Thanks everyone for reviews/approvals here.

PR #242, opened 2024-07-02
Fix tests on new Distributed releases

pentschev:
Latest changes in Distributed have broken `test_nanny_closed_by_keyboard_interrupt` as it used a member attribute that is now supposed to be accessed via a method, the test is now updated to rely on the new behavior.


wence- (approved):

pentschev:
Thanks @wence- !

PR #243, opened 2024-07-02
Use verify-alpha-spec hook

KyleFromNVIDIA:
With the deployment of rapids-build-backend, we need to make sure our dependencies have alpha specs.

Contributes to https://github.com/rapidsai/build-planning/issues/31


KyleFromNVIDIA:
This PR has exposed a weakness in the `verify-alpha-spec` hook: since this project's `VERSION` is different from RAPIDS, we need to tell the hook the actual RAPIDS version that we want to use. I've opened https://github.com/rapidsai/pre-commit-hooks/issues/43 to discuss.

jameslamb (approved):

PR #244, opened 2024-07-05
skip CMake 3.30.0

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/80

Adds constraints to avoid pulling in CMake 3.30.0, for the reasons described in that issue.


bdice (approved):

pentschev (approved):
LGTM, thanks @jameslamb !

PR #250, opened 2024-07-23
Option to enable/disable endpoint error handling in C++ benchmark

pentschev:
Add option to allow enabling/disabling endpoint error handling in C++ benchmark, switching the default to disable matching `ucx_perftest`.

### DIFF
--- a/cpp/benchmarks/perftest.cpp
+++ b/cpp/benchmarks/perftest.cpp
@@ -36,14 +36,15 @@ typedef std::shared_ptr<BufferMap> BufferMapPtr;
 typedef std::shared_ptr<TagMap> TagMapPtr;
 
 struct app_context_t {
-  ProgressMode progress_mode = ProgressMode::Blocking;
-  const char* server_addr    = NULL;
-  uint16_t listener_port     = 12345;
-  size_t message_size        = 8;
-  size_t n_iter              = 100;
-  size_t warmup_iter         = 3;
-  bool reuse_alloc           = false;
-  bool verify_results        = false;
+  ProgressMode progress_mode   = ProgressMode::Blocking;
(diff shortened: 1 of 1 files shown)


cpp/benchmarks/perftest.cpp
```diff
@@ -185,6 +190,7 @@ ucs_status_t parseCommand(app_context_t* app_context, int argc, char* const argv
           return UCS_ERR_INVALID_PARAM;
         }
         break;
+      case 'e': app_context->endpoint_error_handling = false; break;
```
wence-, line 193: I think you want `-e` to turn on endpoint error handling (since the default is no error handling?), so:

```suggestion
      case 'e': app_context->endpoint_error_handling = true; break;
```

cpp/benchmarks/perftest.cpp
```diff
@@ -120,6 +124,7 @@ static void printUsage()
   std::cerr << "              'thread-polling' and 'thread-blocking' (default: 'blocking')"
             << std::endl;
   std::cerr << "  -t          use thread progress mode (disabled)" << std::endl;
+  std::cerr << "  -e          create endpoints with error handling support (false)" << std::endl;
```
wence-, line 127:
```suggestion
  std::cerr << "  -e          create endpoints with error handling support (disabled)" << std::endl;
```
?

wence- (commented):
Is the flag setting the wrong way round?

pentschev:
> Is the flag setting the wrong way round?

You're right, thanks for catching. Changes are applied now. 😄

wence- (approved):

pentschev:
Thanks for reviewing/approving @wence- !

PR #252, opened 2024-07-24
Move Python ucxx code into python/ucxx

KyleFromNVIDIA:
Also fix up the isort configuration to get more correct results.

Contributes to https://github.com/rapidsai/build-planning/issues/33


python/pyproject.toml
pentschev, line 1: Why do we need to keep this files if we have one for each of the subprojects `distributed-ucxx` and `ucxx`?

pentschev (approved):
Assuming there's nothing missing (i.e., CI passes), LGTM. Thanks @KyleFromNVIDIA .

python/pyproject.toml
KyleFromNVIDIA: Because this is common `isort`, `ruff`, etc. configuration for all of the projects.

KyleFromNVIDIA:
This will probably require a devcontainers change: https://github.com/rapidsai/devcontainers/blob/541895260fa9d3d772989c55c64a3ab98323f1c4/features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml#L42

cc @trxcllnt

KyleFromNVIDIA:
I've opened https://github.com/rapidsai/devcontainers/pull/372 for this

python/pyproject.toml
pentschev: Ah ok, I didn't know you could use `pyproject.toml` as common config for subprojects. Thanks for the reply!

jameslamb (approved):

vyasr (approved):

PR #255, opened 2024-07-30
Mark C++ constructors and getters `[[nodiscard]]`

pentschev:
Improve compile-time error reports by marking appropriate C++ constructors and getters with `[[nodiscard]]`.

### DIFF
--- a/cpp/python/src/python_future.cpp
+++ b/cpp/python/src/python_future.cpp
@@ -4,7 +4,6 @@
  */
 #include <memory>
 #include <stdexcept>
-#include <tuple>
 #include <utility>
 
 #include <Python.h>
--- a/cpp/python/src/python_future.cpp
+++ b/cpp/python/src/python_future.cpp
@@ -4,6 +4,7 @@
  */
 #include <memory>
 #include <stdexcept>
+#include <tuple>
 #include <utility>
 
 #include <Python.h>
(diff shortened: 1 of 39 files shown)


cpp/python/src/python_future.cpp
```diff
@@ -4,6 +4,7 @@
  */
 #include <memory>
 #include <stdexcept>
+#include <tuple>
 #include <utility>
```
wence-, line 8: As below std::ignore is in `<utility>`
```suggestion
#include <utility>
```

cpp/src/delayed_submission.cpp
```diff
@@ -4,6 +4,7 @@
  */
 #include <memory>
 #include <mutex>
+#include <tuple>
 #include <utility>
```
wence-, line 8:
```suggestion
#include <utility>
```

cpp/src/listener.cpp
```diff
@@ -5,6 +5,8 @@
 #include <memory>
 #include <netinet/in.h>
 #include <string>
+#include <tuple>
```
wence-, line 8:
```suggestion
#include <utility>
```

cpp/src/worker.cpp
```diff
@@ -9,6 +9,7 @@
 #include <queue>
 #include <sstream>
 #include <string>
+#include <tuple>
 #include <utility>
```
wence-, line 13:
```suggestion
#include <utility>
```

cpp/tests/buffer.cpp
```diff
@@ -4,6 +4,7 @@
  */
 #include <algorithm>
 #include <numeric>
+#include <tuple>
 #include <utility>
```
wence-, line 8:
```suggestion
#include <utility>
```

`std::ignore` should already be in `<utility>` (since C++11), so no need for the `<tuple>` include I think?

wence- (approved):
Trivial possible changes.

pentschev:
> `std::ignore` should already be in `<utility>` (since C++11), so no need for the `<tuple>` include I think?

You're right, thanks for catching that. I misread the docs. Thanks for the quick review too!

PR #257, opened 2024-07-31
Separate libucxx build and ucxx build

KyleFromNVIDIA:
Rather than building `libucxx_python.so` in the `libucxx` build, and transferring it to the `ucxx` package, build it in the `ucxx` build.

Nothing other than the Python wheel is currently using `ucxx::python`, so stop exporting it and installing the headers as well.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -40,9 +40,10 @@ HELP="$0 [clean] [libucxx] [libucxx_python] [ucxx] [distributed_ucxx] [benchmark
    default action (no args) is to build and install 'libucxx' and 'libucxx_python', then 'ucxx' targets, and finally 'distributed_ucxx'
 "
 LIB_BUILD_DIR=${LIB_BUILD_DIR:=${REPODIR}/cpp/build}
+PYTHON_BUILD_DIR=${PYTHON_BUILD_DIR:=${REPODIR}/cpp/python/build}
 UCXX_BUILD_DIR=${REPODIR}/python/ucxx/build
 
-BUILD_DIRS="${LIB_BUILD_DIR} ${UCXX_BUILD_DIR}"
+BUILD_DIRS="${LIB_BUILD_DIR} ${PYTHON_BUILD_DIR} ${UCXX_BUILD_DIR}"
 
 # Set defaults for vars modified by flags to this script
 VERBOSE_FLAG=""
@@ -53,7 +54,7 @@ BUILD_TESTS=OFF
 BUILD_EXAMPLES=OFF
 BUILD_DISABLE_DEPRECATION_WARNINGS=ON
 BUILD_COMPILE_COMMANDS=OFF
-UCXX_ENABLE_RMM=ON
+UCXX_ENABLE_RMM=OFF
 
 # Set defaults for vars that may not have been defined externally
 #  FIXME: if INSTALL_PREFIX is not set, check PREFIX, then check
@@ -137,6 +138,10 @@ if hasArg --show_depr_warn; then
     BUILD_DISABLE_DEPRECATION_WARNINGS=OFF
 fi
 
+if buildAll || hasArg libucxx_python; then
+  UCXX_ENABLE_RMM=ON
+fi
+
 # If clean given, run it prior to any other steps
 if hasArg clean; then
     # If the dirs to clean are mounted dirs in a container, the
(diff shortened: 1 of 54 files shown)


bdice:
The import test for `ucxx` is failing with `ImportError: libucxx_python.so`. It comes from here: https://github.com/rapidsai/ucxx/blob/95f4edacf502520ff6cbf7b74fba216b7f244cc4/conda/recipes/ucxx/meta.yaml#L258-L259

python/ucxx/CMakeLists.txt
```diff
@@ -24,11 +26,12 @@ project(
 option(FIND_UCXX_CPP "Search for existing UCXX C++ installations before defaulting to local files"
        OFF
 )
+option(BUILD_SHARED_LIBS "Build UCXX shared libraries" ON)
```
pentschev, line 29:
```suggestion
option(BUILD_SHARED_LIBS "Build UCXX Python shared libraries" ON)
```

KyleFromNVIDIA:
Thanks for the review @pentschev. This PR is still very much a work in progress. I'll address whatever portions of your review are still relevant once I've actually got it working.

build.sh
```diff
@@ -218,6 +213,34 @@ fi
 # Replace spaces with semicolons in SKBUILD_EXTRA_CMAKE_ARGS
 SKBUILD_EXTRA_CMAKE_ARGS=$(echo ${EXTRA_CMAKE_ARGS} | sed 's/ /;/g')
 
+# Build and install libucxx_python.so
+if buildAll || hasArg libucxx_python; then
+    CMAKE_GENERATOR="${CMAKE_GENERATOR:-Ninja}"
+    pwd
```
bdice, line 219:
```suggestion
```

build.sh
```diff
@@ -218,6 +213,34 @@ fi
...
+          -DCMAKE_BUILD_TYPE=${BUILD_TYPE} \
+          -DCMAKE_EXPORT_COMPILE_COMMANDS=${BUILD_COMPILE_COMMANDS} \
+          ${EXTRA_CMAKE_ARGS}
+
+    cd ${PYTHON_BUILD_DIR}
+
+    compile_start=$(date +%s)
+    cmake --build . -j${PARALLEL_LEVEL} ${VERBOSE_FLAG}
+    compile_end=$(date +%s)
+    compile_total=$(( compile_end - compile_start ))
```
bdice, line 233: Do we typically report times like this?

cpp/python/CMakeLists.txt
```diff
@@ -3,13 +3,78 @@
...
 set(UCXX_PYTHON_LIB Python3::Python)
 
+# ##################################################################################################
+# * compiler options ------------------------------------------------------------------------------
+# Due to https://github.com/openucx/ucx/issues/9614, we cannot export the ucx
+# dependency because users would then have no control over whether ucx is found
+# multiple times, causing potential configure errors. Therefore, we use a raw
+# find_package call instead of rapids_find_package and skip exporting the ucx
+# dependency. Consumers of ucxx must find ucx themselves. Once we move the
+# minimum version to UCX 1.16 we can remove the above find_package in favor of
```
bdice, line 63: What would it take for us to set a lower bound of UCX 1.16 and simplify this? I think we are already using 1.17 according to https://github.com/rapidsai/build-planning/issues/77. Maybe we just need to bump the lower bound?

cpp/python/CMakeLists.txt
```diff
@@ -3,13 +3,78 @@
...
+project(
+  UCXX_PYTHON
+  VERSION ${libucxx_version}
+  LANGUAGES C CXX
+)
+
+option(BUILD_SHARED_LIBS "Build UCXX Python shared libraries" ON)
+option(FIND_UCXX_CPP "Search for existing UCXX C++ installations before defaulting to local files"
+       OFF
+)
```
pentschev, line 31: I'm guessing that when `BUILD_SHARED_LIBS=ON` then `libucxx.so` is found transitively, but if `BUILD_SHARED_LIBS=OFF` then `FIND_UCXX_CPP=ON` is necessary, is that right?

cpp/python/CMakeLists.txt
```diff
@@ -3,13 +3,78 @@
...
+# source if it cannot be found (i.e. both cases must allow prior finding of
+# ucx).
+if(NOT TARGET ucx::ucp)
+  find_package(ucx REQUIRED)
+endif()
+#rapids_find_package(
+#  ucx REQUIRED
+#  BUILD_EXPORT_SET ucxx-exports
+#  INSTALL_EXPORT_SET ucxx-exports
+#)
```
pentschev, line 76:
```suggestion
```

pentschev: Should this be removed?

cpp/python/CMakeLists.txt
```diff
@@ -68,13 +135,23 @@ endif()
...
-                  EXCLUDE_FROM_ALL
+)
+install(TARGETS ucxx_python DESTINATION ${CMAKE_INSTALL_LIBDIR} EXPORT ucxx-python-exports)
+
+rapids_export(
+  INSTALL ucxx-python
+  EXPORT_SET ucxx-python-exports
+  GLOBAL_TARGETS python
+  NAMESPACE ucxx::
 )
```
pentschev, line 148:
```suggestion
```

pentschev: This seems to be duplicated below.

pentschev (commented):
Looks mostly good to me, left a few comments/suggestions.

cpp/python/CMakeLists.txt
KyleFromNVIDIA: It's not a duplicate. This one is `INSTALL`, while the one below is `BUILD`.

cpp/python/CMakeLists.txt
KyleFromNVIDIA: No, because `ucxx::python` has a `PUBLIC` dependency on `ucx::ucp` and thus needs to export it (once the minimum is 1.16).

build.sh
KyleFromNVIDIA: This and the `pwd` were both copied and pasted from another part of the code. Nothing is even printing `compile_total`, so I've removed this and the `pwd` from both places.

cpp/python/CMakeLists.txt
KyleFromNVIDIA: No. `BUILD_SHARED_LIBS` merely determines whether static (`.a`) or shared (`.so`) libraries are built. It has nothing to do with whether or not `FIND_UCXX_CPP` is set.

pentschev: Got it, but when is `FIND_UCXX_CPP` required here then? I assume it always needs to be found, unless maybe `libucxx_python.so` is doing that transitively, is that the case?

cpp/python/CMakeLists.txt
pentschev: Oh right, I missed that. Why does `INSTALL` come first?

cpp/python/CMakeLists.txt
KyleFromNVIDIA: `FIND_UCXX_CPP` is set to `ON` for the Conda package, which depends on the `libucxx` Conda package. It's currently set to `OFF` for the wheel, which instead builds `libucxx` through `add_subdirectory()`, but a future PR that creates a `libucxx` wheel will also set it to `ON`.

cpp/python/CMakeLists.txt
KyleFromNVIDIA: It doesn't matter which comes first. The order is arbitrary. I guess the reason is that `INSTALL` came first in `cpp/CMakeLists.txt`.

cpp/python/CMakeLists.txt
pentschev: I see now, thanks for clarifying.

cpp/python/CMakeLists.txt
pentschev: Huh, I didn't know order was arbitrary, thanks for the info!

pentschev (approved):
LGTM now, thanks @KyleFromNVIDIA .

jameslamb (approved):
Took a quick look, I don't have any additional comments. Thanks!

PR #260, opened 2024-08-07
Add libucxx wheel

KyleFromNVIDIA:
Contributes to https://github.com/rapidsai/build-planning/issues/33


ci/build_wheel.sh
```diff
@@ -17,12 +17,23 @@ if [[ ${package_name} == "distributed-ucxx" ]]; then
...
     RAPIDS_PY_WHEEL_NAME="distributed_ucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-upload-wheels-to-s3 ${package_dir}/dist
-elif [[ ${package_name} == "ucxx" ]]; then
-    SKBUILD_CMAKE_ARGS="-DUCXX_ENABLE_RMM=ON;-DCMAKE_INSTALL_LIBDIR=ucxx/lib64;-DCMAKE_INSTALL_INCLUDEDIR=ucxx/include" \
+elif [[ ${package_name} == "libucxx" ]]; then
+    SKBUILD_CMAKE_ARGS="-DUCXX_ENABLE_RMM=ON" \
         python -m pip wheel "${package_dir}"/ -w "${package_dir}"/dist -vvv --no-deps --disable-pip-version-check
 
     python -m auditwheel repair -w ${package_dir}/final_dist --exclude "libucp.so.0" ${package_dir}/dist/*
 
+    RAPIDS_PY_WHEEL_NAME="libucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-upload-wheels-to-s3 ${package_dir}/final_dist
```
msarahan, line 26:
```suggestion
    RAPIDS_PY_WHEEL_NAME="libucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-upload-wheels-to-s3 cpp ${package_dir}/final_dist
```

ci/build_wheel.sh
```diff
@@ -17,12 +17,23 @@ if [[ ${package_name} == "distributed-ucxx" ]]; then
...
+    CPP_WHEELHOUSE=$(RAPIDS_PY_WHEEL_NAME="libucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 cpp /tmp/libucxx_dist)
+    echo "libucxx-${RAPIDS_PY_CUDA_SUFFIX} @ file://$(echo ${CPP_WHEELHOUSE}/libucxx_*.whl)" > "${package_dir}/constraints.txt"
+
+    PIP_CONSTRAINT="${package_dir}/constraints.txt" \
+    SKBUILD_CMAKE_ARGS="-DFIND_UCXX_CPP=ON;-DCMAKE_INSTALL_LIBDIR=ucxx/lib64;-DCMAKE_INSTALL_INCLUDEDIR=ucxx/include" \
+        python -m pip wheel "${package_dir}"/ -w "${package_dir}"/dist -vvv --no-deps --disable-pip-version-check
+
+    python -m auditwheel repair -w ${package_dir}/final_dist --exclude "libucp.so.0" --exclude "libucxx.so" ${package_dir}/dist/*
+
     RAPIDS_PY_WHEEL_NAME="ucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-upload-wheels-to-s3 ${package_dir}/final_dist
```
msarahan, line 37:
```suggestion
    RAPIDS_PY_WHEEL_NAME="ucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-upload-wheels-to-s3 python ${package_dir}/final_dist
```

msarahan (commented):
These edits might get you on the right track. The argument that is `cpp` or `python` is backwards compatible, but if not present, it behaves as just `python`, and the filenames don't line up.

dependencies.yaml
```diff
@@ -56,32 +56,57 @@ files:
...
       - rapids_build_skbuild
+  py_rapids_build_libucxx:
+    output: pyproject
+    pyproject_dir: python/libucxx
+    extras:
+      table: tool.rapids-build-backend
+      key: requires
+    includes:
+      - build_cpp
+      - depends_on_librmm
```
pentschev, line 74: I was going to say that this is introducing a hard dependency on RMM, but it's already the case in conda packages. Ideally we would have one that doesn't depend on RMM for non-CUDA users but we will need to resolve that for both conda and wheels in the future.

pentschev (approved):
LGTM, thanks @KyleFromNVIDIA .

dependencies.yaml
KyleFromNVIDIA: Wouldn't that involve creating a package like `libucxx-nocuda` or something?

pentschev: Yes, either that or doing what UCX does and `dlopen` resources, although RMM is header-only so we may need to do some other runtime checks instead, and this is why it was done yet.

python/ucxx/pyproject.toml
```diff
@@ -72,6 +73,7 @@ ninja.make-fallback = true
 sdist.exclude = ["*tests*"]
 sdist.reproducible = true
 wheel.packages = ["ucxx"]
+wheel.exclude = ["*.pyx", "*.pxd", "CMakeLists.txt", "CPPLINT.cfg", ".clang-format"]
```
bdice, line 76: I'm not sure if I understand why we have `.clang-format` and `CPPLINT.cfg` files in `python/ucxx/ucxx/examples/`. Maybe @pentschev knows? If we can rely on the clang-format config in the root of the repo, that would be better imo.

bdice (commented):
A few comments -- overall this looks good!

ci/build_wheel.sh
```diff
@@ -16,14 +16,25 @@ rapids-generate-version > ./VERSION
...
+    SKBUILD_CMAKE_ARGS="-DUCXX_ENABLE_RMM=ON" \
         python -m pip wheel "${package_dir}"/ -w "${package_dir}"/dist -vvv --no-deps --disable-pip-version-check
 
     python -m auditwheel repair -w ${package_dir}/final_dist --exclude "libucp.so.0" ${package_dir}/dist/*
 
-    RAPIDS_PY_WHEEL_NAME="ucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-upload-wheels-to-s3 ${package_dir}/final_dist
+    RAPIDS_PY_WHEEL_NAME="libucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-upload-wheels-to-s3 cpp ${package_dir}/final_dist
+elif [[ ${package_name} == "ucxx" ]]; then
+    CPP_WHEELHOUSE=$(RAPIDS_PY_WHEEL_NAME="libucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 cpp /tmp/libucxx_dist)
+    echo "libucxx-${RAPIDS_PY_CUDA_SUFFIX} @ file://$(echo ${CPP_WHEELHOUSE}/libucxx_*.whl)" > "${package_dir}/constraints.txt"
```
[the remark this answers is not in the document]
KyleFromNVIDIA: > Is this meant to enforce that the ucxx wheel build uses the exact libucxx wheel artifact built in the same PR?

Yes. This pattern was lifted from kvikio:

https://github.com/rapidsai/kvikio/blob/ab84fdf24964cbb7cda517e5be22f907c4f931d8/ci/build_wheel_python.sh#L20-L27

ci/test_wheel_distributed_ucxx.sh
```diff
@@ -9,12 +9,14 @@ source "$(dirname "$0")/test_common.sh"
...
-RAPIDS_PY_WHEEL_NAME="${package_name}_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 ./dist
-ucxx_wheelhouse=$(RAPIDS_PY_WHEEL_NAME="ucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 ./local-ucxx-dep)
+RAPIDS_PY_WHEEL_NAME="${package_name}_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 python ./dist
+ucxx_wheelhouse=$(RAPIDS_PY_WHEEL_NAME="ucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 python ./local-ucxx-dep)
+libucxx_wheelhouse=$(RAPIDS_PY_WHEEL_NAME="libucxx_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 cpp ./local-libucxx-dep)
 
 python -m pip install \
     -v \
     "${ucxx_wheelhouse}"/ucxx_${RAPIDS_PY_CUDA_SUFFIX}*.whl \
+    "${libucxx_wheelhouse}"/libucxx_${RAPIDS_PY_CUDA_SUFFIX}*.whl \
```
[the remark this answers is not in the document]
KyleFromNVIDIA: It shouldn't make a difference, but I'll move it anyway.

dependencies.yaml
```diff
@@ -298,6 +327,26 @@ dependencies:
             packages: &cupy_packages_cu11
               - cupy-cuda11x>=12.0.0
           - {matrix: null, packages: *cupy_packages_cu11}
+  depends_on_librmm:
+    common:
+      - output_types: conda
+        packages:
+          - &librmm_conda librmm==24.10.*,>=0.0.0a0
```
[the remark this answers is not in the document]
KyleFromNVIDIA: Updated

python/ucxx/pyproject.toml
KyleFromNVIDIA: Those files were introduced in https://github.com/rapidsai/ucxx/pull/46. My guess is they were copied-and-pasted from somewhere else by mistake, because that PR did not add any C++ files to this directory. I'll remove them and take them out of `wheel.exclude`.

bdice (approved):
Can you comment on the impact on wheel sizes? LGTM.

PR #267, opened 2024-08-14
Test tag user callback with discarded return

pentschev:
We were recently inquired by a user whether relying only on the `ucs_status_t` passed to the user callback and discarding the `ucxx::RequestTag` return from `ucxx::Endpoint::tagSend` and `ucxx::Endpoint::tagRecv` is valid.  This turns out to be a valid usage, although I'm unsure whether encouraging this is a good idea, since requests such as `ucxx::Endpoint::amRecv` require that the user retrieve the resulting buffer from the returned `ucxx::RequestAm`, and if this is discarded the buffer is lost. This PR adds a test for the aforementioned use case but makes no changes to documentation to prevent encouraging this pattern until we can decide whether we should support it or not.

For requests such as `ucxx::Endpoint::amRecv`, it might be worth studying whether we could pass the resulting buffer and any other attributes associated with it to the callback, in that case we may be able to provide a safe pattern to always use the callback if the user doesn't want to keep a reference to the returned `ucxx::Request` object.


wence- (approved):

pentschev:
Thanks @wence- for approving, since this is only adding a test, I'll go ahead and merge it even though we're in code freeze already.

pentschev:
Oh, actually we need admin merge. Since this isn't critical for 24.10 I'll then retarget and merge it once tests pass.

PR #269, opened 2024-08-15
Make several attempts to clone Distributed in case of failure

pentschev:
CI may eventually fail due to unsuccessful attempts to clone the Distributed repository. This change ensures multiple attempts are made to clone it before failing.

### DIFF
--- a/ci/test_common.sh
+++ b/ci/test_common.sh
@@ -173,7 +173,23 @@ install_distributed_dev_mode() {
   # to run non-public API tests in CI.
 
   rapids-logger "Install Distributed in developer mode"
-  git clone https://github.com/dask/distributed /tmp/distributed -b 2024.1.1
+  MAX_ATTEMPTS=5
+  for attempt in $(seq 1 $MAX_ATTEMPTS); do
+    rm -rf /tmp/distributed
+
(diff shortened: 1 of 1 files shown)


AyodeAwe (approved):

AyodeAwe:
Why would this git clone fail though?

pentschev:
> Why would this git clone fail though?

I think this is either due to network errors or GH limits, most likely the first, the errors are not very clear. Two instances from tonight's nightlies [here](https://github.com/rapidsai/ucxx/actions/runs/10400558430/job/28801341190#step:8:2832) and [here](https://github.com/rapidsai/ucxx/actions/runs/10400558430/job/28801341468#step:8:2949) have different failures, first:

```
Cloning into '/tmp/distributed'...
error: RPC failed; curl 56 GnuTLS recv error (-54): Error in the pull function.
fatal: early EOF
fatal: fetch-pack: invalid index-pack output
Error: Process completed with exit code 128.
```

Second:

```
Cloning into '/tmp/distributed'...
fatal: unable to access 'https://github.com/dask/distributed/': GnuTLS recv error (-110): The TLS connection was non-properly terminated.
Error: Process completed with exit code 128.
```

pentschev:
Thanks for the approval @AyodeAwe , I'm going ahead and merging this.

PR #270, opened 2024-08-16
Restore `*.pxd` files in ucxx wheel

KyleFromNVIDIA:
`*.pxd` files were mistakenly removed in https://github.com/rapidsai/ucxx/pull/260. Resume installing them in the ucxx wheel.


vyasr (approved):

vyasr:
Note that none of ucxx's current pxd files are really intended for public consumption, so this change isn't strictly necessary, but there's no harm bundling them and in the event that ucxx does choose to expose a Cython API in the future this could be helpful.

PR #271, opened 2024-08-19
Remove NumPy <2 pin

seberg:
This PR removes the NumPy<2 pin. This is desirable for RAPIDS projects that do not have a strong numpy/cupy dependency.
Not doing any math, they don't run into subtle issues, but the pin was still necessary briefly before NumPy 2 was initially released.

(Other RAPIDS projects rely heavily on CuPy/NumPy and need the soon to be released 13.3.0 to function properly, e.g. due to promotion related changes in NumPy.)


conda/environments/all_cuda-118_arch-x86_64.yaml
```diff
@@ -26,7 +26,7 @@ dependencies:
 - libtool
 - ninja
 - numba>=0.57.1
-- numpy>=1.23,<2.0a0
+- numpy>=1.23,<3
```
msarahan, line 29:
```suggestion
- numpy>=1.23,<3.0a0
```
I think you want to keep the a0 junk. If you don't, then pre-release versions of v3.0 will be allowed.

I could well be operating on inaccurate historical information, though.

msarahan (approved):

jameslamb (approved):
Looks good to me, based on my understanding of https://github.com/rapidsai/ucx-py/pull/1064#discussion_r1725786361.

Please remove the language in the description about this being unsafe combined with `cupy < 13.3.0` before merging.

jakirkham:
Sebastian rewrote the description based on the feedback in the UCX-Py PR and updated it here as well

jakirkham (approved):

jakirkham:
Thanks all! 🙏

PR #273, opened 2024-08-22
fix sed expression in update-version.sh

AyodeAwe:


raydouglass (approved):
Related to https://github.com/rapidsai/ucxx/pull/261

PR #274, opened 2024-08-23
Update docs to suggest Miniforge3

pentschev:
Due to licensing restrictions, suggest users install Miniforge3 from now on.


wence- (approved):

pentschev:
Thanks @wence- !

PR #277, opened 2024-09-11
Update version error message

pentschev:
Update version error message to match the minimum support UCX version and provide instructions to check where UCX is being loaded from.

Ported from https://github.com/rapidsai/ucx-py/pull/1069

### DIFF
--- a/python/ucxx/ucxx/__init__.py
+++ b/python/ucxx/ucxx/__init__.py
@@ -103,10 +103,13 @@ if "UCX_PROTO_ENABLE" not in os.environ:
 
 from ._version import __git_commit__, __version__
 
+__ucx_min_version__ = "1.15.0"
 __ucx_version__ = "%d.%d.%d" % get_ucx_version()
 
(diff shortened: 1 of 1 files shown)


python/ucxx/ucxx/__init__.py
```diff
@@ -103,10 +103,13 @@ def _is_mig_device(handle):
 
 from ._version import __git_commit__, __version__
 
+__ucx_min_version__ = "1.15.0"
 __ucx_version__ = "%d.%d.%d" % get_ucx_version()
 
-if get_ucx_version() < (1, 11, 1):
+if get_ucx_version() < tuple([int(i) for i in __ucx_min_version__.split(".")]):
```
madsbk, line 109:
```suggestion
if get_ucx_version() < tuple(int(i) for i in __ucx_min_version__.split(".")):
```

madsbk (approved):
Looks good, I only have a minor nitpick

pentschev:
Thanks @madsbk for the review/approval!

PR #281, opened 2024-09-24
Fix setting Python futures when they have already completed

pentschev:
Prevent calling `set_result()`/`set_exception()` on Python futures if they are already `cancelled()`/`done()`. Calling `set_result()`/`set_exception()` when the future has already completed raises an exception that may be uncatched as it may happen on a thread different from the application, which seems to have been one of the reasons why `distributed-ucxx` tests have been hanging.

Although not all test failures have been resolved by this, hangs have seemingly also been less frequent in local tests.

Additionally set more sensible log levels for other errors coming from CPython implementation and remove the `PyErr_Print()` which will just print a message to stdout instead of a log message.


wence- (approved):
This makes sense to me, thanks.

pentschev:
Thanks for approval @wence- . Since this has potential to impact at least tests in RAFT/cuML and we're already in code freeze I've retargeted, but I'll merge it once tests pass.

PR #283, opened 2024-09-25
Update RMM imports to use `pylibrmm`

Matt711:
This PR updates the RMM import to use `pylibrmm` now that `rmm._lib` is deprecated . It should be merged after https://github.com/rapidsai/rmm/pull/1676.


pentschev (approved):
LGTM, thanks @Matt711 !

PR #291, opened 2024-09-30
Rewrite distributed-ucxx comms resources tracking

pentschev:
Rewrite tracking of distributed-ucxx communicator resources providing more robust lifetime tracking, allowing better control over termination of notifier and progress threads.

This seems to have resolved deadlock issues in Distributed for good this time and if not, it definitely improved the status substantially. Another benefit is the removal of `distributed_patches.py`, meaning there's no need anymore to monkey-patch Distributed resources to do any tracking, and instead do that on the communication resources alone which are entirely controlled by `distributed-ucxx`, in the long-run it will also cause less breakage since we were previously bound to start/stop mechanisms in Distributed to remain unchanged.


python/distributed-ucxx/distributed_ucxx/ucxx.py
```diff
@@ -90,13 +90,84 @@ def synchronize_stream(stream=0):
...
+
+    Register a Dask resource with the UCXX context to keep track of it, so that
+    the notifier thread and progress tasks may be stopped when no more resources
+    """
+    ctx = ucxx.core._get_ctx()
+
+    with ctx._dask_resources_lock:
+        ctx._dask_resources.add(resource)
+        ctx.start_notifier_thread()
+        ctx.continuous_ucx_progress()
```
wence-, line 116: A little confused here. AIUI, `ucxx.core._get_ctx()` returns a global `ApplicationContext` object. And `ctx.start_notifier_thread` creates a new thread in the application context, overwriting any existing running thread.

So does this mean it's not really "safe" to create more than one "dask resource" per process (and hence per application context). Or does it not matter that we throw away the `ctx.notifier_thread` and replace it with a new one when we call `start_notifier_thread`?

python/distributed-ucxx/distributed_ucxx/ucxx.py
```diff
@@ -90,13 +90,84 @@ def synchronize_stream(stream=0):
...
+
+    # Check if the attribute exists first, in tests the UCXX context may have
+    # been reset before some resources are deregistered.
+    if hasattr(ctx, "_dask_resources_lock"):
+        with ctx._dask_resources_lock:
+            try:
+                ctx._dask_resources.remove(resource)
+            except KeyError:
+                pass
+            _stop_notifier_thread_and_progress_tasks()
```
wence-, line 141: I _think_ this function is only called here. Perhaps prefer to inline it?

It just does:
```
if len(ctx._dask_resources) == 0:
    ctx.stop_notifier_thread()
    ctx.progress_tasks.clear()
```

python/distributed-ucxx/distributed_ucxx/ucxx.py
```diff
@@ -90,13 +90,84 @@ def synchronize_stream(stream=0):
...
+    Allocate a Dask resources tracker in the UCXX context. This is useful to
+    track Distributed communicators so that progress and notifier threads can
+    be cleanly stopped when no UCXX communicators are alive anymore.
+    """
+    ctx = ucxx.core._get_ctx()
+    if not hasattr(ctx, "_dask_resources"):
+        from threading import Lock
+
+        ctx._dask_resources = set()
+        ctx._dask_resources_lock = Lock()
```
wence-, line 156: *note*: If multiple threads are running there is a potential for a race condition here, but I'm not sure if it matters.

python/distributed-ucxx/distributed_ucxx/ucxx.py
```diff
@@ -279,8 +353,12 @@ def __init__(  # type: ignore[no-untyped-def]
         else:
             self._has_close_callback = False
 
+        _register_dask_resource(hash(self))
+
         logger.debug("UCX.__init__ %s", self)
 
+        weakref.finalize(self, _deregister_dask_resource, hash(self))
```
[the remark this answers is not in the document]
wence-: *question*: It seems like this hash of `self` is merely used as a key so that `_deregister_dask_resource` is only ever called once per UCXX comm object.

Since there is a (probably low) potential for hash collisions, would it make sense for instead `_register_dask_resource` to return a sequence number to be used in deregister:

```python

def make_register():
    count = itertools.count()
    def register() -> int:
        tag = next(count)
        ctx._resources.add(tag)
        ...
        return tag
    return register

_register_dask_resource = make_register()

...

self.tag = _register_dask_resource()

weakref.finalize(self, _deregister_dask_resource, self.tag)
...
```

WDYT?

python/distributed-ucxx/distributed_ucxx/ucxx.py
wence-: Specifically: suppose we don't have the attribute and two threads are running.

Both threads can enter line 172. Then suppose thread A continues, and sets the attributes, and continues using the lock to update `_dask_resources`. Eventually thread B goes forward to line 174, it now overwrites `_dask_resources`, so we could lose any resource tags that thread A had added.

wence- (approved):
Tiny docstring movement suggestion

python/distributed-ucxx/distributed_ucxx/ucxx.py
pentschev: Good point, you're indeed correct. Initially I only had `_allocate_dask_resources_tracker()` being called after `ucxx.init(...)`, but later I found the need to add it also to https://github.com/rapidsai/ucxx/pull/291/files#diff-e8b6ecb786c905744fe6e10d9547e9aaffbd732fb70750d5385ac9641babb924R185-R189 due to tests resetting the context.

Currently, I don't think this can happen in a different thread though but I may be overlooking something. One way I think we could prevent risking concurrency on creating the lock is to assign the lock and resource set to this file, such that it is only created once next to `ucxx.init(...)` instead of pushing it into the `ApplicationContext`, technically even that isn't thread-safe but I don't think we ever had a problem with it so it's "safe". If you think it's worth it I can do that but would prefer to address it in a follow-up PR as it will definitely need me to do some more robust testing again on that and I'd still like to get this integrated soon as it's already better than it was before.

With the above being said, I realize now there's a different issue here which is the impossibility of starting a new test that uses a different set of UCX variables in the same process, because `ucxx.init()` is still only called once. I think this is currently not a problem though, as the only such tests are spawned in new processes in https://github.com/rapidsai/ucxx/blob/branch-0.41/python/distributed-ucxx/distributed_ucxx/tests/test_ucxx_config.py .

python/distributed-ucxx/distributed_ucxx/ucxx.py
```diff
@@ -90,13 +91,102 @@ def synchronize_stream(stream=0):
...
+    ctx = ucxx.core._get_ctx()
+
+    # Check if the attribute exists first, in tests the UCXX context may have
+    # been reset before some resources are deregistered.
+    if hasattr(ctx, "_dask_resources_lock"):
+        with ctx._dask_resources_lock:
+            try:
+                ctx._dask_resources.remove(resource_id)
+            except KeyError:
+                pass
```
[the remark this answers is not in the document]
pentschev: That's all correct. I've also relied on the unique values of resource references/hashes for debugging, so that was more useful for debugging like that. As you say, I also don't think it's worth it but may be done if we find a strong reason to follow that path.

python/distributed-ucxx/distributed_ucxx/ucxx.py
```diff
@@ -90,38 +91,51 @@ def synchronize_stream(stream=0):
...
+    `_deregister_dask_resource(resource_id)`, which stops the notifier thread
+    and progress tasks when no more UCXX resources are alive.
 
-    If no Dask resources that make use of UCXX communicator are running anymore,
-    stop the notifier thread and progress tasks to allow for clean shutdown.
+    Returns
+    -------
+    resource_id: int
+        The ID of the registered resource that should be used with
+        `_deregister_dask_resource` during stop/destruction of the resource.
```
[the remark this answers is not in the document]
pentschev: Done in https://github.com/rapidsai/ucxx/pull/291/commits/3b2e7c7f7d31910de4fd608e9eecf8bab4f1c5e1 .

python/distributed-ucxx/distributed_ucxx/ucxx.py
```diff
@@ -90,13 +91,102 @@ def synchronize_stream(stream=0):
...
+            ctx._dask_resources.add(resource_id)
+            ctx.start_notifier_thread()
+            ctx.continuous_ucx_progress()
+            return resource_id
+
+    return register
+
+
+_register_dask_resource = make_register()
+
```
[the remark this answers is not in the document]
pentschev: Also done in https://github.com/rapidsai/ucxx/pull/291/commits/3b2e7c7f7d31910de4fd608e9eecf8bab4f1c5e1 .

python/distributed-ucxx/distributed_ucxx/ucxx.py
wence-: > on creating the lock is to assign the lock and resource set to this file, such that it is only created once next to ucxx.init(...) instead of pushing it into the ApplicationContext

You mean, make the lock a file-level (i.e. module-level variable)? That is thread safe because `import` is protected by a lock in the python interpreter.

Overall, I think as is, this is fine. Perhaps add a comment that creating the comm objects is not thread safe, so we remember if in the future we do need thread safety.

pentschev: Added a comment in https://github.com/rapidsai/ucxx/pull/291/commits/48fa0e5ebc39de249434877db6db3440fb842f33 . I'll nevertheless try to address this in a follow-up PR today or next week.

pentschev:
Thanks @wence- for the thorough review!

PR #293, opened 2024-10-03
Fix issues creating endpoints and with wireup

pentschev:
Fix bug with `create_endpoint`, where a `UCXMessageTruncatedError` may not be properly raised, as well as its use in `distributed-ucxx` not raising the exception expected by `distributed`.

Additionally, fix wireup issues in tests, where only sending (or receiving) a wireup message does not seem to suffice, but sending _and_ receiving suffices to prevent errors in tests.


wence- (approved):

pentschev:
Thanks for the approval @wence- !

PR #295, opened 2024-10-04
Resolve thread-safety issues in distributed-ucxx

pentschev:
Instead of creating a per-UCXX context lock that may result in race conditions during the creation of the locks, use a module-level lock that is guaranteed to be thread-safe as it occurs at import time.

### DIFF
--- a/python/distributed-ucxx/distributed_ucxx/ucxx.py
+++ b/python/distributed-ucxx/distributed_ucxx/ucxx.py
@@ -14,6 +14,7 @@ import os
 import struct
 import weakref
 from collections.abc import Awaitable, Callable, Collection
+from threading import Lock
 from typing import TYPE_CHECKING, Any
 from unittest.mock import patch
 
@@ -49,6 +50,8 @@ device_array = None
 pre_existing_cuda_context = False
 cuda_context_created = False
 multi_buffer = None
+_resources_lock = Lock()
+_resources = dict()
 
 
 _warning_suffix = (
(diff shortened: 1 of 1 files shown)


python/distributed-ucxx/distributed_ucxx/ucxx.py
```diff
@@ -110,9 +113,14 @@ def register() -> int:
             `_deregister_dask_resource` during stop/destruction of the resource.
         """
         ctx = ucxx.core._get_ctx()
-        with ctx._dask_resources_lock:
+        handle = ctx.context.handle
+        with _resources_lock:
+            handle = ctx.context.handle
```
wence-, line 118:
```suggestion
        handle = ctx.context.handle
        with _resources_lock:
```

Don't need to set the handle variable twice.

python/distributed-ucxx/distributed_ucxx/ucxx.py
```diff
@@ -49,6 +50,8 @@
 pre_existing_cuda_context = False
 cuda_context_created = False
 multi_buffer = None
+_resources_lock = Lock()
+_resources = dict()
```
wence-, line 54: Let's add a bit of documenation here.
```suggestion
# Lock protecting access to _resources dict
_resources_lock = Lock()
# Mapping from UCXX context handles to sets of registered dask resource IDs
# Used to track when there are no more users of the context, at which point
# its progress task and notification thread can be shut down.
# See _register_dask_resource and _deregister_dask_resource.
_resources = dict()
```

wence- (approved):
Small suggestions, looks good.

pentschev (commented):
Comment the purpose of new resources

PR #303, opened 2024-10-17
Prevent canceling uncancelable generic callbacks

pentschev:
Generic callbacks should not be canceled once they start running. The guarantee provided by `ucxx::Worker` is that once the function returns it should be safe to destroy the callback and all its associated resources, which becomes invalid if the callback is scheduled for cancellation but it is already running, therefore, it's a requirement to check whether the callback is already executing and block it until it's finished. If the callback never completes this may cause an irrecoverable hang which cannot be dealt with from UCXX since it's impossible to stop a callback from executing once it has started, it's the user's responsibility to guarantee the callback must return. A warning is raised after multiples of 10 attempts have been tried to cancel a callback that is being executed and canceling did not succeed, so that the user is informed of what is happening.

The most notable issue is somewhat frequently observable in CI, where the Python async test `test_from_worker_address_multinode` would segfault, in particular with larger amount of endpoints. This was observable in those tests more frequently because there's a large amount of endpoints being created simultaneously by multiple processes, putting more pressure in the resources and causing endpoint creation to take several seconds to complete. In those cases the generic callback executing `ucp_ep_create` would take longer than the default timeout of 3 seconds and in some cases that would be interpreted as the callback timed out, since `ucp_ep_create` itself took longer than 3 seconds, causing the worker to attempt to cancel the callback while it was still executing. With this change, the callback will still timeout but only if it didn't start executing yet, if `ucp_ep_create` ends up never returning, this will cause a deadlock in the application but there's no way for UCXX to recover on its own and warnings are raised, although those hypothetical deadlocks have not been observed in local tests so far. Segfaults should not occur in this situation anymore.

Additionally, unit tests for generic callbacks are now included, which previously were a gap in the testing suite.

### DIFF
--- a/cpp/include/ucxx/delayed_submission.h
+++ b/cpp/include/ucxx/delayed_submission.h
@@ -45,6 +45,8 @@ class BaseDelayedSubmissionCollection {
   std::string _name{"undefined"};  ///< The human-readable name of the collection, used for logging
   bool _enabled{true};    ///< Whether the resource required to process the collection is enabled.
   ItemIdType _itemId{0};  ///< The item ID counter, used to allow cancelation.
+  std::optional<ItemIdType> _processing{
+    std::nullopt};  ///< The ID of the item being processed, if any.
   std::deque<std::pair<ItemIdType, T>> _collection{};  ///< The collection.
   std::set<ItemIdType> _canceled{};                    ///< IDs of canceled items.
   std::mutex _mutex{};  ///< Mutex to provide access to `_collection`.
@@ -150,9 +152,16 @@ class BaseDelayedSubmissionCollection {
         item = std::move(_collection.front());
         _collection.pop_front();
         if (_canceled.erase(item.first)) continue;
+        _processing = std::optional<ItemIdType>{item.first};
       }
 
       processItem(item.first, item.second);
+    }
+
+    {
+      // Clear the value of `_processing` as no more requests will be processed.
+      std::lock_guard<std::mutex> lock(_mutex);
+      _processing = std::nullopt;
     }
   }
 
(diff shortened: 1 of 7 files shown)


cpp/tests/worker.cpp
```diff
@@ -43,16 +56,18 @@ class WorkerCapabilityTest : public ::testing::Test,
   }
 };
 
-class WorkerProgressTest : public WorkerTest,
-                           public ::testing::WithParamInterface<std::tuple<bool, ProgressMode>> {
+class WorkerProgressTest
+  : public WorkerTest,
+    public ::testing::WithParamInterface<std::tuple<bool, ProgressMode, ExtraParams>> {
```
pentschev, line 61: There's no way for derived classes to inherit again from `WithParamInterface`, thus added an `ExtraParams` attribute to the base class that can be used by derived classes to signify other things to be tested. Not completely happy with this solution, but the alternative seems to be duplicating the entire `WorkerProgressTest` class, potentially more than once.

cpp/tests/worker.cpp
```diff
@@ -67,6 +82,10 @@ class WorkerProgressTest : public WorkerTest,
   }
 };
 
+class WorkerGenericCallbackTest : public WorkerProgressTest {};
+
+class WorkerGenericCallbackSingleTest : public WorkerProgressTest {};
```
pentschev, line 87: It seems that the only form to specify only certain tests should run with a specific parameter set is to have a different derived class to appropriately inform `INSTANTIATE_TEST_SUITE_P`, therefore these classes just inherit the parent to explicitly allow selecting a different set of tests to run.

wence- (approved):
One very minor comment, looks good

cpp/include/ucxx/delayed_submission.h
```diff
@@ -150,10 +152,17 @@ class BaseDelayedSubmissionCollection {
         item = std::move(_collection.front());
         _collection.pop_front();
         if (_canceled.erase(item.first)) continue;
+        _processing = std::optional<ItemIdType>{item.first};
       }
 
       processItem(item.first, item.second);
     }
```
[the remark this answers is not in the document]
pentschev: Right, but the loop may continue and we need to acquire `_mutex` to prevent a race with `cancel()`, so I think it should be a reasonable compromise to just do it at the end of the loop https://github.com/rapidsai/ucxx/pull/303/files#diff-be1ae222c4b0723e2817858a72e0815d919a1c20816ffebfa18ff7dc06a4457fR162-R164. WDYT?

pentschev:
Thanks @wence- for the review, since you've pre-approved and I'd very much like to see how this will impact upcoming CI runs I'm gonna move ahead and merge this in now. However, there's still the open question in https://github.com/rapidsai/ucxx/pull/303#discussion_r1812411503, I'm happy to address that in a follow-up PR if that is still an appropriate change.

PR #309, opened 2024-10-23
Extend Python benchmarks backends

pentschev:
Extend Python benchmarks with `socket` and `asyncio.Stream{Reader,Writer}` APIs. This should help us better understand the overhead of UCXX compared to Python's internal implementations, and thus help us improve potential suboptimal code in our implementation.


wence- (approved):
Looks like a good idea!

pentschev:
Thanks @wence- !

PR #313, opened 2024-11-01
Update to CCCL 2.7.0-rc2.

bdice:
This PR updates to CCCL 2.7.0-rc2. Do not merge until all of RAPIDS is ready to update.

Part of https://github.com/rapidsai/build-planning/issues/115.

### DIFF
--- a/fetch_rapids.cmake
+++ b/fetch_rapids.cmake
@@ -11,6 +11,10 @@
 # or implied. See the License for the specific language governing permissions and limitations under
 # the License.
 # =============================================================================
+
+set(rapids-cmake-repo bdice/rapids-cmake)
+set(rapids-cmake-branch cccl-2.7.0-rc2)
+
 if(NOT EXISTS ${CMAKE_CURRENT_BINARY_DIR}/UCXX_RAPIDS.cmake)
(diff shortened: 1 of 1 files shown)


pentschev:
Why do we need to explicitly add it to UCXX since it doesn't use it directly?

bdice:
Ah, maybe I didn’t need this PR. I had ucxx in the list of repos from my last CCCL update automation and maybe this is a mistake. I’ll double check some things on my side and close if unnecessary.

bdice:
I think previously, this repo was using RMM via CMake CPM, which would trigger RMM CMake logic and thus rapids-cmake's pinnings. That is why we previously we had test PRs for ucxx. However, now ucxx is using librmm wheels, so there isn't a way to trigger the RMM CMake logic directly (it finds the local librmm package and doesn't need to fetch CCCL).

For this migration, we _could_ test upstream artifacts by fetching the librmm wheel from https://github.com/rapidsai/rmm/pull/1715 and using it in this build. However, we think the risk of this migration is low, especially given the advances in CCCL-RAPIDS joint testing, so we are hoping that this migration can be done without the need to test with upstream artifacts. It takes significantly more effort to test with upstream artifacts.

I will close this PR and mark it as complete.

PR #315, opened 2024-11-11
Fixes for AM and enable AM tests that were previously skipped

pentschev:
Fix AM receive message cancelation in C++ implementation that requires a custom `cancel()` implementation, as well as log when allocator raised an exception and return `UCXNoMemoryError` to requests where allocation failed. Reimplement Python AM RMM allocator with a pure C++ function to prevent Cython from introducing Python exception handlers that should not occur in the allocators as the C++ backend should not require the GIL.

Enable AM tests that were previously skipped and update `test_send_recv_two_workers` to match AM implementation.

### DIFF
--- a/ci/test_common.sh
+++ b/ci/test_common.sh
@@ -128,7 +128,7 @@ run_py_tests_async() {
   ENABLE_PYTHON_FUTURE=$3
   SKIP=$4
 
-  CMD_LINE="UCXPY_PROGRESS_MODE=${PROGRESS_MODE} UCXPY_ENABLE_DELAYED_SUBMISSION=${ENABLE_DELAYED_SUBMISSION} UCXPY_ENABLE_PYTHON_FUTURE=${ENABLE_PYTHON_FUTURE} timeout 20m python -m pytest -vs python/ucxx/ucxx/_lib_async/tests/ --durations=50"
+  CMD_LINE="UCXPY_PROGRESS_MODE=${PROGRESS_MODE} UCXPY_ENABLE_DELAYED_SUBMISSION=${ENABLE_DELAYED_SUBMISSION} UCXPY_ENABLE_PYTHON_FUTURE=${ENABLE_PYTHON_FUTURE} timeout 20m python -m pytest -vs python/ucxx/ucxx/_lib_async/tests/ --runslow"
 
   if [ $SKIP -ne 0 ]; then
     echo -e "\e[1;33mSkipping unstable test: ${CMD_LINE}\e[0m"
--- a/cpp/src/request_am.cpp
+++ b/cpp/src/request_am.cpp
@@ -268,19 +268,29 @@ ucs_status_t RequestAm::recvCallback(void* arg,
       amHeader.memoryType = UCS_MEMORY_TYPE_HOST;
     }
 
-    std::shared_ptr<Buffer> buf = amData->_allocators.at(amHeader.memoryType)(length);
+    try {
+      buf = amData->_allocators.at(amHeader.memoryType)(length);
+    } catch (const std::exception& e) {
+      ucxx_debug("Exception calling allocator: %s", e.what());
+    }
 
     auto recvAmMessage =
       std::make_shared<internal::RecvAmMessage>(amData, ep, req, buf, receiverCallback);
 
-    ucp_request_param_t request_param = {.op_attr_mask = UCP_OP_ATTR_FIELD_CALLBACK |
-                                                         UCP_OP_ATTR_FIELD_USER_DATA |
-                                                         UCP_OP_ATTR_FLAG_NO_IMM_CMPL,
-                                         .cb        = {.recv_am = _recvCompletedCallback},
-                                         .user_data = recvAmMessage.get()};
+    ucp_request_param_t requestParam = {.op_attr_mask = UCP_OP_ATTR_FIELD_CALLBACK |
+                                                        UCP_OP_ATTR_FIELD_USER_DATA |
+                                                        UCP_OP_ATTR_FLAG_NO_IMM_CMPL,
+                                        .cb        = {.recv_am = _recvCompletedCallback},
+                                        .user_data = recvAmMessage.get()};
+
+    if (buf == nullptr) {
+      ucxx_debug("Failed to allocate %lu bytes of memory", length);
+      recvAmMessage->_request->setStatus(UCS_ERR_NO_MEMORY);
+      return UCS_ERR_NO_MEMORY;
+    }
 
     ucs_status_ptr_t status =
-      ucp_am_recv_data_nbx(worker->getHandle(), data, buf->data(), length, &request_param);
+      ucp_am_recv_data_nbx(worker->getHandle(), data, buf->data(), length, &requestParam);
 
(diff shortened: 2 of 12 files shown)


ci/test_common.sh
```diff
@@ -128,7 +128,7 @@ run_py_tests_async() {
   ENABLE_PYTHON_FUTURE=$3
   SKIP=$4
 
-  CMD_LINE="UCXPY_PROGRESS_MODE=${PROGRESS_MODE} UCXPY_ENABLE_DELAYED_SUBMISSION=${ENABLE_DELAYED_SUBMISSION} UCXPY_ENABLE_PYTHON_FUTURE=${ENABLE_PYTHON_FUTURE} timeout 20m python -m pytest -vs python/ucxx/ucxx/_lib_async/tests/ --durations=50"
+  CMD_LINE="UCXPY_PROGRESS_MODE=${PROGRESS_MODE} UCXPY_ENABLE_DELAYED_SUBMISSION=${ENABLE_DELAYED_SUBMISSION} UCXPY_ENABLE_PYTHON_FUTURE=${ENABLE_PYTHON_FUTURE} timeout 20m python -m pytest -vs python/ucxx/ucxx/_lib_async/tests/ --runslow"
```
wence-, line 131: Deliberately dropping the `--durations` flag?

cpp/src/request_am.cpp
```diff
@@ -302,7 +332,7 @@ ucs_status_t RequestAm::recvCallback(void* arg,
       return UCS_INPROGRESS;
     }
   } else {
-    std::shared_ptr<Buffer> buf = amData->_allocators.at(UCS_MEMORY_TYPE_HOST)(length);
+    buf = amData->_allocators.at(UCS_MEMORY_TYPE_HOST)(length);
```
wence-, line 335: *question*: Should we check for failed allocations here too?

python/ucxx/ucxx/_lib/libucxx.pyx
```diff
@@ -95,9 +95,20 @@ def _get_host_buffer(uintptr_t recv_buffer_ptr):
...
 
 
-cdef shared_ptr[Buffer] _rmm_am_allocator(size_t length):
-    cdef shared_ptr[RMMBuffer] rmm_buffer = make_shared[RMMBuffer](length)
-    return dynamic_pointer_cast[Buffer, RMMBuffer](rmm_buffer)
+# A pure C++ is used here for the allocator, a Cython function always raises
+# exceptions back to Python, but the allocator here is executed as a callback
+# directly in C++, therefore exceptions have to be raised to the C++ caller
+# directly.
+cdef extern from *:
```
wence-, line 102: I think you can do this in cython by marking the function as `noexcept` (which doesn't mean "nothrow"!, https://cython.readthedocs.io/en/latest/src/userguide/nogil.html#marking-functions-as-able-to-run-without-the-gil), but this change is fine.

python/ucxx/ucxx/_lib/libucxx.pyx
```diff
@@ -95,9 +95,20 @@ def _get_host_buffer(uintptr_t recv_buffer_ptr):
...
+    """
+    #include <memory>
+    #include <ucxx/buffer.h>
+
+    std::shared_ptr<ucxx::Buffer> _rmm_am_allocator(size_t length) {
+        return std::make_shared<ucxx::RMMBuffer>(length);
+    }
+    """
+    void _rmm_am_allocator(int) except+*
```
wence-, line 111:
```suggestion
    void _rmm_am_allocator(size_t) except+*
```
?

python/ucxx/ucxx/_lib_async/tests/test_send_recv_two_workers.py
```diff
@@ -101,39 +106,39 @@ async def read():
...
                     await send(ep, frames)
                 else:
-                    await am_send(ep, frames)
-
-            print("CONFIRM RECEIPT")
+                    while True:
+                        try:
+                            await am_send(ep, frames)
+                        except ucxx.exceptions.UCXNoMemoryError as e:
+                            # Memory pressure due to client taking long to
```
wence-, line 135:
```suggestion
                            # Memory pressure due to client taking too long to
```

wence- (approved):
Tiny nits

ci/test_common.sh
pentschev: Yes, they haven't been adding too much value I think, plus I've found out that you can enable timestamps from GHA or download the raw log with timestamps, for timeouts that is much more useful, for instance [in this failed nightly test](https://github.com/rapidsai/ucxx/actions/runs/11762964521/job/32766417063):

```
Sun, 10 Nov 2024 07:35:10 GMT
python/ucxx/ucxx/_lib_async/tests/test_send_recv_multi.py::test_send_recv_numpy[|u1-8-1] PASSED
Sun, 10 Nov 2024 07:52:49 GMT
python/ucxx/ucxx/_lib_async/tests/test_send_recv_multi.py::test_send_recv_numpy[|u1-8-16] PASSED
Sun, 10 Nov 2024 07:52:49 GMT
Error: Process completed with exit code 124.
```

It's clear `python/ucxx/ucxx/_lib_async/tests/test_send_recv_multi.py::test_send_recv_numpy[|u1-8-1]` got stuck for 17 minutes until the timeout was raised.

We can always reenable if we find more use for it, but I haven't really looked at those numbers in months.

cpp/src/request_am.cpp
pentschev: Given this is only for eager messages (usually up to few KB), I think if we ever hit this here there are bigger problems to worry about. Nevertheless, checking doesn't hurt, done in https://github.com/rapidsai/ucxx/pull/315/commits/93de79bdf0b15ace94959d4a82e2a9b736f7b873 .

python/ucxx/ucxx/_lib/libucxx.pyx
pentschev: This is indeed much better, I could have sworn this existed but couldn't find it for some reason, it seems I don't know how to use Cython's docs. Reverted to the previous plus `noexcept`/`nogil` in https://github.com/rapidsai/ucxx/pull/315/commits/bf54be9367a7d91ffe15a14bd0059954f223f6fa .

python/ucxx/ucxx/_lib/libucxx.pyx
pentschev: Good catch, not sure why I wrote `int` here. Anyways, this was reverted and fixed in https://github.com/rapidsai/ucxx/pull/315/commits/bf54be9367a7d91ffe15a14bd0059954f223f6fa as well.

python/ucxx/ucxx/_lib_async/tests/test_send_recv_two_workers.py
pentschev: Done in https://github.com/rapidsai/ucxx/pull/315/commits/605cea0f61ba06702323606bfb7db6a1eae0910d .

jameslamb (approved):

pentschev:
Thanks @wence- and @jameslamb !

PR #317, opened 2024-11-11
Replace uses of `__del__` with `weakref.finalize`

### LINKED ISSUE #209: Consider using weakref.finalize rather than `__del__` to clean up endpoint
As discussed starting here https://github.com/rapidsai/ucxx/pull/206#discussion_r1524665102, since #206 we attempt to `abort` the endpoint when the python endpoint is collected (via a `__del__` callback). Since this is not the blessed way of implementing finalizers (that's `weakref.finalize`) we should consider moving to that instead. However, that comes with some potential complications if the Cython wrappers (which hold C++ objects) are participating in reference cycles. If the ref-cycle is broken in a cython object then it is possible that the deallocation that then occurs leaves dangling references.

pentschev:
Modern Python should use `weakref.finalize` instead of `__del__`. This change removes all legacy uses of `__del__` in favor of `weakref.finalize`.

Additionally rename the `id` attribute from `ActiveClients` as it's a reserved keyword in Python, thus it's best if we use another name such as `ident`.

Closes #209


python/ucxx/ucxx/_lib_async/endpoint.py
```diff
@@ -107,7 +124,7 @@ def abort(self, period=10**10, max_attempts=1):
             if worker is running a progress thread and `period > 0`.
         """
         if self._ep is not None:
-            logger.debug("Endpoint.abort(): 0x%x" % self.uid)
+            logger.debug(f"Endpoint.abort(): {hex(int(self.uid))}")
```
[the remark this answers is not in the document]
pentschev: It is an integer, that's a good idea. Applied it for both cases in https://github.com/rapidsai/ucxx/pull/317/commits/cbf39cd6350cfb03f5ca632a5822221ef17224a6 .

python/distributed-ucxx/distributed_ucxx/ucxx.py
```diff
@@ -375,11 +394,19 @@ def __init__(  # type: ignore[no-untyped-def]
...
 
-        weakref.finalize(self, _deregister_dask_resource, self._resource_id)
+        weakref.finalize(self, _finalizer, ep, resource_id)
+
+    def abort(self):
+        self._closed = True
+        if self._ep is not None:
+            self._ep.abort()
+            self._ep = None
+            _deregister_dask_resource(self._resource_id)
```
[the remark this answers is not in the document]
wence-: Thanks!

ci/test_common.sh
```diff
@@ -128,13 +128,13 @@ run_py_tests_async() {
   ENABLE_PYTHON_FUTURE=$3
   SKIP=$4
 
-  CMD_LINE="UCXPY_PROGRESS_MODE=${PROGRESS_MODE} UCXPY_ENABLE_DELAYED_SUBMISSION=${ENABLE_DELAYED_SUBMISSION} UCXPY_ENABLE_PYTHON_FUTURE=${ENABLE_PYTHON_FUTURE} timeout 20m python -m pytest -vs python/ucxx/ucxx/_lib_async/tests/ --runslow"
+  CMD_LINE="UCXPY_PROGRESS_MODE=${PROGRESS_MODE} UCXPY_ENABLE_DELAYED_SUBMISSION=${ENABLE_DELAYED_SUBMISSION} UCXPY_ENABLE_PYTHON_FUTURE=${ENABLE_PYTHON_FUTURE} timeout 30m python -m pytest -vs python/ucxx/ucxx/_lib_async/tests/ --runslow"
```
KyleFromNVIDIA, line 131: Why the change in timeout?

ci/test_common.sh
```diff
@@ -128,13 +128,13 @@ run_py_tests_async() {
...
 
   if [ $SKIP -ne 0 ]; then
     echo -e "\e[1;33mSkipping unstable test: ${CMD_LINE}\e[0m"
   else
     log_command "${CMD_LINE}"
-    UCXPY_PROGRESS_MODE=${PROGRESS_MODE} UCXPY_ENABLE_DELAYED_SUBMISSION=${ENABLE_DELAYED_SUBMISSION} UCXPY_ENABLE_PYTHON_FUTURE=${ENABLE_PYTHON_FUTURE} timeout 20m python -m pytest -vs python/ucxx/ucxx/_lib_async/tests/ --durations=50
+    UCXPY_PROGRESS_MODE=${PROGRESS_MODE} UCXPY_ENABLE_DELAYED_SUBMISSION=${ENABLE_DELAYED_SUBMISSION} UCXPY_ENABLE_PYTHON_FUTURE=${ENABLE_PYTHON_FUTURE} timeout 30m python -m pytest -vs python/ucxx/ucxx/_lib_async/tests/ --runslow
```
KyleFromNVIDIA, line 137: Why the change in timeout and the change from `--durations=50` to `--runslow`?

ci/test_common.sh
pentschev: I’ve seen 2-3 times in a row the same test hitting timeout, which may be either due to higher load on CI resources or just by the natural increase in number of tests pushing it closer to the once reasonable limit.

ci/test_common.sh
pentschev: Adding `--runslow` and removing `--durations=50` was intended to be done in https://github.com/rapidsai/ucxx/pull/315/files#diff-ad34f6e6a948b202cb081a192d0345f90a75dca1bffd8aeef9e3b6c11c4f31a5, unfortunately I only changed the CMD_LINE variable which is used for logging, but not the actual command line that gets executed. Durations has not been adding too much value, and `--runslow` is to run some of the tests that were previously disabled.

KyleFromNVIDIA (approved):

jakirkham:
A couple jobs had test failures. Am guessing these were sporadic. Have tried restarting them

pentschev:
Thanks @jakirkham for rerunning, I was actually thinking it wasn't sporadic and tried reproducing locally without success, which I still may need to do later. For now tests have passed and I'll go ahead and merge this.

Thanks @wence- and @KyleFromNVIDIA for reviews too!

jakirkham:
Ah ok. The logs should still be there from [the previous run]( https://github.com/rapidsai/ucxx/actions/runs/11825675428 ). The errors were

1. Cluster teardown error in `test_from_worker_address_error[unreachable]`
2. Uploading docs step failed

pentschev:
> * Cluster teardown error in `test_from_worker_address_error[unreachable]`

Yes, indeed this is what worried me, I wasn't able to reproduce it locally so I thought it was related to changes in this PR.

> * Uploading docs step failed

This was probably due to some changes in `shared-workflow` that temporarily broke GHA for us, should have been fixed already.

Given that there's not anything to be done as a follow-up to this PR, I'll keep an eye on CI as usual though.

pentschev:
The unreachable test should hopefully be fixed in https://github.com/rapidsai/ucxx/pull/324 .

PR #318, opened 2024-11-12
Fix finalizers when Python futures are enabled

pentschev:
The Python futures pool retains a reference to `_notifier`, which prevents further cleanup of the `ucxx::{python::,}Worker`. Thus a new `clearFuturesPool` method is introduced to allow clearing the pool so that these references are dropped, which is called before the Python `_notifierThread` exits.

Additionally, move `ApplicationContext.progress_tasks` to a global `ProgressTasks` dictionary. Having the progress tasks as a member attribute of `ApplicationContext` forces us to keep a reference to it that prevents from properly cleaning up in the `weakref.finalize`. Using an external object allows us to prevent the self-reference and clean it up properly when no further references to `ApplicationContext` exist or `ucxx.reset()` is called.

### DIFF
--- a/cpp/python/src/worker.cpp
+++ b/cpp/python/src/worker.cpp
@@ -71,6 +71,21 @@ void Worker::populateFuturesPool()
   }
 }
 
+void Worker::clearFuturesPool()
+{
+  if (_enableFuture) {
+    ucxx_trace_req("ucxx::python::Worker::%s, Worker: %p, populateFuturesPool: %p",
+                   __func__,
+                   this,
+                   shared_from_this().get());
+    std::lock_guard<std::mutex> lock(_futuresPoolMutex);
+    PyGILState_STATE state = PyGILState_Ensure();
+    decltype(_futuresPool) newFuturesPool;
+    std::swap(_futuresPool, newFuturesPool);
+    PyGILState_Release(state);
+  }
+}
+
 std::shared_ptr<::ucxx::Future> Worker::getFuture()
 {
   if (_enableFuture) {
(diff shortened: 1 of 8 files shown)


cpp/python/src/worker.cpp
```diff
@@ -71,6 +71,21 @@ void Worker::populateFuturesPool()
...
+void Worker::clearFuturesPool()
+{
+  if (_enableFuture) {
+    ucxx_trace_req("ucxx::python::Worker::%s, Worker: %p, populateFuturesPool: %p",
+                   __func__,
+                   this,
+                   shared_from_this().get());
+    std::lock_guard<std::mutex> lock(_futuresPoolMutex);
+    PyGILState_STATE state = PyGILState_Ensure();
+    decltype(_futuresPool) newFuturesPool;
```
wence-, line 83: *note*: I guess this also applies to `populateFuturesPool` as well.

This function takes two locks, so we need to think about lock ordering. Is there a scenario where a thread holding the GIL might be attempting to call `getFuture` (which will need to acquire `_futuresPoolMutex`), while a different thread not holding the GIL has called `clearFuturesPool` and therefore holds the lock => deadlock?

Or is it the case that only one python thread is allowed to talk to these worker functions?

wence- (approved):
I think this looks good with one question about lock ordering

cpp/python/src/worker.cpp
pentschev: Currently yes, only one Python thread is supported (aka, tested/used). However, this is a very much valid concern as we have almost everything in place for multi-thread Python. I've filed https://github.com/rapidsai/ucxx/issues/325 to track that and fix it in a follow-up PR.

pentschev:
Thanks Lawrence for this review as well. As mentioned above I've filed an issue to think better about the double lock case, it shouldn't be a problem for our current use cases though so I'll go ahead and merge this for now.

PR #320, opened 2024-11-13
CI testing for continue-on-error with telemetry

msarahan:


msarahan:
This served its purpose and can be removed

PR #322, opened 2024-11-14
prefer wheel-provided libucxx.so in load_library, use RTLD_LOCAL

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/118

Modifies `libucxx.load_library()` in the following ways:

* prefer wheel-provided `libucxx.so` to system installation
* expose environment variable `RAPIDS_LIBUCXX_PREFER_SYSTEM_LIBRARY` for switching that preference
* load `libucxx.so` with `RTLD_LOCAL`, to prevent adding symbols to the global namespace ([dlopen docs](https://linux.die.net/man/3/dlopen))


vyasr (approved):

python/libucxx/libucxx/load.py
```diff
@@ -28,27 +55,29 @@ def load_library():
...
+        # before packaging the wheel), look for a system installation.
+        libucxx_lib = _load_wheel_installation(soname)
+        if libucxx_lib is None:
+            libucxx_lib = _load_system_installation(soname)
 
     # The caller almost never needs to do anything with this library, but no
     # harm in offering the option since this object at least provides a handle
-    # to inspect where libucxx was loaded from.
+    # to inspect where the libucxx was loaded from.
     return libucxx_lib
```
wence-, line 83: Note that this logic breaks the case of editable installs for developers in (for example) the devcontainer setup.

In that scenario, we don't have a wheel-provided DSO, and we don't have a "system" provided one. Instead, the cython files in the ucxx package have the rpath to the locally built DSO baked in, so ld.so happily finds it. However, here, since we're just asking ld.so to resolve a DSO name, we fail.

python/libucxx/libucxx/load.py
```diff
@@ -28,27 +55,29 @@ def load_library():
         libucx.load_library()
         del libucx
 
-    # Dynamically load libucxx.so. Prefer a system library if one is present to
-    # avoid clobbering symbols that other packages might expect, but if no
-    # other library is present use the one in the wheel.
+    prefer_system_installation = (
+        os.getenv("RAPIDS_LIBUCXX_PREFER_SYSTEM_LIBRARY", "false").lower() != "false"
+    )
```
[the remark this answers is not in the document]
jameslamb: Sorry, I'm gonna need some context on what "uses with HPC-X containers" is, I'm not familiar with that. Can you describe (here or offline) what that is and how `libucxx.so` is getting installed and expected to be found there?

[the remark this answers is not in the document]
pentschev: As per discussion offline, this won't affect containers using HPC-X as they do not contain UCXX, so it should be good as is.

pentschev (approved):
LGTM, thanks @jameslamb .

python/libucxx/libucxx/load.py
```diff
@@ -28,27 +55,35 @@ def load_library():
...
+        except OSError:
+            # If none of the searches above succeed, just silently return None
+            # and rely on other mechanisms (like RPATHs on other DSOs) to
+            # help the loader find the library.
+            pass
 
     # The caller almost never needs to do anything with this library, but no
     # harm in offering the option since this object at least provides a handle
-    # to inspect where libucxx was loaded from.
+    # to inspect where the libucxx was loaded from.
```
jameslamb, line 88:
```suggestion
    # to inspect where libucxx was loaded from.
```

python/libucxx/libucxx/load.py
jameslamb: Thanks, just pushed a commit with the changes we discussed offline and seen in https://github.com/rapidsai/kvikio/pull/553. That should address that.

PR #323, opened 2024-11-15
Set PyNVML upper pin

pentschev:
The plan is for `pynvml` to become a metapackage in version 12, where it not anymore ship NVML bindings but instead install `nvidia-ml-py` to serve the official NVML bindings. To provide a better future support, setting an upper pin for `pynvml` is a safer choice.


jakirkham:
Thanks Peter! 🙏

Can you please add a pin here as well?

https://github.com/rapidsai/ucxx/blob/042261d41c554eb49958477a8a2178bc9c560aa6/conda/recipes/ucxx/meta.yaml#L243

pentschev:
Thanks @jakirkham and good catch! Updated in https://github.com/rapidsai/ucxx/pull/323/commits/9e6df8e56dc41062231d3e9b9959309c94c9a8db .

jakirkham (approved):
Thanks Peter! 🙏

PR #324, opened 2024-11-15
Fix `test_from_worker_address_error[unreachable]` test

pentschev:
The error raised may manifest in different ways depending on the ordering and timing actions occur, as well as transports used. The approach implemented in this PR seems to prevent timing issues by waiting for the remote process to terminate entirely, and not only depend on the `ucxx.reset()` call.

In a local environment the issue was reproducible in under 5 minutes rerunning the test, with this change it didn't reproduce after 2 hours of consecutive rerunning.


wence- (approved):

pentschev:
Thanks @wence- !

PR #337, opened 2024-11-27
Test against rmm

vyasr:


vyasr:
#343 uncovered that ucxx does not need the rmm logger at all, so this is a poor test case. I'll reconstruct this in another repo.

PR #338, opened 2024-11-27
Update cpplint version

vyasr:
This PR updates our pre-commit to use the latest version of cpplint. This brings in [my fix](https://github.com/cpplint/cpplint/pull/269) that ensures that the pre-commit hook actually works in all environments. Without that fix, depending on the version of Python and setuptools in the base environment the cpplint installation may fail.

### DIFF
--- a/cpp/src/remote_key.cpp
+++ b/cpp/src/remote_key.cpp
@@ -6,6 +6,7 @@
 #include <memory>
 #include <sstream>
 #include <string>
+#include <vector>
 
 #include <ucp/api/ucp.h>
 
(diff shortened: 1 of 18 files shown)


cpp/src/remote_key.cpp
```diff
@@ -84,7 +85,7 @@ std::shared_ptr<RemoteKey> createRemoteKeyFromSerialized(std::shared_ptr<Endpoin
 
 size_t RemoteKey::getSize() const { return _memorySize; }
 
-uint64_t RemoteKey::getBaseAddress() { return (uint64_t)_memoryBaseAddress; }
+uint64_t RemoteKey::getBaseAddress() { return static_cast<uint64_t>(_memoryBaseAddress); }
```
pentschev, line 88:
```suggestion
uint64_t RemoteKey::getBaseAddress() { return _memoryBaseAddress; }
```

pentschev: The typecast is redundant since [the attribute already is `uint64_t`](https://github.com/rapidsai/ucxx/blob/be0965b67390cdc78fc916ddf8ea4bf8bd8259d0/cpp/include/ucxx/remote_key.h#L37), we're safe to just remove it then.

pentschev (approved):
Pushing one change that was not related to you but that made me realize a redundancy in casting, otherwise this looks good. Thanks @vyasr !

KyleFromNVIDIA (approved):

PR #339, opened 2024-11-29
Move [[nodiscard]] to non-friend declaration

tongyuantongyu:
`[[nodiscard]]` attribute does not have effect on `friend` declaration, and Clang compiler rejects it:
```
ucxx/cpp/include/ucxx/context.h:75:3: error: an attribute list cannot appear here
   75 |   [[nodiscard]] friend std::shared_ptr<Context> createContext(ConfigMap ucxConfig,
      |   ^~~~~~~~~~~~~
```

It also triggers `-Wattributes` in GCC:
```
ucxx/cpp/include/ucxx/context.h:75:49: warning: attribute ignored [-Wattributes]
   75 |   [[nodiscard]] friend std::shared_ptr<Context> createContext(ConfigMap ucxConfig,
      |                                                 ^~~~~~~~~~~~~
ucxx/cpp/include/ucxx/context.h:75:49: note: an attribute that appertains to a friend declaration that is not a definition is ignored
```

Move the `[[nodiscard]]` attribute to the non-friend declaration.

### DIFF
--- a/cpp/include/ucxx/address.h
+++ b/cpp/include/ucxx/address.h
@@ -63,8 +63,7 @@ class Address : public Component {
    *
    * @returns The `shared_ptr<ucxx::Address>` object.
    */
-  [[nodiscard]] friend std::shared_ptr<Address> createAddressFromWorker(
-    std::shared_ptr<Worker> worker);
+  friend std::shared_ptr<Address> createAddressFromWorker(std::shared_ptr<Worker> worker);
 
   /**
    * @brief Constructor for `shared_ptr<ucxx::Address>` from string.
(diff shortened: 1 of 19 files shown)


wence-:
I wonder if we should add `-Wall -Wattributes -Werror` to the cmake cxxflags? (Probably here? https://github.com/rapidsai/ucxx/blob/be0965b67390cdc78fc916ddf8ea4bf8bd8259d0/cpp/CMakeLists.txt#L59)

pentschev:
> I wonder if we should add `-Wall -Wattributes -Werror` to the cmake cxxflags?

Yeah, I was thinking the same, probably best to do that in a follow-up PR though.

pentschev:
> > I wonder if we should add `-Wall -Wattributes -Werror` to the cmake cxxflags?
> 
> Yeah, I was thinking the same, probably best to do that in a follow-up PR though.

I started https://github.com/rapidsai/ucxx/pull/340 for that. There's a bunch of warnings and errors that will need to be fixed.

pentschev (approved):
LGTM, thank you @tongyuantongyu for catching this and submitting the fix!

PR #340, opened 2024-11-29
Enable various C++ build-time warnings and treat them as errors

pentschev:
Enable various C++ build-time warnings to improve code quality and treat them as errors moving forward. Additionally fix all of those entries that are now considered warnings.


cpp/CMakeLists.txt
```diff
@@ -56,7 +56,7 @@ set(UCXX_BUILD_TESTS ${BUILD_TESTS})
 set(UCXX_BUILD_BENCHMARKS ${BUILD_BENCHMARKS})
 set(UCXX_BUILD_EXAMPLES ${BUILD_EXAMPLES})
 
-set(UCXX_CXX_FLAGS "")
+set(UCXX_CXX_FLAGS "-Wall" "-Wattributes" "-Werror")
```
wence-, line 59: Other things that might be worthwhile adding `-Wextra`, `-Wsign-conversion`.

pentschev: Thanks Lawrence for the suggestions. I've extended those in https://github.com/rapidsai/ucxx/pull/340/commits/ba26e7e65e26324077346a3e3275c3c4368379a7 . Additionally I added `-Wno-missing-field-initializers` this is because we otherwise hit those warnings when initializing many of the UCP structures where we don't (and probably shouldn't, at least not in all cases) fill those values. Alternatively we could use `#pragma`s to filter them out, but I'm not sure if it's worth it. If you have other better ideas please let me know as well.

cpp/CMakeLists.txt
```diff
@@ -56,7 +56,7 @@ set(UCXX_BUILD_TESTS ${BUILD_TESTS})
 set(UCXX_BUILD_BENCHMARKS ${BUILD_BENCHMARKS})
 set(UCXX_BUILD_EXAMPLES ${BUILD_EXAMPLES})
 
-set(UCXX_CXX_FLAGS "")
+set(UCXX_CXX_FLAGS "-Wall" "-Wattributes" "-Werror" "-Wextra" "-Wsign-conversion" "-Wno-missing-field-initializers")
```
vyasr, line 59: You shouldn't need all the quotes here.

cpp/src/request_am.cpp
```diff
@@ -145,11 +145,11 @@ RequestAm::RequestAm(std::shared_ptr<Component> endpointOrWorker,
             callbackData)
 {
   std::visit(data::dispatch{
-               [this](data::AmSend amSend) {
+               [this](data::AmSend) {
```
vyasr, line 148: It seems like there are multiple cases where you dispatch to an empty functor, maybe that should be added as a method for simplification?

vyasr (approved):
LGTM.

cpp/CMakeLists.txt
pentschev: Thanks for the hint, removed in https://github.com/rapidsai/ucxx/pull/340/commits/f974f2c4a0ab4235c3140fc76fe3e110019f865f .

cpp/src/request_am.cpp
pentschev: Maybe I'm misunderstanding your suggestion, but the empty functors are not (necessarily) exactly the same, their signature changes with the argument type which is used to match the dispatcher, so I don't think this is possible.

cpp/src/worker.cpp
```diff
@@ -427,10 +427,7 @@ void Worker::clearFuturesPool() { THROW_FUTURE_NOT_IMPLEMENTED(); }
 
 std::shared_ptr<Future> Worker::getFuture() { THROW_FUTURE_NOT_IMPLEMENTED(); }
 
-RequestNotifierWaitState Worker::waitRequestNotifier(uint64_t periodNs)
-{
-  THROW_FUTURE_NOT_IMPLEMENTED();
-}
+RequestNotifierWaitState Worker::waitRequestNotifier(uint64_t) { THROW_FUTURE_NOT_IMPLEMENTED(); }
```
pentschev, line 430: Workaround for https://github.com/cpplint/cpplint/issues/131 .

vyasr (approved):

cpp/src/request_am.cpp
pentschev: You nailed it, that's exactly why we need it regardless of whether we use it or not, the type carries at a minimum the meaning of the kind of transfer the request will perform.

pentschev:
Thanks @wence- and @vyasr !

PR #343, opened 2024-12-17
Remove RMM logger CMake targets and use forward declaration

pentschev:
Remove RMM logger CMake targets that are unused by UCXX, and thus prevent unnecessary linking to spdlog/fmt. Use a forward declaration for `rmm::device_buffer` to avoid symbols being added to the symbol table.


vyasr (approved):

pentschev:
Thanks @vyasr !

wence- (approved):

PR #345, opened 2024-12-19
Use PyNVML 12

jakirkham:
Bump `pynvml` from `11` to `12`. This version of `pynvml` also now depends on `nvidia-ml-py` for core functionality.


jakirkham:
Given the previous commit passed and the last one is simply merging in the latest from `branch-25.02`, I don't think it is related

jakirkham:
@pentschev does the error above in James' comment look familiar to you?

pentschev:
> parser.c:2305 UCX WARN (set UCX_WARN_UNUSED_ENV_VARS=n to suppress this warning)
> parser.c:1339 UCX ERROR Invalid value for SEG_SIZE: 'invalid-size'. Expected: memory units: [b|kb|mb|gb], "inf", or "auto"

Those are normal when we test invalid arguments, they are errors that we expect to catch.

> Fatal Python error: Segmentation fault
> does the error above in James' comment look familiar to you?

That is unfortunately known, I thought it had been fixed a while ago, had not seen this one in particular for weeks, it seems there's still something flaky. 😞 -- Nevertheless, it should not be related, so triggered a rerun.

jameslamb:
Alright thanks @pentschev , I'll approve this then and we can merge if CI passes.

jameslamb (approved):

pentschev (approved):
LGTM, thanks @jakirkham .

jakirkham:
Thanks Peter and James! 🙏

Looks like the rerun worked 🎉

Let's keep an eye on it and we can follow up as needed

PR #348, opened 2025-01-15
Improve `tagProbe` and AM receiver callback

pentschev:
Improve `tagProbe` by accepting a tag mask for matching and return probed tag information. Expose also the sender endpoint handle to AM receive callback so that the callback is capable of knowing the origin of the message.

Additionally, fix C++ request tests that were being unintentionally skipped.

### DIFF
--- a/cpp/include/ucxx/typedefs.h
+++ b/cpp/include/ucxx/typedefs.h
@@ -65,20 +65,27 @@ enum Tag : ucp_tag_t {};
  */
 enum TagMask : ucp_tag_t {};
 
-struct TagRecvInfo {
-  Tag senderTag;  ///< Sender tag
-  size_t length;  ///< The size of the received data
-
-  TagRecvInfo() = delete;
-  explicit TagRecvInfo(const ucp_tag_recv_info_t&);
-};
-
 /**
  * @brief A full UCP tag mask.
  *
  * A convenience constant providing a full UCP tag mask (all bits set).
  */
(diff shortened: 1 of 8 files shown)


cpp/include/ucxx/typedefs.h
```diff
@@ -72,6 +72,21 @@ enum TagMask : ucp_tag_t {};
...
+ *
+ * Contains information returned when probing by a tag message received by the worker but
+ * not yet consumed.
+ */
+class TagRecvInfo {
+ public:
+  Tag senderTag;  ///< Sender tag
+  size_t length;  ///< The size of the received data
+
+  TagRecvInfo() = delete;
```
madsbk, line 86:
```suggestion
```

madsbk: I think this is needed since you define your own ctor

python/ucxx/ucxx/_lib/ucxx_api.pxd
```diff
@@ -174,10 +178,13 @@ cdef extern from "<ucxx/api.h>" namespace "ucxx" nogil:
         pass
     cdef enum TagMask:
         pass
+    cdef cppclass TagRecvInfo:
+        TagRecvInfo(const ucp_tag_recv_info_t&)
+
```
madsbk, line 183:
```suggestion
```

madsbk (approved):
LGTM, I only have some minor nitpick.

pentschev:
Thanks for the review @madsbk !

PR #349, opened 2025-01-16
Retain copy of header for AM send requests

pentschev:
Retain a copy of headers for AM send requests as workaround for possible UCX bug https://github.com/openucx/ucx/issues/10424 .

Unfortunately, reproducing this is not straightforward and it wasn't observed in a stack that can be made into UCXX tests currently, so testing this is not possible at the moment.


madsbk (approved):
Looks good

pentschev:
Thanks @madsbk !

PR #355, opened 2025-01-24
Remove `std::mutex()` from list-initialization

pentschev:
Remove the `std::mutex()` constructor call from list-initialization in `ucxx::Endpoint`. This is totally unnecessary and may cause issues with some compilers (or compiler options):

```
...
error: function "std::mutex::mutex(const std::mutex &)" (declared at line 94 of /opt/conda/envs/base/lib/gcc/x86_64-conda-linux-gnu/11.4.0/include/c++/bits/std_mutex.h) cannot be referenced -- it is a deleted function
```


seberg (approved):
Thanks, not that I can approve: This fixed the compile issue where I ran into it (legate-raft under a specific setup).

madsbk (approved):

PR #363, opened 2025-01-31
disallow fallback to Make, add devcontainers CI

### LINKED ISSUE #262: Adding devcontainers for UCXX
Raising this issue to track adding devcontainers for UCXX

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/146

Closes #262

Proposes:

* setting `[tool.scikit-build].ninja.make-fallback = false`, so `scikit-build-core` will not silently fallback to using GNU Make if `ninja` is not available
* adding `devcontainers` CI jobs here

### Why add `devcontainers` CI jobs in this PR?

The `devcontainers` are useful for interactive development on this project (including for outside contributors).

`ucxx` building successfully in them is also important for other all-of-RAPIDS builds, like those from https://github.com/rapidsai/devcontainers and other builds based on that repo. Having CI jobs run in PRs (as most of RAPIDs does) allows us to catch issues right here in `ucxx` when they happen, instead of only finding out when `ucxxx` is built as part of a larger devcontainers-based build.

### How did you set up the `.devcontainers` folder?

Copied https://github.com/rapidsai/cugraph/tree/branch-25.04/.devcontainer and changed all the `cugraph` references to `ucxx`.


wence-:
> setting [tool.scikit-build].ninja.make-fallback = false, so scikit-build-core will not silently fallback to using GNU Make if ninja is not available

Why is falling back problematic?

.devcontainer/cuda11.8-conda/devcontainer.json
```diff
@@ -0,0 +1,42 @@
+{
+  "build": {
+    "context": "${localWorkspaceFolder}/.devcontainer",
+    "dockerfile": "${localWorkspaceFolder}/.devcontainer/Dockerfile",
+    "args": {
+      "CUDA": "11.8",
+      "PYTHON_PACKAGE_MANAGER": "conda",
+      "BASE": "rapidsai/devcontainers:25.04-cpp-cuda11.8-mambaforge-ubuntu22.04"
```
bdice, line 8: We'll need to copy in all the `ci/release/update-version.sh` logic that we use to support this. It'll need to be different for ucxx than most libraries due to the version scheme differences. Note that both full (`25.04`) and PEP 440 (`25.4`) versions are needed in this file.

.devcontainer/cuda11.8-pip/devcontainer.json
```diff
@@ -0,0 +1,49 @@
...
+  "runArgs": [
+    "--rm",
+    "--name",
+    "${localEnv:USER:anon}-rapids-${localWorkspaceFolderBasename}-25.04-cuda11.8-pip"
+  ],
+  "hostRequirements": {"gpu": "optional"},
+  "features": {
+    "ghcr.io/rapidsai/devcontainers/features/cuda:25.4": {
+      "version": "11.8",
+      "installcuBLAS": true,
```
bdice, line 20: Do we need these math libraries for UCXX? I don't think we do.

cuDF doesn't have them: https://github.com/rapidsai/cudf/blob/branch-25.04/.devcontainer/cuda11.8-pip/devcontainer.json

Same for CUDA 12.8 pip devcontainers.

.devcontainer/cuda11.8-conda/devcontainer.json
jameslamb: 🤦🏻 you're right of course, sorry for not being more careful about this.

Just pushed https://github.com/rapidsai/ucxx/pull/363/commits/7980f3c0600c2d1544e3f5ccdf683f840e80b024 with fixes.

Tested like this:

```shell
./ci/release/update-version.sh '0.44.00'
git grep -E '25\.2|25\.02|25\.4|25\.04|0\.42|0\.43'
```

.devcontainer/cuda11.8-pip/devcontainer.json
jameslamb: You're totally right, those are unnecessary here. Removed in https://github.com/rapidsai/ucxx/pull/363/commits/7980f3c0600c2d1544e3f5ccdf683f840e80b024

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+FROM ${BASE} as pip-base
+
+ENV DEFAULT_VIRTUAL_ENV=rapids
+ENV RAPIDS_LIBUCX_PREFER_SYSTEM_LIBRARY=true
+
+RUN apt update -y \
+ && DEBIAN_FRONTEND=noninteractive apt install -y \
+    libblas-dev liblapack-dev \
+ && rm -rf /tmp/* /var/tmp/* /var/cache/apt/* /var/lib/apt/lists/*;
+
```
bdice, line 15: This can probably be removed too.
```suggestion
```

RMM is a more minimal example: https://github.com/rapidsai/rmm/blob/branch-25.04/.devcontainer/Dockerfile

ci/release/update-version.sh
```diff
@@ -75,3 +75,10 @@ for FILE in .github/workflows/*.yaml; do
 done
 
 sed_runner "s/--rapids-version=[[:digit:]]\{2\}.[[:digit:]]\{2\}/--rapids-version=${NEXT_RAPIDS_SHORT_TAG}/g" .pre-commit-config.yaml
+
+# .devcontainer files
+find .devcontainer/ -type f -name devcontainer.json -print0 | while IFS= read -r -d '' filename; do
+    sed_runner "s@rapidsai/devcontainers:[0-9.]*@rapidsai/devcontainers:${NEXT_RAPIDS_SHORT_TAG}@g" "${filename}"
+    sed_runner "s@rapidsai/devcontainers/features/rapids-build-utils:[0-9.]*@rapidsai/devcontainers/features/rapids-build-utils:${NEXT_RAPIDS_SHORT_TAG_PEP4400}@" "${filename}"
```
bdice, line 82:
```suggestion
    sed_runner "s@rapidsai/devcontainers/features/rapids-build-utils:[0-9.]*@rapidsai/devcontainers/features/rapids-build-utils:${NEXT_RAPIDS_SHORT_TAG_PEP440}@" "${filename}"
```

bdice (changes requested):

.devcontainer/Dockerfile
jameslamb: Yep agree!

After this change, the only difference between this dockerfile and that `rmm` one would be:

```dockerfile
ENV RAPIDS_LIBUCX_PREFER_SYSTEM_LIBRARY=true
```

And that really is needed. It was added across devcontainers for projects relying on `libucx` wheels in https://github.com/rapidsai/build-planning/issues/118

ci/release/update-version.sh
jameslamb: blegh, this mistake shows that the way I tested in https://github.com/rapidsai/ucxx/pull/363#discussion_r1939570571 wasn't great. There I was checking for "nothing was missed", should have also manually scanned the diff for "all the updates look right".

Fixed in https://github.com/rapidsai/ucxx/pull/363/commits/3c774c2314d6fee7d3b68440af17b838b439d989

Sorry for making you catch these things in review, I know it's annoying :/

.devcontainer/Dockerfile
jameslamb: fixed in https://github.com/rapidsai/ucxx/pull/363/commits/3c774c2314d6fee7d3b68440af17b838b439d989

bdice (approved):

ci/release/update-version.sh
bdice: No worries! Thanks for your work on this PR. :)

PR #364, opened 2025-02-03
Use new rapids-logger library

vyasr:
ucxx only needs minor changes here to adapt for the upstream rmm changes.

Contributes to https://github.com/rapidsai/build-planning/issues/104.


python/libucxx/libucxx/load.py
```diff
@@ -44,6 +44,11 @@ def _load_wheel_installation(soname: str):
 
 def load_library():
     """Dynamically load libucxx.so and its dependencies"""
+    try:
+        import librmm  # noqa: F401
```
pentschev, line 48: Just to be safe, this won't create a CUDA context upon import, will it?

pentschev (commented):
Overall looks good, just want to confirm `import rmm` won't create a CUDA context or any other resources when RMM is installed (but potentially remain unused dependending on the application needs).

python/libucxx/libucxx/load.py
vyasr: No, it won't. At the moment really all this is doing is making the underlying rapids-logger library available.

pentschev: Makes sense, I just want to ensure this remains usable for non-CUDA users if they have RMM installed but don't intend on using it.

pentschev (approved):

bdice (approved):

python/libucxx/libucxx/load.py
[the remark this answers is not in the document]
vyasr: Done in https://github.com/rapidsai/ucxx/pull/364/commits/853984de3d8d5356129a3c2068b0d616baf666d9.

vyasr:
I needed https://github.com/rapidsai/ucxx/pull/364/commits/806692a0652e499785e773a7f9bf1f7dbd8c3695 to unblock the devcontainer build. I am not currently sure whether our environment was previously underspecified or if something has changed in conda-forge CUDA packages to break a case that we intended to support. I will follow up, but I want to get this PR in to unblock the downstream chain so I'm to merge it with the additional dependency.

PR #370, opened 2025-02-10
Enable label checker ops-bot plugin

AyodeAwe:
Activates the label-checker function described here: https://docs.rapids.ai/resources/label-checker/

### DIFF
--- a/.github/ops-bot.yaml
+++ b/.github/ops-bot.yaml
@@ -3,7 +3,7 @@
 
 auto_merger: true
 branch_checker: false
-label_checker: false
+label_checker: true
 release_drafter: false
 recently_updated: true
 forward_merger: true


pentschev:
@AyodeAwe could we also add the "default label list" for this repo as well then? E.g., things like levels such as "3 - Ready for review" are missing.

AyodeAwe:
Looking at https://docs.rapids.ai/resources/label-checker/, I think the current list is sufficient for the label checker's function. Thoughts?

raydouglass (approved):

pentschev:
> Looking at https://docs.rapids.ai/resources/label-checker/, I think the current list is sufficient for the label checker's function. Thoughts?

Oh, so those are non-default. I'm fine omitting them, just thought we had a more extensive list that was default and included those.

PR #372, opened 2025-02-11
Expose class structure in `libucxx.pxd`

pentschev:
Expose class structure in `libucxx.pxd`. This change is useful so that projects building on top of the Python interface can take `shared_ptr<ucxx::Component>` objects and plug those objects into their C++ backends.

Although it's not possible to define private attributes in a `cdef class`, we attempt to mark them private with the leading underscore in their names, such as `UCXContext._context`, and then provide a `UCXContext.get_ucxx_shared_ptr` for those members that are expected to be externally usable.

Testing will be handled as part of #376.


python/ucxx/ucxx/_lib/libucxx.pxd
```diff
@@ -0,0 +1,103 @@
...
+
+
+cdef class UCXAddress:
+    cdef:
+        shared_ptr[Address] _address
+        size_t _length
+        ucp_address_t *_handle
+        string _string
+
+    cdef shared_ptr[Address] get_ucxx_shared_ptr(self) nogil
```
wence-, line 47: suggestion/question: What does this `get_ucxx_shared_ptr` setup buy us?

Is it reasonable to instead write:
```
cdef class UCXFoo:
    cdef:
        shared_ptr[Foo] c_obj
```
And just access `c_obj` directly?

test_getters.pyx
```diff
@@ -0,0 +1,39 @@
...
+    context = ucx_api.UCXContext(feature_flags=tuple(feature_flags))
+
+    get_context(context)
+
+    # test_shared(<UCXContext>context.get_ucxx_shared_ptr())
+
+    # ucxx_shared_ptr = context.get_ucxx_shared_ptr()
+
+    # assert ucxx_shared_ptr.get() != NULL
+    # assert <uint64_t>ucxx_shared_ptr.get() == context.ucxx_ptr
```
wence-, line 39: WIP: commented tests?

wence- (approved):
I think this looks fine with one question/suggestion. Though I realise you're just mostly moving code around.

python/ucxx/ucxx/_lib/libucxx.pxd
wence-: Ok, makes sense.

pentschev:
Getting tests done here is getting too complicated and I'd like to make the changes available for projects that currently need those changes. Due to that, I've removed the tests and infrastructure to #376 and will try to get this merged as soon as tests pass.

pentschev:
Thanks all, I'll continue pushing on #376 for the remaining changes.

PR #374, opened 2025-02-13
Switch from `conda-build` to `rattler-build`

gforsyth:
Porting over the conda-builds to `rattler-build`


conda/recipes/ucxx/recipe.yaml
```diff
@@ -0,0 +1,229 @@
...
+        SCCACHE_S3_NO_CREDENTIALS: ${{ env.get("SCCACHE_S3_NO_CREDENTIALS", default="") }}
+        SCCACHE_S3_KEY_PREFIX: libucxx-${{ env.get("RAPIDS_CONDA_ARCH", default="") }}
+
+outputs:
+  - package:
+      name: libucxx
+      version: ${{ version }}
+    build:
+      string: cuda${{ cuda_major }}_${{ date_string }}_${{ head_rev }}
+      script: install_libucxx.sh
```
gforsyth, line 64: There are comments in this build-script that I'm not sure how to preserve in the `script` directive here, so I've opted to leave that standalone script in place.

conda/recipes/ucxx/recipe.yaml
```diff
@@ -0,0 +1,244 @@
...
+      - ninja
+      - ${{ compiler("c") }}
+      - ${{ compiler("cxx") }}
+      - ${{ compiler("cuda") }}
+      - cuda-version =${{ cuda_version }}
+      - ${{ stdlib("c") }}
+    host:
+      - cuda-version =${{ cuda_version }}
+      - if: cuda_major == "11"
+        then: cuda-cudart-dev =${{ cuda_version }}
```
gforsyth, line 31: This fixed this locally for me with cuda 11 -- not sure if `cuda-cudart-dev` is constrained in some weird way version-wise, or if `rattler-build` is doing something unexpected re: version solving

conda/recipes/ucxx/recipe.yaml
```diff
@@ -0,0 +1,247 @@
...
+      string: ${{ date_string }}_${{ head_rev }}
+      script: cmake --install cpp/build --component examples
+    requirements:
+      build:
+        - cmake ${{ cmake_version }}
+      run:
+        - ${{ pin_subpackage("libucxx", exact=True) }}
+      ignore_run_exports:
+        by_name:
+          - cuda-version
```
gforsyth, line 130: Without this explicit `cuda-version` ignore, the `libucxx-examples` output was getting a runtime dependency of `cuda-version>12.8,<13` -- inherited from the `compiler('cuda')`:

```
 │ │ Run dependencies:
 │ │ ╭──────────────────┬────────────────────────────────────────────────────────╮
 │ │ │ Name             ┆ Spec                                                   │
 │ │ ╞══════════════════╪════════════════════════════════════════════════════════╡
 │ │ │ Run dependencies ┆                                                        │
 │ │ │ libucxx          ┆ ==0.43.0a37 cuda12_250218_2a6940ca (PS)                │
 │ │ │ libgcc           ┆ >=13 (RE of [cache-build: gcc_linux-64])               │
 │ │ │ __glibc          ┆ >=2.28,<3.0.a0 (RE of [cache-build: sysroot_linux-64]) │
 │ │ │ cuda-version     ┆ >=12.8,<13 (RE of [cache-build: cuda-nvcc_linux-64])   │    <---- here
 │ │ │ libstdcxx        ┆ >=13 (RE of [cache-build: gxx_linux-64])               │
 │ │ │ libgcc           ┆ >=13 (RE of [cache-build: gxx_linux-64])               │
 │ │ │ ucx              ┆ >=1.15.0,<1.16.0a0 (RE of [cache-host: ucx])           │
 │ │ │ python_abi       ┆ 3.12.* *_cp312 (RE of [cache-host: python])            │
 │ │ │ librmm           ┆ >=25.4.0a26,<25.5.0a0 (RE of [cache-host: librmm])     │
 │ │ ╰──────────────────┴────────────────────────────────────────────────────────╯
```

this meant that in the `test_cpp` job running on cuda 12.0, the just-built package for `examples` was invalid, leading to the failures.

This looks to me like outputs in `rattler-build` are inheriting _runtime_ dependencies from the build cache, which is _usually_ fine but in this case is not what we want.

ci/build_cpp.sh
```diff
@@ -19,9 +19,24 @@ rapids-logger "Begin C++ and Python builds"
...
+                    --experimental \
+                    --no-build-id \
+                    --channel-priority disabled \
+                    --output-dir "$RAPIDS_CONDA_BLD_OUTPUT_DIR" \
+                    "${RATTLER_CHANNELS[@]}"
 
 sccache --show-adv-stats
 
+# remove build_cache directory
+rm -rf "$RAPIDS_CONDA_BLD_OUTPUT_DIR"/build_cache
```
msarahan, line 40: This seems like something that should be an ignore pattern in our upload script instead of something that every build script has to include.

msarahan (approved):

ci/build_cpp.sh
gforsyth: Agreed. We opened an issue upstream, too, to add a CLI option to clear the build-cache post-build: https://github.com/prefix-dev/rattler-build/issues/1424

conda/recipes/ucxx/recipe.yaml
bdice: (^everything Vyas said, and...)

CUDA 12 builds should not need the `nvidia` channel for anything. The only CUDA 11 packages we need from the `nvidia` channel are math libraries, so that we can get the C++ headers that don't ship with `conda-forge::cudatoolkit` for CUDA 11. I think that only affects RAFT/cuML/cuGraph/cuVS.

We are currently only including the `nvidia` channel for consistency with CUDA 11, and will drop it from our build commands once CUDA 11 is no longer supported. Perhaps we should even consider dropping `nvidia` now, for the packages that don't need its math library headers.

gforsyth: What would we prefer?

- drop `-c nvidia` from the `rapids-rattler-channel-string` and then add it piecemeal to raft, cuml, cugraph, and cuvs?
- wait until we drop cuda 11

conda/recipes/ucxx/recipe.yaml
```diff
@@ -0,0 +1,263 @@
...
+    about:
+      homepage: ${{ load_from_file("python/ucxx/pyproject.toml").project.urls.Homepage }}
+      license: ${{ load_from_file("python/ucxx/pyproject.toml").project.license.text }}
+      summary: ${{ load_from_file("python/ucxx/pyproject.toml").project.description }}
+
+  - package:
+      name: distributed-ucxx
+      version: ${{ version }}
+    build:
+      string: py${{ python }}_${{ date_string }}_${{ head_rev }}
```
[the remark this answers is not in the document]
gforsyth: building it as `noarch` now and that seems to be working

conda/recipes/ucxx/recipe.yaml
bdice: I think `rapids-rattler-channel-string` should check the value of `RAPIDS_CUDA_VERSION` and only add `--channel nvidia` on CUDA 11. We can implement that as part of the rattler-build migration, no need to wait until we drop CUDA 11.

conda/recipes/ucxx/recipe.yaml
bdice: We had a recent conversation about ARM / Tegra CUDA support in conda-forge. We may be adding new run-exports to the CUDA compiler packages on ARM.

Currently we use `ignore_run_exports_from` the CUDA compiler package. Instead, we should `ignore_run_exports:` on just `cuda-version`. So what you're saying here is right, and we should do this universally for all packages that support CUDA compatibility within the major version (that applies to all of RAPIDS).

conda/recipes/ucxx/recipe.yaml
bdice: The version of this package should be constrained by `cuda-version`. We should never pin CTK package versions in our recipes (for CUDA 12 only -- [CUDA 11 requires exact pinnings like these](https://github.com/rapidsai/raft/blob/0ba3cc00fe38dc10d757ed63089e00355a738d54/conda/recipes/libraft/conda_build_config.yaml#L22-L48)). CUDA 12 packages should always float to whatever is allowed by `cuda-version`.
```suggestion
        then: cuda-cudart-dev
```

conda/recipes/ucxx/recipe.yaml
```diff
@@ -0,0 +1,273 @@
...
+      ignore_run_exports:
+        from_package:
+          - ucx
+          - python
+          - cuda-cudart-dev
+        by_name:
+          - cuda-version
+    tests:
+      - package_contents:
+          files:
```
bdice, line 88: Can we alphabetically sort this whole list of files?

(Also we used to have lists of files to verify for cudf but we dropped it there because it was annoying to maintain. But I am okay with the status quo here.)

conda/recipes/ucxx/recipe.yaml
```diff
@@ -0,0 +1,273 @@
...
+        - cmake ${{ cmake_version }}
+      run:
+        - ${{ pin_subpackage("libucxx", exact=True) }}
+      ignore_run_exports:
+        from_package:
+          - cuda-cudart-dev
+          - python
+          - ucx
+        by_name:
+          - cuda-version
```
bdice, line 140: None of these packages are in the build/host sections, so these run-exports shouldn't exist in the first place. Can we delete this block?
```suggestion
```

conda/recipes/ucxx/recipe.yaml
```diff
@@ -0,0 +1,273 @@
...
+          else:
+            - cuda-cudart
+        - ${{ pin_compatible("cuda-version", upper_bound="x", lower_bound="x") }}
+      ignore_run_exports:
+        from_package:
+          - cuda-cudart-dev
+          - python
+          - ucx
+        by_name:
+          - cuda-version
```
bdice, line 170: Same as above. Nothing in build/host for this package has run-exports that we need to ignore.
```suggestion
```

conda/recipes/ucxx/recipe.yaml
gforsyth: `rattler-build` has different behavior to `condabuild` when there's a cache involved.  If I remove these lines, these are the run dependencies for `libucxx-examples` for a cuda 12 build:

```
 │ │ Run dependencies:
 │ │ ╭──────────────────┬─────────────────────────────────────────────────────────╮
 │ │ │ Name             ┆ Spec                                                    │
 │ │ ╞══════════════════╪═════════════════════════════════════════════════════════╡
 │ │ │ Run dependencies ┆                                                         │
 │ │ │ libucxx          ┆ ==0.43.0a48 cuda12_250220_bba22757 (PS)                 │
 │ │ │ libgcc           ┆ >=13 (RE of [cache-build: gcc_linux-64])                │
 │ │ │ __glibc          ┆ >=2.28,<3.0.a0 (RE of [cache-build: sysroot_linux-64])  │
 │ │ │ cuda-version     ┆ >=12.8,<13 (RE of [cache-build: cuda-nvcc_linux-64])    │
 │ │ │ libstdcxx        ┆ >=13 (RE of [cache-build: gxx_linux-64])                │
 │ │ │ libgcc           ┆ >=13 (RE of [cache-build: gxx_linux-64])                │
 │ │ │ ucx              ┆ >=1.15.0,<1.16.0a0 (RE of [cache-host: ucx])            │
 │ │ │ cuda-cudart      ┆ >=12.8.57,<13.0a0 (RE of [cache-host: cuda-cudart-dev]) │
 │ │ │ librmm           ┆ >=25.4.0a26,<25.5.0a0 (RE of [cache-host: librmm])      │
 │ │ │ python_abi       ┆ 3.12.* *_cp312 (RE of [cache-host: python])             │
 │ │ ╰──────────────────┴─────────────────────────────────────────────────────────╯
```

conda/recipes/ucxx/recipe.yaml
```diff
@@ -0,0 +1,323 @@
...
+    build:
+      string: ${{ date_string }}_${{ head_rev }}
+      script: cmake --install cpp/build --component examples
+      dynamic_linking:
+        overlinking_behavior: "error"
+    requirements:
+      build:
+        - ${{ compiler("c") }}
+        - ${{ compiler("cxx") }}
+        - ${{ compiler("cuda") }}
```
bdice, line 147: Looks like these examples are already built by the top-level script, so no compilers are needed here.
```suggestion
```

conda/recipes/ucxx/recipe.yaml
```diff
@@ -0,0 +1,323 @@
...
+        - cmake ${{ cmake_version }}
+      host:
+        - libucxx
```
bdice, line 150: Maybe pin the subpackage here?
```suggestion
        - ${{ pin_subpackage("libucxx", exact=True) }}
```

conda/recipes/ucxx/recipe.yaml
```diff
@@ -0,0 +1,323 @@
...
+    build:
+      string: cuda${{ cuda_major }}_${{ date_string }}_${{ head_rev }}
+      script: cmake --install cpp/build --component testing
+      dynamic_linking:
+        overlinking_behavior: "error"
+    requirements:
+      build:
+        - ${{ compiler("c") }}
+        - ${{ compiler("cxx") }}
+        - ${{ compiler("cuda") }}
```
bdice, line 179:
```suggestion
```

conda/recipes/ucxx/recipe.yaml
```diff
@@ -0,0 +1,323 @@
...
+        - cmake ${{ cmake_version }}
+      host:
+        - cuda-version =${{ cuda_version }}
+        - if: cuda_major == "11"
+          then: cudatoolkit
+          else: cuda-cudart-dev
+        - libucxx
```
bdice, line 186: Pin subpackage?
```suggestion
        - ${{ pin_subpackage("libucxx", exact=True) }}
```

conda/recipes/ucxx/recipe.yaml
```diff
@@ -0,0 +1,323 @@
...
+
+  - package:
+      name: ucxx
+      version: ${{ version }}
+    build:
+      string: cuda${{ cuda_major }}_py${{ py_version }}_${{ date_string }}_${{ head_rev }}
+      dynamic_linking:
+        overlinking_behavior: "error"
+        missing_dso_allowlist:
+          - "lib/libpython3.12.so*"
```
bdice, line 214: iirc, I think we expect this to link to Python. We shouldn't ignore this. This could be indicating an issue with our use of variant builds for each Python version?

conda/recipes/ucxx/recipe.yaml
```diff
@@ -0,0 +1,323 @@
...
+        - scikit-build-core >=0.10.0
+        - cython >=3.0.0
+        - ${{ pin_subpackage("libucxx", exact=True) }}
+        - ucx
+        - rapids-build-backend >=0.3.0,<0.4.0.dev0
+        - rmm ${{ rapids_version }}
+        - cuda-version =${{ cuda_version }}
+        - pynvml >=12.0.0,<13.0.0a0
+      run:
+        - python * *_cpython
```
bdice, line 256: I think this constraint (`*_cpython`) is unnecessary. Having `python` in `host` exports a `python_abi` constraint that I think covers this. I see we're ignoring that below but that might be a mistake.

conda/recipes/ucxx/recipe.yaml
```diff
@@ -0,0 +1,347 @@
...
+            - include/ucxx/request_tag.h
+            - include/ucxx/request_tag_multi.h
+            - include/ucxx/typedefs.h
+            - include/ucxx/utils/file_descriptor.h
+            - include/ucxx/utils/python.h
+            - include/ucxx/utils/sockaddr.h
+            - include/ucxx/utils/ucx.h
+            - include/ucxx/worker.h
+            - include/ucxx/worker_progress_thread.h
+            - lib/libucxx.so
```
pentschev, line 143: Is it possible to check that some files were _NOT_ included? If so I think it would be useful here to ensure `libucxx_python.so` is not in this package and only in the `ucxx` package.

conda/recipes/ucxx/recipe.yaml
```diff
@@ -0,0 +1,347 @@
...
+        # cache
+        set -x
+        export CFLAGS=$(echo $CFLAGS | sed -E 's@\-fdebug\-prefix\-map[^ ]*@@g')
+        export CXXFLAGS=$(echo $CXXFLAGS | sed -E 's@\-fdebug\-prefix\-map[^ ]*@@g')
+        set +x
+
+        ucxx_ROOT="$(realpath ./cpp/build)"
+        export ucxx_ROOT
+
+        ./build.sh -n -v libucxx libucxx_python libucxx_benchmarks libucxx_examples libucxx_tests --cmake-args=\"-DCMAKE_INSTALL_LIBDIR=lib\"
```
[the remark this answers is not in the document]
pentschev: Ok, let's leave it like that for now, if it becomes difficult to deal with in the future we can always split it into a separate file.

conda/recipes/ucxx/recipe.yaml
```diff
@@ -0,0 +1,349 @@
...
+      license: ${{ load_from_file("python/libucxx/pyproject.toml").project.license.text }}
+      summary: libucxx tests executables
+
+  - package:
+      name: ucxx
+      version: ${{ version }}
+    build:
+      string: cuda${{ cuda_major }}_py${{ py_version }}_${{ date_string }}_${{ head_rev }}
+      # dynamic_linking:
+      #   overlinking_behavior: "error"
```
pentschev, line 225: Can we remove this?

pentschev (approved):
Left another suggestion, but otherwise it looks good to me. I'm approving from the UCXX build perspective, I don't know much about all the conda details so it would be best to have @bdice or someone else from the build team to look into this too.

conda/recipes/ucxx/recipe.yaml
gforsyth: That's a TODO before this gets merged -- `libucxx_python.so` and others link against `libpython.so` but this is flagging the overlinking checks in `rattler` for reasons I haven't figured out yet.

pentschev: Ok, I'll switch my review to request change just to remind ourselves of that.

pentschev (changes requested):
Address https://github.com/rapidsai/ucxx/pull/374#discussion_r2033365748

gforsyth:
Ok, I think the `libpython.so` overlinking error is a false positive -- I downloaded the packages built in 5850b2d and things import fine.
Looking at the shared libraries for `libucxx_python.so`, I see

```sh
🐚 ldd -v miniforge3/envs/ucxx_test/lib/libucxx_python.so
	linux-vdso.so.1 (0x00007f9c07bf6000)
	libucxx.so => /home/gforsyth/miniforge3/envs/ucxx_test/lib/libucxx.so (0x00007f9c07b41000)
	libpython3.10.so.1.0 => /home/gforsyth/miniforge3/envs/ucxx_test/lib/libpython3.10.so.1.0 (0x00007f9c07600000)
	libstdc++.so.6 => /home/gforsyth/miniforge3/envs/ucxx_test/lib/libstdc++.so.6 (0x00007f9c0741e000)
	libgcc_s.so.1 => /home/gforsyth/miniforge3/envs/ucxx_test/lib/libgcc_s.so.1 (0x00007f9c07b22000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f9c07200000)
	libucp.so.0 => /home/gforsyth/miniforge3/envs/ucxx_test/lib/libucp.so.0 (0x00007f9c07a27000)
	libcudart.so.12 => /home/gforsyth/miniforge3/envs/ucxx_test/lib/libcudart.so.12 (0x00007f9c06e00000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f9c07a22000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f9c07a1d000)
	libutil.so.1 => /lib/x86_64-linux-gnu/libutil.so.1 (0x00007f9c07a18000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f9c07117000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f9c07bf8000)
	libuct.so.0 => /home/gforsyth/miniforge3/envs/ucxx_test/lib/./libuct.so.0 (0x00007f9c079d8000)
	libucs.so.0 => /home/gforsyth/miniforge3/envs/ucxx_test/lib/./libucs.so.0 (0x00007f9c06d9b000)
	libucm.so.0 => /home/gforsyth/miniforge3/envs/ucxx_test/lib/./libucm.so.0 (0x00007f9c079bc000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f9c079b7000)
```

pentschev (approved):
LGTM if we think overlinking errors are just a false positive. Thanks @gforsyth .

PR #375, opened 2025-02-14
Consolidate more Conda solves in CI

KyleFromNVIDIA:
Issue: https://github.com/rapidsai/build-planning/issues/22

### DIFF
--- a/ci/build_docs.sh
+++ b/ci/build_docs.sh
@@ -1,12 +1,14 @@
 #!/bin/bash
-# Copyright (c) 2023, NVIDIA CORPORATION.
+# Copyright (c) 2023-2025, NVIDIA CORPORATION.
 
 set -euo pipefail
+
+rapids-logger "Downloading artifacts from previous jobs"
+CPP_CHANNEL=$(rapids-download-conda-from-s3 cpp)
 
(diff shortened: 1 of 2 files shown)


ci/build_docs.sh
```diff
@@ -1,33 +1,30 @@
...
 
 set -euo pipefail
 
+rapids-logger "Downloading artifacts from previous jobs"
+CPP_CHANNEL=$(rapids-download-conda-from-s3 cpp)
+
 rapids-logger "Create test conda environment"
 . /opt/conda/etc/profile.d/conda.sh
 
-export UCXX_VERSION="$(head -1 ./VERSION)"
```
jameslamb, line 9: https://github.com/rapidsai/ucxx/blob/67f8f48daddc6474316496cb10755ef8391be7eb/cpp/doxygen/Doxyfile#L41

I think this needs to be exported in this job, for Doxygen.

jameslamb (approved):
This needs one change. But it's small and non-controversial, approving so you can trigger a merge after fixing that, without needing another review.

jakirkham:
Thanks Kyle and James! 🙏

PR #376, opened 2025-02-18
Test Cython code exposed externally

pentschev:
This is a follow-up for #372 that exposed certain parts of the Cython API for external usage. In this PR, proper tests are added, including the require infrastructure to build the tests and run in CI.


build.sh
```diff
@@ -18,13 +18,14 @@ ARGS=$*
...
-VALIDARGS="clean libucxx libucxx_python ucxx distributed_ucxx benchmarks tests examples -v -g -n -c --show_depr_warn -h"
-HELP="$0 [clean] [libucxx] [libucxx_python] [ucxx] [distributed_ucxx] [benchmarks] [tests] [examples] [-vcgnh] [--cmake-args=\\\"<args>\\\"]
+VALIDARGS="clean libucxx libucxx_python ucxx ucxx_tests distributed_ucxx benchmarks tests examples -v -g -n -c --show_depr_warn -h"
+HELP="$0 [clean] [libucxx] [libucxx_python] [ucxx] [ucxx_tests] [distributed_ucxx] [benchmarks] [tests] [examples] [-vcgnh] [--cmake-args=\\\"<args>\\\"]
    clean                         - remove all existing build artifacts and configuration (start
                                    over)
    libucxx                       - build the UCXX C++ module
    libucxx_python                - build the UCXX C++ Python support module
    ucxx                          - build the ucxx Python package
+   ucxx_tests                    - build the ucxx Cython tests
```
madsbk, line 28: Maybe change to `--cython-tests` ?

madsbk (approved):
LGTM

build.sh
pentschev: I'm here attempting to match the [name of the conda package](https://github.com/rapidsai/ucxx/pull/376/files#diff-34a51f09f049b5254ca75250d9eb1ff334e82cd485488a56ea29c43d4f2b5c07R254) we're creating to have installable tests. I agree this isn't the best name but there's so much mismatch currently that's hard to keep track of all that. On that note, I think it makes sense to rename those variables to include the prefix `libucxx_` and match also the [existing package names like libucxx-tests](https://github.com/rapidsai/ucxx/blob/fb624992ea9607ce2507f41807b71f631abffb5c/conda/recipes/ucxx/meta.yaml#L147), so I've done that in https://github.com/rapidsai/ucxx/pull/376/commits/b7c3d4202341b8a937882db57c1a53aa7806b56c but kept the name here to `ucxx_tests` in an attempt to make things more consistent for now.

madsbk: Sounds good

build.sh
```diff
@@ -238,10 +245,21 @@ if buildAll || hasArg ucxx; then
...
 
     cd ${REPODIR}/python/ucxx/
     SKBUILD_CMAKE_ARGS="-DCMAKE_PREFIX_PATH=${INSTALL_PREFIX};-DCMAKE_BUILD_TYPE=${BUILD_TYPE};${SKBUILD_EXTRA_CMAKE_ARGS}" \
-        python -m pip install --no-build-isolation --no-deps --config-settings rapidsai.disable-cuda=true .
+        python -m pip install \
+            --no-build-isolation \
+            --no-deps \
+            --config-settings rapidsai.disable-cuda=true \
+            --config-settings skbuild.install.components=ucxx \
+            --config-settings skbuild.install.components=examples \
```
KyleFromNVIDIA, line 260: This doesn't seem right. Does this append to a list, or just overwrite a string?

ci/test_common.sh
```diff
@@ -117,9 +117,17 @@ run_cpp_port_retry() {
...
-  CMD_LINE="timeout 4m python -m pytest -vs python/ucxx/ucxx/_lib/tests/"
+  RUN_CYTHON=$1
+
+  if [ $RUN_CYTHON -ne 0 ]; then
+    ARGS="--run-cython"
+  else
+    ARGS=""
+  fi
+
+  CMD_LINE="timeout 4m python -m pytest -vs python/ucxx/ucxx/_lib/tests/ ${ARGS}"
```
KyleFromNVIDIA, line 128: Since we're using Bash, use a list instead:

```suggestion
  if [ $RUN_CYTHON -ne 0 ]; then
    ARGS=("--run-cython")
  else
    ARGS=()
  fi

  CMD_LINE="timeout 4m python -m pytest -vs python/ucxx/ucxx/_lib/tests/ ${ARGS[@]}"
```

ci/test_common.sh
```diff
@@ -117,9 +117,17 @@ run_cpp_port_retry() {
...
   log_command "${CMD_LINE}"
-  timeout 4m python -m pytest -vs python/ucxx/ucxx/_lib/tests/
+  timeout 4m python -m pytest -vs python/ucxx/ucxx/_lib/tests/ ${ARGS}
```
KyleFromNVIDIA, line 130:
```suggestion
  timeout 4m python -m pytest -vs python/ucxx/ucxx/_lib/tests/ "${ARGS[@]}"
```

python/ucxx/CMakeLists.txt
```diff
@@ -33,6 +34,8 @@ else()
   set(ucxx-python_FOUND OFF)
 endif()
 
+message(ucxx-python_FOUND="${ucxx-python_FOUND}")
+
```
KyleFromNVIDIA, line 38: Is this intended to be here, or were you just using it for debugging?

KyleFromNVIDIA (changes requested):

build.sh
pentschev: What part doesn't seem right? From `pip install --help`:
>   -C, --config-settings <settings>
>                               Configuration settings to be passed to the PEP 517 build backend. Settings take the form KEY=VALUE. Use multiple --config-settings options
>                               to pass multiple keys to the backend.

So we should use multiple `--config-settings`. AFAICT, passing multiple `--config-settings` will build a list, which would be equivalent to:

```
[tool.scikit-build]
install.components = ["ucxx", "examples"]
```

If you know of a different way please let me know, I've attempted other things, like using a comma, but none worked except as above.

python/ucxx/CMakeLists.txt
pentschev: Good catch, it was used for debugging. Removed now via https://github.com/rapidsai/ucxx/pull/376/commits/1b88ffc07281fe1dcfcec77da1878ec056210f6c .

build.sh
KyleFromNVIDIA: OK, so that is how it's intended to work. Thanks, that answers my question.

PR #378, opened 2025-02-25
Use conda-build instead of conda-mambabuild

bdice:
This changes from `conda mambabuild` to `conda build`. Conda now uses the mamba solver so no performance regressions are expected.

This is a temporary change as we plan to migrate to `rattler-build` in the near future. However, this is needed sooner to drop `boa` and unblock Python 3.13 migrations.

xref: https://github.com/rapidsai/build-planning/issues/149

### DIFF
--- a/ci/build_cpp.sh
+++ b/ci/build_cpp.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# SPDX-FileCopyrightText: Copyright (c) 2022-2023, NVIDIA CORPORATION & AFFILIATES.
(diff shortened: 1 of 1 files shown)


jameslamb (approved):
Looks good! I'll note here (haven't in the other reviews), I've also been spot-checking that `sccache` is still working and yeah, looks fine.

```text
Compile requests                    186
Compile requests executed           186
Cache hits                          186
Cache hits (c [gcc])                 21
Cache hits (c++ [gcc])              165
```

([build link](https://github.com/rapidsai/ucxx/actions/runs/13530173792/job/37810403296?pr=378#step:9:4614))

PR #382, opened 2025-03-13
Make `tagProbe` non-blocking

pentschev:
Make `tagProbe` non-blocking, the old way it was implemented enforced progressing the worker, which is not something we should do given the whole UCXX API is asynchronous.


madsbk (approved):
Nice!

pentschev:
Thanks @madsbk !

PR #383, opened 2025-03-13
Fix update-version.sh

raydouglass:
Adds new dependency to `update-version.sh`

Skips CI because this script is not tested in CI


jameslamb (approved):

pentschev (approved):
Thanks @raydouglass !

PR #385, opened 2025-03-16
Set a timeout for event loop in async tests

pentschev:
The current way subprocesses using `run_until_complete()` timeout only allows the process to be immediately killed 
which doesn't provide any information about the stack, by using an async task with a timeout we allow the event loop to timeout first and print the stack and only then kill the process, which should help debugging issues especially in CI.

This change adds a new test utility function `compute_timeouts()` that uses the timeout from the `asyncio_timeout` fixture to compute both async and process timeouts, ensuring better traceability of errors when any part of the test fails.

Additionally the hardcoded timeouts that were defined for some tests relying on subprocesses are now replaced with `compute_timeouts()` and thus indirectly increasing those timeouts as a factor of the timeout defined in `asyncio_timeout`. This seems to resolve recent failures in `test_from_worker_address` that are seemingly facing timeout issues on A100 runner that use NVKS, the runtime for such tests on non-NVKS runners is usually smaller than 10s, but on NVKS nodes can slightly exceed 30s, causing the previous hardcoded timeout to hit before the test completes. It is not clear why only NVKS runners are affected.


madsbk (approved):
LGTM

pentschev:
Thanks @madsbk !

PR #387, opened 2025-03-17
Support IPv6 in `get_address()`

### LINKED ISSUE #386: Adding support for ipv6 in python binding
In python binding of ucxx, ipv6 address seem not supported: https://github.com/rapidsai/ucxx/blob/branch-0.43/python/ucxx/ucxx/utils.py#L157

Unfortunately I'm working on a network environment where ipv4 address has dried up. From the document of UCX, ipv6 address should be supported: https://docs.nvidia.com/doca/archive/doca-v1.3/ucx-programming-guide/index.html#ucp-listener-create

Is it possible to support ipv6 address in this ucxx python binding?

pentschev:
Uses `psutil` to retrieve IPv6 address of a network interface when the optional `use_ipv6=True` kwarg is set.

Note: testing is not included as it's not safely possible at the moment because it requires that the system itself has IPv6 support enabled, there's at least one interface with an IPv6 address and that the test itself can determine those conditions a priori.

Closes #386 .

### DIFF
--- a/python/ucxx/ucxx/utils.py
+++ b/python/ucxx/ucxx/utils.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2022-2023, NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES.
 # SPDX-License-Identifier: BSD-3-Clause
 
 import fcntl
@@ -126,7 +126,7 @@ def get_ucxpy_logger():
     return logger
 
 
-def get_address(ifname=None):
+def get_address(ifname=None, use_ipv6=False):
     """
     Get the address associated with a network interface.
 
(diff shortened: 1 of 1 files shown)


python/ucxx/ucxx/utils.py
```diff
@@ -150,16 +153,31 @@ def get_address(ifname=None):
...
-            )
+        if use_ipv6:
+            import psutil
+
+            addrs = psutil.net_if_addrs()
+            if ifname in addrs:
+                for addr in addrs[ifname]:
+                    if addr.family == socket.AF_INET6:
+                        return addr.address.split("%")[0]
+            return None
```
madsbk, line 170: Raise an exception here? Or add `None` to the docstring

madsbk (approved):
Looks good, only a one suggestion

python/ucxx/ucxx/utils.py
pentschev: Good catch, fixed in https://github.com/rapidsai/ucxx/pull/387/commits/463b5f94070fffd6ba41c3633a8b11243de2f667 .

madsbk (approved):

pentschev:
Thanks @madsbk !

PR #390, opened 2025-03-20
Remove `UCXAddress.string` in favor of `__bytes__`

pentschev:
Currently the Python core API has a `UCXAddress.string` property that returns `bytes`. This is a confusing match, so instead of having that property this now switches to supporting `__bytes__` that can use proper Python API to return the serializable `bytes` object by calling `bytes(ucx_address_object)`.

Also removes the `create_from_string` method, as there were two methods that are essentially redundant and might cause more confusion than good.

### DIFF
--- a/python/ucxx/ucxx/_lib/libucxx.pyx
+++ b/python/ucxx/ucxx/_lib/libucxx.pyx
@@ -463,6 +463,9 @@ cdef class UCXAddress():
     def __bytes__(self) -> bytes:
         return bytes(self._string)
 
+    def __str__(self) -> str:
+        return self._string.decode('utf-8')
+
     def __getbuffer__(self, Py_buffer *buffer, int flags) -> None:
         if bool(flags & PyBUF_WRITABLE):
             raise BufferError("Requested writable view on readonly data")
--- a/python/ucxx/ucxx/_lib/libucxx.pyx
+++ b/python/ucxx/ucxx/_lib/libucxx.pyx
@@ -460,8 +460,7 @@ cdef class UCXAddress():
     def length(self) -> int:
         return int(self._length)
 
-    @property
-    def string(self) -> bytes:
+    def __bytes__(self) -> bytes:
         return bytes(self._string)
 
     def __getbuffer__(self, Py_buffer *buffer, int flags) -> None:
@@ -489,10 +488,10 @@ cdef class UCXAddress():
         pass
 
     def __reduce__(self) -> tuple:
-        return (UCXAddress.create_from_buffer, (self.string,))
+        return (UCXAddress.create_from_buffer, (bytes(self),))
 
     def __hash__(self) -> int:
-        return hash(bytes(self.string))
+        return hash(bytes(self))
 
 
 cdef void _generic_callback(void *args) with gil:
--- a/python/ucxx/ucxx/_lib/libucxx.pyx
+++ b/python/ucxx/ucxx/_lib/libucxx.pyx
@@ -405,14 +405,9 @@ cdef class UCXAddress():
         return address
 
     @classmethod
-    def create_from_buffer(cls, bytes buffer) -> UCXAddress:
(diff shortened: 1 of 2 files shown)


madsbk (approved):
Looks good, but why not use Python string?

python/ucxx/ucxx/_lib/libucxx.pyx
```diff
@@ -460,8 +460,7 @@ cdef class UCXAddress():
     def length(self) -> int:
         return int(self._length)
 
-    @property
-    def string(self) -> bytes:
+    def __bytes__(self) -> bytes:
         return bytes(self._string)
```
jakirkham, line 464: We could add a conversion function for both `bytes` and `str`. That way users can get either

```suggestion
    def __bytes__(self) -> bytes:
        return bytes(self._string)

    def __str__(self) -> str:
        return self._string.decode('utf-8')
```

pentschev:
> Looks good, but why not use Python string?

We could, but we were already returning `bytes` before and we know that works well, additionally `bytes` (at least in my mind) conveys the idea that this is not a human-readable string a little bit better, but that may just be a thing of my head.

python/ucxx/ucxx/_lib/libucxx.pyx
pentschev: I was going to say that this would need us to duplicate constructors, but [we already do](https://github.com/rapidsai/ucxx/blob/0d29ad849578777d5be9f45c1913ff50e8bbeb04/python/ucxx/ucxx/_lib/libucxx.pyx#L407-L429), so we might as well do this. 😄

pentschev:
> > Looks good, but why not use Python string?
> 
> We could, but we were already returning `bytes` before and we know that works well, additionally `bytes` (at least in my mind) conveys the idea that this is not a human-readable string a little bit better, but that may just be a thing of my head.

This was also addressed by John's suggestion.

pentschev:
Actually I'll have to retract what I said to both of you @jakirkham @madsbk . Since the C++ string is not human-readable at all, I don't think it makes sense to even support it as a string here, we would need to decode it to some format and I think this only creates unnecessary confusion (i.e., "should I use string or buffer"), when we can provide a single interface instead. I've thus removed `__str__` and `create_from_string`. Please let me know if you feel differently.

pentschev:
Also, there are [no other uses of `create_from_string`](https://github.com/search?q=org%3Arapidsai%20create_from_string&type=code) other than in UCXX itself, so no further updates should be necessary.

madsbk (approved):
I think this is good (I also misunderstood the meaning of string, though it was a human readable address :=) )

python/ucxx/ucxx/_lib/libucxx.pyx
```diff
@@ -489,10 +482,10 @@ cdef class UCXAddress():
         pass
 
     def __reduce__(self) -> tuple:
-        return (UCXAddress.create_from_buffer, (self.string,))
+        return (UCXAddress.create_from_buffer, (bytes(self),))
 
     def __hash__(self) -> int:
-        return hash(bytes(self.string))
+        return hash(bytes(self))
```
jakirkham, line 488: Once we have `string_view` available in Cython ( https://github.com/cython/cython/pull/6652 ) ( 3.1 ), we may want to revisit this. Then we can write the code to leverage views (fewer copies)

jakirkham (commented):
Thanks Peter! 🙏

These are good simplifications and clarifications

Read over the context that you provided and the code. Think there are a couple smaller simplifications we can include (below)

With a future version of Cython, think we can simplify the logic here further. Though nothing to do on that point for now

jakirkham (approved):
LGTM. Thanks Peter! 🙏

Also thank you for taking a look at the other PR

pentschev:
Thanks @jakirkham and @madsbk !

jakirkham:
Thanks Peter! 🙏

PR #392, opened 2025-03-21
Zero-copy creation of `Address` objects

jakirkham:
When setting or getting the underlying string data used for an `Address` object, avoid extra copying. This is done by using `string_view` where possible in C++ & Cython and by using `bytes`-like objects in Python. This cuts down on the allocation and copying of memory for temporary strings and deferring them to the relevant point where an allocation and copy must happen for the owning object.

Users remain free to copy and persist the `Address`'s as a `std::string` or `bytes` object at any point. This merely tries to defer that decision to the user instead of making it for them


pentschev:
Thanks for submitting this John! Unsurprisingly the tests failed:

```
ImportError: /opt/conda/conda-bld/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pl/lib/python3.12/site-packages/ucxx/_lib/libucxx.cpython-312-x86_64-linux-gnu.so: undefined symbol: _ZN4ucxx23createAddressFromStringENSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE
```

That's because the Cython API has not been updated as well. Should we make all the necessary changes here, including bumping Cython to 3.1 and leave the PR ready for the time Cython 3.1 is out?

jakirkham:
Ah sometimes the Cython API will let these things slide as it is usually just checking that something will work with the underlying C++ API (as opposed to enforcing it)

We could certainly do that

For now would it make sense to use `const std::string&`? That would save us some copies between these API calls

cpp/src/address.cpp
```diff
@@ -38,21 +38,21 @@ std::shared_ptr<Address> createAddressFromWorker(std::shared_ptr<Worker> worker)
...
-std::shared_ptr<Address> createAddressFromString(std::string addressString)
+std::shared_ptr<Address> createAddressFromString(std::string_view addressString)
 {
-  ucp_address_t* address = reinterpret_cast<ucp_address_t*>(new char[addressString.length()]);
   size_t length          = addressString.length();
-  memcpy(address, addressString.c_str(), length);
+  ucp_address_t* address = reinterpret_cast<ucp_address_t*>(new char[length]);
+  memcpy(address, addressString.data(), length);
   return std::shared_ptr<Address>(new Address(nullptr, address, length));
 }
```
wence-, line 47: This is fine.

cpp/src/address.cpp
```diff
@@ -38,21 +38,21 @@ std::shared_ptr<Address> createAddressFromWorker(std::shared_ptr<Worker> worker)
...
 ucp_address_t* Address::getHandle() const { return _handle; }
 
 size_t Address::getLength() const { return _length; }
 
-std::string Address::getString() const
+std::string_view Address::getString() const
 {
-  return std::string{reinterpret_cast<char*>(_handle), _length};
+  return std::string_view{reinterpret_cast<const char*>(_handle), _length};
 }
```
wence-, line 56: NAK, this is horrible because now I have to remember to tie the lifetime of the returned object to the `Address`. _And_ the docstring is now wrong.

I cannot believe this is worth it.

wence- (dismissed):

cpp/src/address.cpp
pentschev: This is a good point, we can make it the user's responsibility to ensure the lifetime, [this is already the case for the UCX handles](https://github.com/rapidsai/ucxx/blob/4cd6354fc23e62f19b9715c04fd69dd1a1298997/cpp/include/ucxx/address.h#L83-L86). I don't have a strong objection either way, I'm a little more inclined towards Lawrence's point simply because that would avoid the potential to dereference an invalid pointer, whereas if attempting to use the address (that has become invalid because the worker is not available anymore) would simply result in an error from UCX/exception from UCXX.

@wence- @jakirkham do you have strong opinions?

pentschev:
> For now would it make sense to use `const std::string&`? That would save us some copies between these API calls

Yes, using `const std::string&` immediately sounds appropriate to me

cpp/src/address.cpp
jakirkham: For additional context, this is used in one place in Cython where we already have something that could be `std::string_view` compliant (a Python `bytes` object). Currently in the construction we...

1. Copy the `bytes` object into a `std::string`
2. Copy the `std::string` to the `createAddressFromString` function call (no `const std::string&` currently)
3. Allocate and copy the `std::string` to a `char[]`

https://github.com/rapidsai/ucxx/blob/4cd6354fc23e62f19b9715c04fd69dd1a1298997/python/ucxx/ucxx/_lib/libucxx.pyx#L407-L416

4. Note there is a `getString` call at the end doing one more copy

cpp/src/address.cpp
jakirkham: AFAICT there are only two uses of this method. Both involve the creation of a Cython `UCXAddress` object

1. From `bytes` as described above (currently 4 allocations & copies): https://github.com/rapidsai/ucxx/pull/392#discussion_r2034379410
2. From the Cython `UCXWorker` described below

In the Cython `UCXWorker` case, we take and store a copy of the `std::shared_ptr<Address>`. The `Address` in turn owns and keeps alive the `ucp_address_t*`/`char*` representation. Additionally we...

1. Allocate and copy `ucp_address_t*` to `std::string`
2. Copy this `std::string` into the Cython `UCXAddress`

https://github.com/rapidsai/ucxx/blob/4cd6354fc23e62f19b9715c04fd69dd1a1298997/python/ucxx/ucxx/_lib/libucxx.pyx#L395-L405

In both cases we are doing a fair bit of additional object creation and copying. The goal of this PR was to cutdown on that

<hr>

Though would acknowledge that

1. Many ways to achieve the same goal
2. Some amount of creation and copying may be needed

That said, think it would be good if we could reduce the object creation and copying. Ideally deferring copying until the final object is constructed

<hr>

Side note: `std::string` can be trivially constructed from a `std::string_view`

jakirkham: That said, if we would rather call this behavior `getStringView` that also seems reasonable

python/ucxx/ucxx/_lib/libucxx.pyx
```diff
@@ -581,8 +583,11 @@ cdef class UCXAddress():
     def __releasebuffer__(self, Py_buffer *buffer) -> None:
         pass
 
-    def __reduce__(self) -> tuple:
-        return (UCXAddress.create_from_buffer, (bytes(self),))
+    def __reduce_ex__(self, Py_ssize_t protocol) -> tuple:
+        if protocol >= 5:
+            return (UCXAddress.create_from_buffer, (PickleBuffer(self),))
+        else:
+            return (UCXAddress.create_from_buffer, (bytes(self),))
```
wence-: Why use a `PickleBuffer`? AFAICT it's basically just a `bytes`?

pentschev (approved):
I think this looks good now, thanks John!

jakirkham:
Thanks Peter and Lawrence! 🙏

Am planning to add a few tests for this as well. Will do that next week

cpp/src/address.cpp
jakirkham: Should add the docstring has been updated. So think this is addressed

Happy to discuss more as needed

jakirkham:
Have included a few more tests

Changes are done from my end

Happy to discuss anything else as needed

pentschev (approved):
Thanks John!

python/ucxx/ucxx/_lib/libucxx.pxd
```diff
@@ -42,7 +42,7 @@ cdef class UCXAddress:
         shared_ptr[Address] _address
         size_t _length
         ucp_address_t *_handle
-        string _string
+        string_view _string
```
jakirkham, line 45: Realized this could also just be a `string_view` since it is just viewing the memory owned by `_address`. The same memory that the pointer `_handle` refers to

This attribute is only internally used. So it just avoids some copies when working with internal data in class methods

pentschev (approved):
Still looking good, thanks John. We can leave this open until EOD in case there are further comments and hit merge if not.

pentschev:
I just realized this is in fact breaking, the change of `getString()`'s return type _is breaking_.

jakirkham:
Thanks Peter! 🙏

Ah thank you for catching that. Had done a pretty extensive search of the RAPIDS code base and only found this function used in UCXX and in that Cython code. So not expecting other impact. That said, agree that it is more accurate to say this is a breaking change

If we would rather have a deprecation cycle for good measure, am open to doing that. Just let me know

Edit: Also asking OPS if we can still have this marked as breaking in the changelog

jakirkham:
> If we would rather have a deprecation cycle for good measure, am open to doing that. Just let me know

Went ahead and put something together in case we want it in PR: https://github.com/rapidsai/ucxx/pull/622

Ok either way

pentschev:
> Thanks Peter! 🙏
> 
> Ah thank you for catching that. Had done a pretty extensive search of the RAPIDS code base and only found this function used in UCXX and in that Cython code. So not expecting other impact. That said, agree that it is more accurate to say this is a breaking change
> 
> If we would rather have a deprecation cycle for good measure, am open to doing that. Just let me know
> 
> Edit: Also asking OPS if we can still have this marked as breaking in the changelog

I only found that because [we use it in RapidsMPF as well](https://github.com/rapidsai/rapidsmpf/blob/6103323e9dc954095ccd63c0405bca5f0ad6f717/cpp/src/bootstrap/ucxx.cpp#L47-L50) and it was using `auto` but passing to a function that expected a `std::string`, which in the linked PR was adjusted to accept `std::string_view` now. Yesterday I pondered whether we needed some deprecation cycle and though we maybe don't, the amount of work will be relatively the same for users and we don't (yet) guarantee API/ABI stability.

jakirkham:
Ah sorry for the disruption there and not catching those previously. Thank you for following up on those issue and letting me know! 🙏

Will take a closer look at RapidsMPF going forward

PR #401, opened 2025-05-01
fix(rattler): produce multiple python outputs

gforsyth:
- **fix(rattler): iterate over variants in conda_build_config for python builds**
- **chore(conda-build): remove old config files and scripts**
- **fix(rattler): only build distributed_ucxx once**
- **fix(rattler): fix `rapids_version`**

`ucxx` build is a bit different than most of our builds, missed this in an early rattler port.  Now the `ucxx` package gets built for each python variant declared in the `conda_build_config`.

Also cleaning up some conda build stuff that should've been removed in the earlier PR.

xref #374

**edit** this ended up being a bit gnarlier than expected.  

Changes made:

- fix so that `libucxx_tests` doesn't have a hidden dependency on `libucxx_python` (when built via `build.sh`)
- stop building `libucxx_python` in the pre-build/cache stage and instead build it once-per-Python variant
- remove `python` from the cache `host` since we aren't building `libucxx_python` there
- pin `distributed_ucxx` to build against lowest supported Python (3.10) because it should run against any of our supported Python versions.


conda/recipes/ucxx/conda_build_config.yaml
```diff
@@ -34,4 +34,4 @@ gtest:
   - ">=1.13.0"
 
 rapids_version:
-  - "25.6.*"
+  - ">=25.6,<25.8"
```
raydouglass, line 37: Is this necessary? If so, https://github.com/rapidsai/ucxx/blob/branch-0.44/ci/release/update-version.sh#L44 needs to be updated, but the script doesn't track the next version currently, so it would need to calculate it.

Also, if so, I think it might need to be `<25.8.0a0`.

bdice: I think this change is not necessary.

conda/recipes/ucxx/recipe.yaml
```diff
@@ -328,7 +326,10 @@ outputs:
     requirements:
       host:
         - pip
-        - python
+        # Pin this to a specific Python because we only need one build of it,
+        # then ignore the `python_abi` run export so that it can be used with
+        # any version of `ucxx`
+        - python =3.12
```
bdice, line 332: Can we build with the oldest instead? I have more faith in forward compatibility if it ever becomes relevant.
```suggestion
        - python =3.10
```

conda/recipes/ucxx/conda_build_config.yaml
gforsyth: No, I got hung up on some channel priority conflicts but it was due to a local package I had sitting in my build output.  Reverted.

raydouglass:
Can confirm that I see the build listing each of the python variants now. For example, [here](https://github.com/rapidsai/ucxx/actions/runs/14779706506/job/41495849268?pr=401#step:9:541) & [here](https://github.com/rapidsai/ucxx/actions/runs/14779706506/job/41495849268?pr=401#step:9:5769).

bdice (approved):

PR #410, opened 2025-05-06
Add Python to distributed-ucxx Conda build string

jakirkham:
The Python version was dropped from the `distributed-ucxx` build string when migrating to `rattler-build`. Manually add this back.

### DIFF
--- a/conda/recipes/ucxx/recipe.yaml
+++ b/conda/recipes/ucxx/recipe.yaml
@@ -401,7 +401,7 @@ outputs:
       version: ${{ version }}
     build:
       noarch: python
-      string: ${{ date_string }}_${{ head_rev }}
+      string: py${{ python | version_to_buildstring }}_${{ date_string }}_${{ head_rev }}
       script: ./build.sh distributed_ucxx
     requirements:
       host:
--- a/conda/recipes/ucxx/recipe.yaml
+++ b/conda/recipes/ucxx/recipe.yaml
@@ -401,7 +401,7 @@ outputs:
       version: ${{ version }}
     build:
       noarch: python
-      string: py${{ python | version_to_buildstring }}_${{ date_string }}_${{ head_rev }}
+      string: py_${{ date_string }}_${{ head_rev }}
       script: ./build.sh distributed_ucxx
     requirements:
       host:


conda/recipes/ucxx/recipe.yaml
```diff
@@ -401,7 +401,7 @@ outputs:
       version: ${{ version }}
     build:
       noarch: python
-      string: ${{ date_string }}_${{ head_rev }}
+      string: py${{ python | version_to_buildstring }}_${{ date_string }}_${{ head_rev }}
```
bdice, line 404: @jakirkham This is `noarch: python` so presumably we don't want a Python version in the buildstring?

bdice (changes requested):

conda/recipes/ucxx/recipe.yaml
jakirkham: Good point!

Was basing this off of [previous builds]( https://anaconda.org/rapidsai-nightly/distributed-ucxx/files?sort=time&sort_order=asc ), which did have the Python version (for [example]( https://anaconda.org/rapidsai-nightly/distributed-ucxx/files/modal/info/67fca0e65729148dffdf803b ))

Looks like `noarch: python` was added as part of the migration to `rattler-build` in PR: https://github.com/rapidsai/ucxx/pull/374

We can simplify this to `py_`, which is typical for `noarch: python` packages

```suggestion
      string: py_${{ date_string }}_${{ head_rev }}
```

bdice:
Do we want `py_` in the build string here? I don't think a change is necessary here unless you know of a reason that we should include `py_` in the build string. I pushed a fix to https://github.com/rapidsai/integration/pull/757 that supports the existing build string, which I think is a better outcome.

jakirkham:
That is typically how `noarch: python` packages are indicated

Please see these examples:

* https://anaconda.org/conda-forge/setuptools/files
* https://anaconda.org/rapidsai/rapids-dask-dependency/files

The underscore isn't strictly necessary, but it may help with readability

The main benefit of this is we know the package is `noarch: python` and not `noarch: generic`, which behave differently

bdice (approved):
Thanks for clarifying `noarch: python` vs. `noarch: generic`. Then this seems fine! Approving.

jakirkham:
Thanks Bradley! 🙏

PR #411, opened 2025-05-07
Exclude librmm.so from auditwheel

bdice:
librmm will ship a shared library component in 25.06 (xref: https://github.com/rapidsai/rmm/issues/1779). This PR updates `auditwheel` calls to exclude `librmm.so`.


pentschev (approved):
LGTM, thanks @bdice !

gforsyth (approved):

PR #418, opened 2025-05-16
Fix persist usage in tests

TomAugspurger:
Recent versions of dask removed the ability to call the synchronous `<collection>.persist()` with an async client.

Use `client.persist()` instead.

Closes rapidsai/rapids-dask-dependency#103


TomAugspurger:
The commits linked from https://github.com/rapidsai/rapids-dask-dependency/issues/103 fixed issues that should have been caught here. We need to figure out why those tests didn't run / didn't fail.

pentschev (approved):
Thanks @TomAugspurger !

PR #419, opened 2025-05-19
Branch 0.45 merge branch 0.44

vyasr:
Replaces https://github.com/rapidsai/ucxx/pull/417


pentschev (approved):
Approving, but I think we still need cuDF packages for Python 3.13 first, and I presume an admin merge too.

vyasr:
I was hoping to get CI passing on this PR but at this point we have so many blocked forward mergers that it's probably not feasible. We should just admin merge all of them and then see how the nightlies look tonight.

PR #420, opened 2025-05-20
Add back Python 3.13 builds

vyasr:
These were dropped in https://github.com/rapidsai/ucxx/commit/010e0dac43b50f0fb756814c987f0bbd57994c7d, I'm not sure why.


gforsyth (approved):
Thanks for catching that.  We disabled them initially to get CI passing, because there weren't 3.13 packages available for some of the dependencies and there was no way to exclude them using  a matrix filter.

PR #424, opened 2025-05-22
Add fix for devices that do not have memory resources

pentschev:
Devices without a dedicated memory resource will raise a `pynvml.NVMLError_NotSupported` exception, we then leave those to UCX's default.

Additionally the exception has been renamed to `pynvml.NVMLError_NotSupported`, it was either wrong or previously existed under that submodule but it does not exist anymore.

Finally, fix missing exception handling around BAR1 memory info querying.


madsbk (approved):
Looks good.

pentschev:
Thanks @madsbk !

jakirkham (approved):

PR #427, opened 2025-05-23
Avoid linking to libpython

### LINKED ISSUE #425: Remove libpython linkage from libucxx_python
Currently libucxx_python.so links to libpython.so. This linkage is typically fine for conda packages, but it can break in subtle ways and it is explicitly forbidden by the manylinux standard for wheels. The problem is that libpython may not always be available, but the python symbols should be, so we should be dropping that linkage at compile time. This should be relatively straightforward for us if we simply drop the linkage in our CMake. We may have to do a bit of extra handling to get the include directories from the target and pass them along, though.

vyasr:
Resolves #425


bdice (approved):

pentschev (approved):
I confirmed locally as well that this indeed works, very cool! Thanks Vyas!

PR #437, opened 2025-06-06
Add support to query UCP debug info from requests

pentschev:
Add an opt-in mechanism for querying UCP request attributes (memory type, debug string) on every `ucxx::Request`. Enabled via a new `ucxx::experimental::WorkerBuilder::requestAttributes(true)` option. When enabled, every `ucxx::Request` submit site funnels through a small `Request::publishRequest()` helper that stores the UCP handle and queries `ucp_request_query` under the existing `_mutex`. `ucp_request_free` moves from `Request::callback` into `Request::setStatus`, making the query and the free mutually exclusive without any new atomics or callback-side locking. Wired through every request type and exposed to users via `Request::queryAttributes()`, which throws `ucxx::UnsupportedError` when the feature is disabled on the owning worker and `ucxx::NoElemError` when UCX took an inline path that produced no UCP handle to query (e.g., an eager UCX transfer).

Tag, AM, and MemoryGet test are asserted strictly above the rendezvous threshold, where UCX deterministically allocates a queryable request on every transport. Stream and MemoryPut use lenient assertions (substring-check on success, throw is acceptable) because stream has no rendezvous protocol and small RMA puts are fire-and-forget, both of which have transport-dependent inline-completion behavior that no fixed size threshold can portably predict.

The toggle is worker-scoped: enabling it queries attributes on **every** request created from that worker, which has potentially non-negligible per-request cost. Fine-grained per-request opt-in (so callers can attribute-query only the requests they care about) is not implemented here, it requires a builder-pattern constructor at the request level which doesn't exist yet, and is deferred to a follow-up. For now, users who need attributes accept the worker-wide cost, and users who don't, opt out by leaving the default.


cpp/include/ucxx/request.h
```diff
@@ -54,6 +63,8 @@ class Request : public Component {
   bool _enablePythonFuture{true};  ///< Whether Python future is enabled for this request
   RequestCallbackUserFunction _callback{nullptr};  ///< Completion callback
   RequestCallbackUserData _callbackData{nullptr};  ///< Completion callback data
+  Attributes _requestAttr{};        ///< Request attributes queried when request is posted
+  bool _isRequestAttrValid{false};  ///< Whether the request attributes are valid
```
wence-: A default-constructed attribute has `UCS_MEMORY_TYPE_UNKNOWN`. But a valid one must have a known memory type, can we use that to check if the request attribute is valid rather than having a separate flag?

cpp/include/ucxx/request.h
```diff
@@ -235,6 +246,63 @@ class Request : public Component {
...
+   *
+   * Single critical section that stores the UCP request pointer in `_request` and, when
+   * the owning worker has request attributes querying enabled, immediately queries those
+   * attributes. The completion path frees the UCP request inside `setStatus` under the
+   * same `_mutex`, so this helper guarantees the query and the free are mutually
+   * exclusive and that there are no use-after-free in threaded progress modes.
+   *
+   * Every submit site (all `request` methods from child classes and the AM
+   * rendezvous-receive path) calls this after obtaining the request handle from the
+   * corresponding `ucp_*_nbx` function.
```
wence-: Trying to list the submission sites like this parenthetical is a recipe for documentation going out of date.

cpp/include/ucxx/worker.h
```diff
@@ -492,6 +494,19 @@ class Worker : public Component {
...
+   *
+   * Check whether the worker has been created with request attributes querying enabled.
+   * When enabled, each `ucxx::Request` will have its UCP attributes (such as the debug
+   * string) queried immediately after submission, making them available via
+   * `ucxx::Request::queryAttributes()`. Querying request attributes has a
+   * non-negligible runtime cost and is therefore disabled by default.
+   *
+   * @returns `true` if request attributes querying is enabled, `false` otherwise.
+   */
+  [[nodiscard]] bool isRequestAttributesEnabled() const;
```
wence-:
```suggestion
  [[nodiscard]] bool isRequestAttributesEnabled() const noexcept;
```
?

cpp/src/endpoint.cpp
```diff
@@ -288,8 +288,38 @@ void Endpoint::closeBlocking(uint64_t period, uint64_t maxAttempts)
...
+                status = ucp_ep_close_nbx(_handle, &param);
+                // Invalidate _handle synchronously immediately, to prevent
+                // time window where _handle` points to freed UCP memory, usually
+                // observed in `populateDelayedSubmission()`.
+                _originalHandle = _handle;
+                _handle         = nullptr;
+              },
+              period))
           continue;
         submitted = true;
```
wence-: This doesn't seem to be anything to do with request attributes.

cpp/src/endpoint.cpp
```diff
@@ -288,8 +288,38 @@ void Endpoint::closeBlocking(uint64_t period, uint64_t maxAttempts)
...
+        // Why atomic with FORCE close (not as a separate pre-callback):
+        //   When cancelAll and FORCE close were separate pre-callbacks (the
+        //   old cancelInflightRequestsBlocking path), a full ucp_worker_progress()
+        //   ran between them.  That intermediate progress could leave UCT-level
+        //   TCP pending entries half-dispatched (mid-cuMemcpyAsync staging of
+        //   a CUDA send); the next progress after FORCE close then crashed
+        //   dispatching them on a freed staging buffer (uct_cuda_copy_ep_get_short
+        //   -> cuMemcpyAsync -> SIGSEGV).  Running them in a single pre-callback
+        //   matches the safe single-threaded ordering proven by the regression
+        //   test in cpp/tests/endpoint_close_force_tcp_cuda_race.cpp.
```
wence-: Please rework this comment so that it makes sense independently of any historical route to the code we have now.

Something like:

> Cancellation and forced closing of requests must happen without UCX progress so that handles are not left in a half-initialised state. For example, if cancellation and close are separate, progress could result in the requests being in the middle of a cuMemcpyAsync when the close runs.

?

wence-: This is a completely hallucinated filename?

cpp/src/request.cpp
```diff
@@ -235,11 +234,75 @@ void Request::setStatus(ucs_status_t status)
         _ownerString.c_str(), this, _request, _operationName.c_str(), "invoking user callback");
       _callback(status, _callbackData);
     }
+
+    // Free the UCP request inside the lock so it is mutually exclusive with
+    // `publishRequest()`/`queryRequestAttributes()` on the submit thread. Clearing
+    // `_request` afterwards keeps this idempotent if `setStatus` ever re-enters.
+    if (UCS_PTR_IS_PTR(_request)) {
```
wence-, line 240:
```suggestion
    // Free the UCP request inside the lock so it is mutually exclusive with
    // `publishRequest()`/`queryRequestAttributes()` on the submit thread.
    if (UCS_PTR_IS_PTR(_request)) {
```
Yes, we obviously must clear the request after freeing it.

cpp/src/request.cpp
```diff
@@ -235,11 +234,75 @@ void Request::setStatus(ucs_status_t status)
...
+
+  if (_isRequestAttrValid) return;
+  if (!_worker->isRequestAttributesEnabled()) return;
+
+  ucp_request_attr_t result;
+
+  // Get the debug string size from worker attributes
+  auto worker_attr = _worker->queryAttributes();
+
+  // Allocate buffer for debug string with size from worker attributes
```
wence-:
```suggestion
```

cpp/src/request.cpp
```diff
@@ -235,11 +234,75 @@ void Request::setStatus(ucs_status_t status)
...
+void Request::queryRequestAttributes()
+{
+  std::lock_guard<std::recursive_mutex> lock(_mutex);
+
+  if (_isRequestAttrValid) return;
+  if (!_worker->isRequestAttributesEnabled()) return;
+
+  ucp_request_attr_t result;
+
+  // Get the debug string size from worker attributes
```
wence-:
```suggestion
```

cpp/src/request.cpp
```diff
@@ -235,11 +234,75 @@ void Request::setStatus(ucs_status_t status)
...
+  auto worker_attr = _worker->queryAttributes();
+
+  // Allocate buffer for debug string with size from worker attributes
+  std::vector<char> debug_str(worker_attr.maxDebugString, '\0');
+
+  result.field_mask = UCP_REQUEST_ATTR_FIELD_MEM_TYPE |         // Memory type
+                      UCP_REQUEST_ATTR_FIELD_INFO_STRING |      // Debug string
+                      UCP_REQUEST_ATTR_FIELD_INFO_STRING_SIZE;  // Debug string size
+
```
wence-:
```suggestion
  result.field_mask = UCP_REQUEST_ATTR_FIELD_MEM_TYPE |
                      UCP_REQUEST_ATTR_FIELD_INFO_STRING |
                      UCP_REQUEST_ATTR_FIELD_INFO_STRING_SIZE;

```
It feels like the enum names are clear enough.

cpp/src/request.cpp
```diff
@@ -235,11 +234,75 @@ void Request::setStatus(ucs_status_t status)
...
+  // Set up the debug string buffer
+  result.debug_string      = debug_str.data();
+  result.debug_string_size = debug_str.size();
+
```
wence-:
```suggestion
  result.debug_string      = debug_str.data();
  result.debug_string_size = debug_str.size();

```

cpp/src/request.cpp
```diff
@@ -235,11 +234,75 @@ void Request::setStatus(ucs_status_t status)
...
+  if (UCS_PTR_IS_PTR(_request)) {
+    auto queryStatus = ucp_request_query(_request, &result);
+    if (queryStatus == UCS_OK && result.debug_string != nullptr) {
+      _requestAttr.debugString = std::string(result.debug_string);
```
wence-: So we allocated space for the debug string, then we copy it. Why not resize the `debug_string` based on its size and `std::move()` it?

cpp/src/request.cpp
```diff
@@ -235,11 +234,75 @@ void Request::setStatus(ucs_status_t status)
...
 }
 
 const std::string& Request::getOwnerString() const { return _ownerString; }
 
+void Request::queryRequestAttributes()
+{
+  std::lock_guard<std::recursive_mutex> lock(_mutex);
+
+  if (_isRequestAttrValid) return;
+  if (!_worker->isRequestAttributesEnabled()) return;
```
wence-: *nit*: Return early before trying to grab the lock if we're not enabled.

cpp/src/request.cpp
```diff
@@ -235,11 +234,75 @@ void Request::setStatus(ucs_status_t status)
...
+    if (UCS_PTR_IS_PTR(_request)) {
+      ucp_request_free(_request);
+      _request = nullptr;
+    }
   }
 }
 
 const std::string& Request::getOwnerString() const { return _ownerString; }
 
+void Request::queryRequestAttributes()
```
wence-: This method is _only_ called from `publishRequest`. I think we could just inline the implementation there.

cpp/src/request.cpp
```diff
@@ -235,11 +234,75 @@ void Request::setStatus(ucs_status_t status)
...
+Request::Attributes Request::queryAttributes()
+{
+  std::lock_guard<std::recursive_mutex> lock(_mutex);
+
+  if (_isRequestAttrValid) return _requestAttr;
+
+  if (!_worker->isRequestAttributesEnabled())
+    throw ucxx::UnsupportedError(
+      "Request attributes querying is disabled on the owning worker; build the worker "
+      "with `ucxx::experimental::WorkerBuilder::requestAttributes(true)` to enable it");
```
wence-, line 289: Again, move this before grabbing the lock.

wence- (changes requested):

cpp/src/endpoint.cpp
pentschev: Ugh, I'm really sorry about this. Those changes were for a completely different branch, I was developing and testing this and the other branch on a remote machine and scp-ing patches, and I accidentally applied this to the current branch. I've reverted this now.

cpp/src/endpoint.cpp
pentschev: Same here, accidentally applied patch to the wrong branch, reverted.

cpp/include/ucxx/request.h
pentschev: Good point, cleaned up in https://github.com/rapidsai/ucxx/pull/437/commits/3716e5d6c0c75c35ab6dce8f9087e4357b2e27c8 .

cpp/include/ucxx/request.h
pentschev: Done in https://github.com/rapidsai/ucxx/pull/437/commits/6cd44e5b090a3713b4ef2aca762bc38297f09df6 .

cpp/src/request.cpp
pentschev: Done in https://github.com/rapidsai/ucxx/pull/437/commits/a37d5e020bd0c5f886fba2acc669323988b8f673 .

cpp/src/request.cpp
pentschev: Done in https://github.com/rapidsai/ucxx/pull/437/commits/3e92f4ba1487f7499c85b815d572d91f502a9b9e .

cpp/src/request.cpp
pentschev: Done in https://github.com/rapidsai/ucxx/pull/437/commits/3e92f4ba1487f7499c85b815d572d91f502a9b9e .

cpp/src/request.cpp
pentschev: Done in https://github.com/rapidsai/ucxx/pull/437/commits/a1a5c177f8fb64fae2a6f497c1e88fdf7d69453e .

pentschev (commented):
Thanks @wence- for the review. I think I have addressed all your comments, please take another look when you have a chance.

ci/run_cpp.sh
```diff
@@ -45,7 +45,7 @@ else
 fi
 
 run_cpp_tests() {
-  CMD_LINE="python ${TIMEOUT_TOOL_PATH} $((10*60)) ${GTESTS_PATH}/UCXX_TEST"
+  CMD_LINE="python ${TIMEOUT_TOOL_PATH} $((20*60)) ${GTESTS_PATH}/UCXX_TEST"
```
pentschev, line 48: Some of the slow runners are already running close to the current 10 minutes timeout ([sample from last night](https://github.com/rapidsai/ucxx/actions/runs/25784126090/job/75794489664) ran from 07:19:36 until 07:28:13, for a total of 08:37 minutes), with the addition of new tests they occasionally do timeout.

wence- (approved):

bdice (approved):
Approving CI.

pentschev:
Thanks Lawrence and Bradley.

PR #438, opened 2025-06-10
Remove pytest pin

vyasr:
Contributes to https://github.com/rapidsai/build-planning/issues/105.


pentschev:
> I'm holding off on opening PRs to other repos until I can see if the issues around newer pytest version compatibility with plugins have been resolved.

Note that UCXX uses a bunch of advanced pytest features, like [here](https://github.com/rapidsai/ucxx/blob/branch-0.45/python/distributed-ucxx/distributed_ucxx/utils_test.py) and [here](https://github.com/rapidsai/ucxx/blob/branch-0.45/python/ucxx/ucxx/_lib_async/tests/conftest.py), so it may not be the best "guinea pig" for this, or perhaps it makes it one the best exactly because it uses more of those features, depending on what exactly you're trying to understand.

vyasr:
> depending on what exactly you're trying to understand.

Well the issue is that we are now more than a year behind the latest pytest releases, including some bug fixes and feature improvements that we would like to take advantage of. That will eventually become problematic for us, so we'd like to be able to use the newer versions. It seems like there are still some issues that we need to get resolved here though.

vyasr:
I haven't had time to look at this again. Unfortunately it seems like there is something in ucxx's test suite that is incompatible with the latest versions of pytest, and I don't know what without some deeper investigation.

AyodeAwe (approved):

python/ucxx/ucxx/_lib_async/tests/conftest.py
pentschev, line 1: The logic here needed to be rewritten here to support `pytest-asyncio>0.23`, which is also required for `pytest>=8.0`, so these changes are needed here. See https://github.com/pytest-dev/pytest-asyncio/issues/706 for details about this change from pytest-asyncio.

pentschev (commented):
@rapidsai/ucxx-python-codeowners could one of you please review/approve the changes here?

@vyasr @bdice FYI this is now passing.

TomAugspurger (approved):

vyasr:
Oh wow thank you for tackling this Peter! I don't know much about pytest-asyncio and was expecting to have to spend a bit of time reading to figure out what could be going wrong, this is super helpful.

python/ucxx/ucxx/_lib_async/tests/conftest.py
vyasr: Hmm interesting OK. I may have some questions after looking through.

PR #439, opened 2025-06-11
Remove CUDA 11 from dependencies.yaml

KyleFromNVIDIA:
Issue: https://github.com/rapidsai/build-planning/issues/184


bdice (approved):
I dropped the `nvidia` channel, only needed for CUDA 11. Otherwise approved.

PR #449, opened 2025-06-30
Disable Python garbage collection during distributed-ucxx registration

pentschev:
During registration of UCXX resources in distributed-ucxx it is possible that the garbage collector runs and the finalizer that deregisters resources then attempts to grab the same lock and deadlock. Here's an example of a stack in which this may occur:

```
Thread 1 (Thread 0x7ffff7c96740 (LWP 487526) "python"):
Traceback (most recent call first):
  <built-in method __enter__ of _thread.lock object at remote 0x7fffb0ebc340>
  File "/opt/conda/envs/test/lib/python3.10/site-packages/distributed_ucxx/ucxx.py", line 151, in _deregister_dask_resource
    with ctx._dask_resources_lock:
  File "/opt/conda/envs/test/lib/python3.10/site-packages/distributed_ucxx/ucxx.py", line 326, in _finalizer
    _deregister_dask_resource(resource_id)
  File "/opt/conda/envs/test/lib/python3.10/weakref.py", line 591, in __call__
    return info.func(*info.args, **(info.kwargs or {}))
  Garbage-collecting
  File "/opt/conda/envs/test/lib/python3.10/socket.py", line 658, in socketpair
    b = socket(family, type, proto, b.detach())
  File "/opt/conda/envs/test/lib/python3.10/site-packages/ucxx/_lib_async/continuous_ucx_progress.py", line 121, in __init__
    self.rsock, wsock = socket.socketpair()
  File "/opt/conda/envs/test/lib/python3.10/site-packages/ucxx/_lib_async/application_context.py", line 481, in continuous_ucx_progress
    task = BlockingMode(self.worker, loop)
  File "/opt/conda/envs/test/lib/python3.10/site-packages/distributed_ucxx/ucxx.py", line 117, in register
    ctx.continuous_ucx_progress()
  File "/opt/conda/envs/test/lib/python3.10/site-packages/distributed_ucxx/ucxx.py", line 720, in start
    self._resource_id = _register_dask_resource()
  File "/opt/conda/envs/test/lib/python3.10/site-packages/distributed/comm/core.py", line 256, in _
    await self.start()
  File "/opt/conda/envs/test/lib/python3.10/site-packages/distributed/core.py", line 723, in listen
    listener = await listen(
...
```

The solution here is to disable the garbage collector from running when registering, preventing finalizers to run during that stage.

An alternative could have been to use a recursive lock, but that _may_ be dangerous if the garbage collector ends up running while `ctx._dask_resources` is being modified.

One other alternative could be to just move
`ctx.start_notifier_thread()` + `ctx.continuous_ucx_progress()` out of the block where the lock is acquired. I believe this would be fine too, but disabling the garbage collector seems like a safer approach overall.

### DIFF
--- a/python/distributed-ucxx/distributed_ucxx/ucxx.py
+++ b/python/distributed-ucxx/distributed_ucxx/ucxx.py
@@ -1,3 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023-2025, NVIDIA CORPORATION & AFFILIATES.
+# SPDX-License-Identifier: BSD-3-Clause
+
 """
 :ref:`UCX`_ based communications for distributed.
 
@@ -8,6 +11,7 @@ See :ref:`communications` for more.
 from __future__ import annotations
 
 import functools
+import gc
 import itertools
 import logging
 import os
@@ -91,6 +95,17 @@ def synchronize_stream(stream=0):
     stream.synchronize()
 
 
+class gc_disabled:
+    def __enter__(self):
+        self.was_enabled = gc.isenabled()
+        if self.was_enabled:
+            gc.disable()
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if self.was_enabled:
+            gc.enable()
+
+
 def make_register():
     count = itertools.count()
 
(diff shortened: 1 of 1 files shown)


wence-:
> `ctx.start_notifier_thread()` + `ctx.continuous_ucx_progress()` out of the block where the lock is acquired. I believe this would be fine too, but disabling the garbage collector seems like a safer approach overall.

This would not be safe, I think, if multiple threads are running.

wence- (approved):

pentschev:
Thanks for the review and (offline) discussion @wence- !

PR #451, opened 2025-07-01
Move CI test to run scripts

pentschev:
This makes CI scripts more aligned with the rest of the RAPIDS projects, making it also easier to reuse the same test set for both conda and wheels.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -9,15 +9,21 @@ on:
   workflow_dispatch:
     inputs:
       branch:
+        description: |
+          branch: git branch the workflow run targets.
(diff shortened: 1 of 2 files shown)


jameslamb (approved):
I support this! I only skimmed the changes, assuming this was mostly just moving code around and didn't introduce any behavior changes.

Left one other small, non-blocking suggestion, which could be done in a follow-up.

pentschev:
> I support this! I only skimmed the changes, assuming this was mostly just moving code around and didn't introduce any behavior changes.
> 
> Left one other small, non-blocking suggestion, which could be done in a follow-up.

That is the case, no behavioral changes exist, except that we now run the full test-suite for wheels too.

PR #453, opened 2025-07-02
Override default progress mode for distributed-ucxx

pentschev:
Switch the default distributed-ucxx progress mode to `"blocking"`, which is equivalent to UCX-Py's default progress mode, allowing a more transparent move of Dask to UCXX. This change may later be reverted once we have more confidence in a successful transition to UCXX.

Additionally allow both `"ucx://"` and `"ucxx://"` address prefixes, since with the change from https://github.com/rapidsai/rapids-dask-dependency/pull/116 both will be valid to enable UCXX when distributed-ucxx is installed.

### DIFF
--- a/python/distributed-ucxx/distributed_ucxx/ucxx.py
+++ b/python/distributed-ucxx/distributed_ucxx/ucxx.py
@@ -1,3 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023-2025, NVIDIA CORPORATION & AFFILIATES.
+# SPDX-License-Identifier: BSD-3-Clause
+
 """
 :ref:`UCX`_ based communications for distributed.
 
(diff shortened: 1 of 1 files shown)


python/distributed-ucxx/distributed_ucxx/ucxx.py
```diff
@@ -253,7 +256,10 @@ def init_once():
         # that don't override ucx_config or existing slots in the
         # environment, so the user's external environment can safely
         # override things here.
-        ucxx.init(options=ucx_config, env_takes_precedence=True)
+        progress_mode = None if "UCXPY_PROGRESS_MODE" in os.environ else "blocking"
```
TomAugspurger, line 259: Do we care what the value of the `UCXPY_PROGRESS_MODE` environment variable is?

python/distributed-ucxx/distributed_ucxx/ucxx.py
```diff
@@ -378,8 +384,8 @@ def __init__(  # type: ignore[no-untyped-def]
         self._ep = ep
         self._ep_handle = int(self._ep._ep.handle)
         if local_addr:
-            assert local_addr.startswith("ucxx")
-        assert peer_addr.startswith("ucxx")
+            assert local_addr.startswith(("ucx://", "ucxx://"))
```
TomAugspurger, line 387: Can this be tested yet, or not until the other PR is merged?

pentschev: Not until https://github.com/rapidsai/rapids-dask-dependency/pull/116 is merged, and I think not even after, because we can't use another protocol to hit this conditional, or at least not easily. Probably not worth the time invested at the moment.

pentschev:
@TomAugspurger anything else here or are you good with previous answers?

TomAugspurger (approved):
All good, thanks.

PR #384, opened 2025-03-13
Enable protov2 by default on UCX >= 1.18.0

pentschev:
With protov1 not being actively developed anymore and protov2 now being more stable, it's time to switch permanently. See https://github.com/rapidsai/ucx-py/issues/1121 for more details.

Closes https://github.com/rapidsai/ucx-py/issues/1121 .

PR #456, opened 2025-07-07
Run all C++ tests with protov2

pentschev:
Remove C++ tests with protov1 and enable all C++ tests with protov2.

UCX protov1 is now outdated, with multiple features required for best performance in RAPIDS that are only supported on protov2. UCXX has been using protov2 as the default for a while (see https://github.com/rapidsai/ucxx/pull/384) so it should be good to disable the last remaining protov1 testing and test only protov2.

### DIFF
--- a/ci/run_cpp.sh
+++ b/ci/run_cpp.sh
@@ -17,17 +17,11 @@ run_cpp_tests() {
   RUNTIME_PATH=${CONDA_PREFIX:-./}
   BINARY_PATH=${RUNTIME_PATH}/bin
 
-  # Disable memory get/put with RMM in protov1, it always segfaults.
-  CMD_LINE="timeout 10m ${BINARY_PATH}/gtests/libucxx/UCXX_TEST --gtest_filter=-*RMM*Memory*"
+  # Only test memory get/put with RMM in protov2, as protov1 segfaults.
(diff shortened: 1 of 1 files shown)


ci/run_cpp.sh
```diff
@@ -17,17 +17,11 @@ run_cpp_tests() {
   RUNTIME_PATH=${CONDA_PREFIX:-./}
   BINARY_PATH=${RUNTIME_PATH}/bin
 
-  # Disable memory get/put with RMM in protov1, it always segfaults.
-  CMD_LINE="timeout 10m ${BINARY_PATH}/gtests/libucxx/UCXX_TEST --gtest_filter=-*RMM*Memory*"
-
-  log_command "${CMD_LINE}"
-  UCX_TCP_CM_REUSEADDR=y ${CMD_LINE}
-
   # Only test memory get/put with RMM in protov2, as protov1 segfaults.
```
bdice, line 20: Do we still need this comment about protov1?

Also, where is the proto version being set? I see `UCX_PROTO_ENABLE=y` was removed below. What does that mean?

bdice (approved):
I left a question but I trust you did this correctly since CI is passing. Merge once you feel my question is sufficiently addressed.

pentschev:
Thanks @bdice !

PR #462, opened 2025-07-15
Update to numba-cuda>=0.16

pentschev:
Update to numba-cuda>=0.16 [to match cuDF](https://github.com/rapidsai/cudf/pull/19213), and update stream synchronization code in distributed-ucxx.

Closes https://github.com/rapidsai/dask-upstream-testing/issues/71.


python/distributed-ucxx/distributed_ucxx/ucxx.py
```diff
@@ -86,13 +87,18 @@ def _warn_cuda_context_wrong_device(
...
     import numba.cuda
 
-    ctx = numba.cuda.current_context()
-    cu_stream = numba.cuda.driver.drvapi.cu_stream(stream)
-    stream = numba.cuda.driver.Stream(ctx, cu_stream, None)
-    stream.synchronize()
+    if stream == CudaStream.Default:
+        numba_stream = numba.cuda.default_stream()
+    else:
+        raise ValueError("Unsupported stream")
```
TomAugspurger, line 100: Is this a behavior change, and if so is it acceptable? about why `synchronize_stream` accepted a `stream` argument when it didn't seem to use anything other than the default anywhere.

TomAugspurger (approved):

pentschev:
I've preemptively triggered an auto-merge, without this PR many of our projects that test Dask+UCXX will fail nightlies because of the change to numba-cuda=0.16 that had some breaking API changes. @jakirkham if you're still around would appreciate if you could approve this on behalf of `ucxx-python-codeowners`, you're the only one who's not currently on PTO besides me.

TomAugspurger:
Gil got me added to `ucxx-python-codeowners` so we should be good to go here.

PR #464, opened 2025-07-16
Fix CI run script

pentschev:
In https://github.com/rapidsai/ucxx/pull/451 the test run logic got moved to `ci/run_*` scripts to match other RAPIDS projects. One of the scripts mistakenly still uses `rapids-logger`, which is incorrect, this change fixes that.

Additionally pick the same `distributed` version as the currently installed one for tests required editable install.


bdice (approved):

pentschev:
Thanks for this one too @bdice !

PR #469, opened 2025-07-18
Make site-package discovery more robust to remove pycache

pentschev:
Use `site.getsitepackages()` to discover the proper site-package directories and subsequently remove `distributed` directory containing `__pycache__` from previous runs to ensure editable install works correctly for certain tests requiring Distributed installed in editable mode.


bdice (approved):

pentschev:
Thanks @bdice !

PR #476, opened 2025-07-25
Return immediately if AM request was already completed

pentschev:
If an Active Message receive request had its status set before the callback executed, return immediately. I believe the only case this may occur currently is if a `amRecv()` request is posted and canceled by the user before the request arrived, but cover all statuses just in case.


madsbk (approved):
Looks good to me

PR #484, opened 2025-08-11
Use build cluster in devcontainers

trxcllnt:
RAPIDS has deployed an autoscaling cloud build cluster that can be used to accelerate building large RAPIDS projects. This contributes to https://github.com/rapidsai/build-planning/issues/209.


.devcontainer/cuda12.9-conda/devcontainer.json
```diff
@@ -11,7 +11,9 @@
   "runArgs": [
     "--rm",
     "--name",
-    "${localEnv:USER:anon}-rapids-${localWorkspaceFolderBasename}-25.10-cuda12.9-conda"
+    "${localEnv:USER:anon}-rapids-${localWorkspaceFolderBasename}-25.10-cuda12.9-conda",
+    "--ulimit",
+    "nofile=500000"
```
pentschev, line 16: It might be worth replicating options we use in CI:

https://github.com/rapidsai/ucxx/blob/a5ebe1244db51645970f862aa1bfa7ae629f1a7c/.github/workflows/pr.yaml#L106

Those are important for UCX.

AyodeAwe (approved):

.devcontainer/cuda12.9-conda/devcontainer.json
pentschev: I don't ever use devcontainers, so I'd probably not be able to test it either. Let's leave it like this for now.

PR #494, opened 2025-08-29
Pin pytest-rerunfailures<16

mroeschke:
Preemptively pinning due to xref https://github.com/pytest-dev/pytest-rerunfailures/issues/302 cc @jameslamb


jameslamb (approved):
Thank you!

PR #495, opened 2025-09-01
Remove setting of RMM_LOG_ACTIVE_LEVEL

wence-:
Now that RMM is (at least partially) a shared library, we need to ensure we pick up the active log level from the upstream install. So remove these overrides.


pentschev (approved):
LGTM.

Issue #350, opened 2025-01-21
Expose connect timeout configuration


pentschev:
Similar to https://github.com/rapidsai/ucx-py/pull/1103, connect timeout must be exposed as configuration here.

DrBinary9:
@pentschev I would like to take this issue. I have gone through the changes in the PR mentioned above. Could you please specify where these changes have to be implemented?

pentschev:
Hi @DrBinary9 , all changes included in the PR linked in the description need to be implemented. The PR essentially does the following:

1. Rename the `stream_timeout` argument to `connect_timeout`, this needs to be done in https://github.com/rapidsai/ucxx/blob/7785a23ac67451333cc168f59d9c95142747c735/python/ucxx/ucxx/_lib_async/exchange_peer_info.py#L16-L50
2. Add an environment variable to allow configuration of the option, needs to be added to https://github.com/rapidsai/ucxx/blob/7785a23ac67451333cc168f59d9c95142747c735/python/ucxx/ucxx/_lib_async/application_context.py#L43-L77
3. Add an argument to the `init` function, needs to be added to https://github.com/rapidsai/ucxx/blob/7785a23ac67451333cc168f59d9c95142747c735/python/ucxx/ucxx/core.py#L32-L89.
4. We should probably rename `exchange_peer_info_timeout` to `connect_timeout` as well in https://github.com/rapidsai/ucxx/blob/7785a23ac67451333cc168f59d9c95142747c735/python/ucxx/ucxx/_lib_async/application_context.py#L49 and https://github.com/rapidsai/ucxx/blob/7785a23ac67451333cc168f59d9c95142747c735/python/ucxx/ucxx/_lib_async/listener.py#L140 .

Due to the [upcoming archival of UCX-Py](https://docs.rapids.ai/notices/rsn0050/) which is currently ongoing https://github.com/rapidsai/ucx-py/issues/1160, this is something that will need to be done relatively soon, before the [burn down on September 18 starts](https://docs.rapids.ai/maintainers/). Please let me know if this is indeed something you'd be interested in working on.

DrBinary9:
@pentschev Yes, I'll start working on it right away

DrBinary9:
@pentschev I have made the changes. Can you please assign this task to me?

PR #496, opened 2025-09-01
Replaced every occurrence of stream_timeout and exchange_peer_info_timeout to connect_timeout, added env var support, and added connect_timeout as an argument to init in core.py

DrBinary9:
Issue #350 

Renamed stream_timeout → connect_timeout in exchange_peer_info, application_context.py

Renamed exchange_peer_info_timeout → connect_timeout in application_context.py and listener.py, core.py, test_multiple_nodes.py.

Added support for configuring connect_timeout via environment variable,

Exposed connect_timeout as an argument in init() and propagated it to ApplicationContext.

### DIFF
--- a/python/ucxx/ucxx/_lib_async/application_context.py
+++ b/python/ucxx/ucxx/_lib_async/application_context.py
@@ -46,7 +46,7 @@ class ApplicationContext:
         progress_mode=None,
         enable_delayed_submission=None,
         enable_python_future=None,
-        exchange_peer_info_timeout=10.0,
+        connect_timeout=10.0,
     ):
         self.notifier_thread_q = None
         self.notifier_thread = None
@@ -57,7 +57,12 @@ class ApplicationContext:
         self.enable_delayed_submission = enable_delayed_submission
         self.enable_python_future = enable_python_future
 
-        self.exchange_peer_info_timeout = exchange_peer_info_timeout
+        if "UCXPY_CONNECT_TIMEOUT" in os.environ:
+            self.connect_timeout = float(os.environ["UCXPY_CONNECT_TIMEOUT"])
+        elif connect_timeout is not None:
+            self.connect_timeout = connect_timeout
+        else:
+            self.connect_timeout = 10.0
 
         # For now, a application context only has one worker
         self.context = ucx_api.UCXContext(config_dict)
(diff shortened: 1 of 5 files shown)


python/ucxx/ucxx/_lib_async/application_context.py
```diff
@@ -46,7 +46,7 @@ def __init__(
         progress_mode=None,
         enable_delayed_submission=None,
         enable_python_future=None,
-        exchange_peer_info_timeout=10.0,
+        connect_timeout=10.0,
```
pentschev, line 49:
```suggestion
        connect_timeout=None,
```

pentschev: Let's switch to `None`, like in UCX-Py, this allows us to simplify the conditional.

python/ucxx/ucxx/_lib_async/application_context.py
```diff
@@ -57,7 +57,12 @@ def __init__(
         self.enable_delayed_submission = enable_delayed_submission
         self.enable_python_future = enable_python_future
 
-        self.exchange_peer_info_timeout = exchange_peer_info_timeout
+        if "UCXPY_CONNECT_TIMEOUT" in os.environ:
+            self.connect_timeout = float(os.environ["UCXPY_CONNECT_TIMEOUT"])
+        elif connect_timeout is not None:
+            self.connect_timeout = connect_timeout
+        else:
+            self.connect_timeout = 10.0
```
pentschev, line 65:
```suggestion
        if connect_timeout is None:
            self.connect_timeout = float(os.environ.get("UCXPY_CONNECT_TIMEOUT", 5))
        else:
            self.connect_timeout = connect_timeout
```

pentschev: With `None`, we can simplify this now, plus we keep the same order as in the rest of UCXX, priority to the explicit argument, then to the environment variable.

python/ucxx/ucxx/core.py
```diff
@@ -35,6 +35,7 @@ def init(
     progress_mode=None,
     enable_delayed_submission=None,
     enable_python_future=None,
+    connect_timeout=None,
```
pentschev, line 38: Can you add the missing docstring too? See https://github.com/rapidsai/ucx-py/blob/75c49e661947d55efda3f2bcdf6d72bb1bf01e91/ucp/core.py#L933-L936 for reference.

pentschev (changes requested):
Left a few minor change requests, but this is looking good. Thanks @DrBinary9 for quickly working on this!

DrBinary9:
Done

PR #499, opened 2025-09-02
Prevent installation of pytest-rerunfailures 16.0.0

pentschev:
The `pytest-rerunfailures=16.0.0` package had a breaking change (see https://github.com/pytest-dev/pytest-rerunfailures/issues/302) and thus in https://github.com/rapidsai/ucxx/pull/494 we pinned to prevent installation of version 16 and higher. Although some of the bad builds were yanked, the PyPI build is still not yanked, so this change prevents installing only the bad release.


KyleFromNVIDIA (approved):

pentschev:
> Yes this is a nice improvement, and something similar should be done in `cudf`.
> 
> The key fact here is that a `16.0.1` version with a fix only went up (at least on PyPI) 7 hours ago: https://pypi.org/project/pytest-rerunfailures/#history
> 
> As of #494 there was no such release, so I think a `<` was justifiable there (to avoid needing to update again if a subsequent release still didn't fix the issue).

Yes, this is indeed the case. I'm sorry if I made this sound like a previous mistake from https://github.com/rapidsai/ucxx/pull/494, it definitely wasn't the intent, a different fix/workaround was not available at the time the previous PR was open.

With that said, I think you were just pointing that out, there's no reason why this shouldn't be merged as is now, can you confirm @jameslamb ?

pentschev:
Thanks James and Kyle!

Issue #502, opened 2025-09-02
Capability to measure GIL contention in Python benchmarks


pentschev:
In UCX-Py we have added a command-line argument that allows measuring GIL contention in benchmarks via https://github.com/rapidsai/ucx-py/pull/937. It would be good to add it to the UCXX benchmarks to avoid losing this feature.

DrBinary9:
@pentschev I'll start with this

DrBinary9:
if self.args.report_gil_contention:
                from gilknocker import KnockKnock

                # Use smallest polling interval
                # possible to ensure, contention will always
                # be zero for small messages otherwise
                # and inconsistent for large messages.
                knocker = KnockKnock(polling_interval_micros=1)
                knocker.start()

@pentschev I have defined knocker over here and pre-commit is throwing this error

python/ucxx/ucxx/benchmarks/backends/ucxx_core.py:301:28: F821 Undefined name `knocker`
Found 1 error.
in this line
        if self.args.report_gil_contention:
            self.queue.put(knocker.contention_metric)

How to resolve this

pentschev:
This has already been fixed, thanks @DrBinary9 !

PR #503, opened 2025-09-02
Fix issue 502

DrBinary9:
Issue #502 

Implement the code mentioned in https://github.com/rapidsai/ucx-py/pull/937 
A few assumptions were made 
1. knocker.stop() if we assume it stops once (line 116 in tornado.py)
2.  ucxx.stop_notifier_thread()  # why is this not there in ucx-py (line 200 ucxx_async.py)
3.  in the placement of the code blocks in as ucxx_core.py is quite different from ucp/benchmarks/backends/ucp_core.py


DrBinary9:
I am closing this as I PRed the wrong commits

python/ucxx/ucxx/benchmarks/backends/ucxx_async.py
```diff
@@ -177,11 +185,17 @@ async def run(self):
...
+            knocker.stop()
         if self.args.cuda_profile:
             xp.cuda.profiler.stop()
+
         self.queue.put(times)
+        if self.args.report_gil_contention:
+            self.queue.put(knocker.contention_metric)
 
-        ucxx.stop_notifier_thread()
+        ucxx.stop_notifier_thread()
```
pentschev, line 198:
```suggestion
        ucxx.stop_notifier_thread()
```

pentschev: This will fail style check.

pentschev (changes requested):
Thanks @DrBinary9 , the new changes look good. There's one minor request that will fail style check, I triggered testing to check if there are any others, when that's fixed and tests pass we're good to merge.

pentschev:
There are a few more style issues, could you do the same as I've suggested in the previous PR https://github.com/rapidsai/ucxx/pull/496#issuecomment-3244059731?

DrBinary9:
@pentschev Done

python/ucxx/ucxx/benchmarks/backends/ucxx_core.py
```diff
@@ -277,13 +286,17 @@ async def _transfer():
...
+                knocker.stop()
             if self.args.cuda_profile:
                 xp.cuda.profiler.stop()
 
         loop = get_event_loop()
         loop.run_until_complete(_transfer())
 
         self.queue.put(times)
+        if self.args.report_gil_contention:
+            self.queue.put(knocker.contention_metric)
```
pentschev, line 299: Unfortunately this doesn't work, because `knocker` wasn't defined in this scope, I'm surprised `pre-commit run -a` didn't fail for you.

What I suggest we do is define a new variable `contention_metric = None` after in line 248 (after `times = []`), unfortunately GitHub doesn't allow making suggestions on code unmodified by the PR. Then we could mark the variable `nonlocal contention_metric` immediately after the `async def _transfer():` (line 250), and then assign it immediately after stopping the knocker `contention_metric = knocker.contention_metric` (line 291), and finally here just remove `knocker.` to address directly the variable we created. Could you do that?

DrBinary9: Make the changes. Not sure if I had to define contention_metric = knocker.contention_metric within  
if self.args.report_gil_contention or not

python/ucxx/ucxx/benchmarks/backends/ucxx_core.py
```diff
@@ -277,13 +288,18 @@ async def _transfer():
                 if i >= self.args.n_warmup_iter:
                     times.append(stop - start)
 
+            if self.args.report_gil_contention:
+                knocker.stop()
+            contention_metric = knocker.contention_metric
```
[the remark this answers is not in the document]
DrBinary9: Done

Issue #173, opened 2024-01-22
Export ucx dependency


vyasr:
Once https://github.com/openucx/ucx/issues/9614 is resolved we should revert https://github.com/rapidsai/ucxx/pull/172

pentschev:
Those changes probably won't make their way in until UCX 1.17.x, maybe 1.16.x but I'll have to check when we get the fix merged.

vyasr:
Fix was merged and backported to 1.16.x, so we can rely on it in ucxx after 1.16 is released and we bump to building against that.

jameslamb:
Noting here that if/when we make that change, it could simplify `raft-dask` too:

https://github.com/rapidsai/raft/blob/3720d8e91c21ec95d3dbe8e0d1a4515eb60fa7fa/python/raft-dask/CMakeLists.txt#L33-L35

pentschev:
Started https://github.com/rapidsai/ucxx/pull/500 to resolve this.

pentschev:
And now opened a follow-up to #500 for raft-dask: https://github.com/rapidsai/raft/pull/2798

vyasr:
Fixed in https://github.com/rapidsai/ucxx/pull/507.

PR #507, opened 2025-09-06
Make ucx target global

vyasr:
This change is necessary for the target to be visible when ucxx is built as part of a CPM fetch by another repository (see https://github.com/rapidsai/ucxx/issues/173#issuecomment-3252421137).


pentschev (approved):
LGTM, thanks @vyasr !

PR #516, opened 2025-09-19
Make send/put buffer pointers const

pentschev:
To prevent unintentional modification of send/put buffers and also to make it clear for users that those methods are not supposed modify data/pointers, mark them `const`.

In `tagMultiSend` only the data can be `const`, the pointers themselves need to be modifiable by `std::vector`. Perhaps it would make sense to have a specialized container for `tagMultiSend` in the future.


pentschev:
> I suggest dropping the `void* const` part. As a user, I don’t care if the function modifies the pointer internally, that’s not visible to me anyway.

I know the user doesn't care about that, but this is intentional so that the implementation cannot modify it either since the implementation is not supposed to do that.

pentschev:
Thanks Mads!

Issue #435, opened 2025-06-05
Better C++ API for constructors (and methods that return new `ucxx` objects)


pentschev:
# Summary

I'm getting a bit concerned and annoyed with passing parameters to constructors. Take for example [`ucxx::createRequestTag`](https://github.com/rapidsai/ucxx/blob/c50c9452b1204c152aff686df54896324e71dce9/cpp/include/ucxx/request_tag.h#L87-L92):

```c++
[[nodiscard]] std::shared_ptr<RequestTag> createRequestTag(
  std::shared_ptr<Component> endpointOrWorker,
  const std::variant<data::TagSend, data::TagReceive> requestData,
  const bool enablePythonFuture,
  RequestCallbackUserFunction callbackFunction,
  RequestCallbackUserData callbackData);
 ```

The above is generally not called directly by the user, and the user normally would call [`ucxx::Endpoint::tag{Send,Recv}(...)`](https://github.com/rapidsai/ucxx/blob/c50c9452b1204c152aff686df54896324e71dce9/cpp/include/ucxx/endpoint.h#L584-L645) as a proxy, so we would have to implement something similar for those calls just like to the direct constructors. For simplicity, I'll focus solely on the constructor `createRequestTag` here.

First, the constructor should have defined defaults for `enablePythonFuture`/`callbackFunction`/`callbackData` like in the [private constructor](https://github.com/rapidsai/ucxx/blob/c50c9452b1204c152aff686df54896324e71dce9/cpp/include/ucxx/request_tag.h#L58-L60) and [`tagSend`](https://github.com/rapidsai/ucxx/blob/c50c9452b1204c152aff686df54896324e71dce9/cpp/include/ucxx/endpoint.h#L610-L612)/[`tagRecv`](https://github.com/rapidsai/ucxx/blob/c50c9452b1204c152aff686df54896324e71dce9/cpp/include/ucxx/endpoint.h#L643-L645), which have been missed and is a minor issue, but still an inconsistency.

I'll soon need to introduce another parameter to the requests allowing user to query for UCP debug information for each request, and thus another constructor argument is necessary. What I would prefer to have is some API where users don't have to remember order of arguments and is less prone to API/ABI breakage and to avoid introducing yet another argument. For that, I came up with 3 suggestions below.

## Suggestion 1: Using a `struct`

This is the simplest, and would look something as follows:

```c++
struct RequestTagOptions {
  std::shared_ptr<Component> endpointOrWorker;
[...]

PR #544, opened 2025-11-06
Worker builder

pentschev:
Continuation of #435 and follow-up to #535, adds a new `ucxx::WorkerBuilder` to be used for `ucxx::Worker` construction allowing to specify only the options desired.

For now, this change is backwards-compatible but we may want to remove the constructors at some point and break the API once to do that (to be discussed).

The Python implementation is not adapted to the new constructor because it uses `ucxx::python::Worker` specialization of `ucxx::Worker` that will be covered in a follow-up PR.


jakirkham (approved):

bdice (approved):
Approving CMake

PR #545, opened 2025-11-07
Bump ruff version, use ruff-format instead of black, isort

KyleFromNVIDIA:
Also start linting `__init__.py` and ignore its star import of `ucxx.core`.

Issue: https://github.com/rapidsai/build-planning/issues/130

### DIFF
--- a/python/distributed-ucxx/distributed_ucxx/__init__.py
+++ b/python/distributed-ucxx/distributed_ucxx/__init__.py
@@ -2,9 +2,16 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 from .ucxx import (
-    UCXXBackend as UCXXBackend,
-    UCXXBackendLegacyPrefix as UCXXBackendLegacyPrefix,
+    UCXXBackend,
+    UCXXBackendLegacyPrefix,
 )
 
 
-from ._version import __git_commit__ as __git_commit__, __version__ as __version__
+from ._version import __git_commit__, __version__
+
+__all__ = [
+    "UCXXBackend",
+    "UCXXBackendLegacyPrefix",
+    "__git_commit__",
+    "__version__",
+]
--- a/python/ucxx/ucxx/__init__.py
+++ b/python/ucxx/ucxx/__init__.py
@@ -28,10 +28,10 @@ if "UCX_MEMTYPE_CACHE" not in os.environ:
     logger.debug("Setting env UCX_MEMTYPE_CACHE=n, which is required by UCX")
     os.environ["UCX_MEMTYPE_CACHE"] = "n"
 
-from . import exceptions as exceptions, types as types, testing as testing  # noqa: E402
-from ._lib import libucxx as libucxx  # noqa: E402
+from . import exceptions, types, testing  # noqa: E402
+from ._lib import libucxx  # noqa: E402
 from .core import *  # noqa: E402, F403
-from .utils import get_address as get_address, get_ucxpy_logger  # noqa: E402
+from .utils import get_address, get_ucxpy_logger  # noqa: E402
 
 try:
     import pynvml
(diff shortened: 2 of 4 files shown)


python/pyproject.toml
```diff
@@ -1,59 +0,0 @@
...
-[tool.ruff]
-select = ["E", "F", "W"]
-ignore = [
-    # whitespace before :
-    "E203",
-]
-fixable = ["ALL"]
-exclude = [
-    # TODO: Remove this in a follow-up where we fix __all__.
-    "__init__.py",
```
KyleFromNVIDIA: Requesting review from @charlesbluca since this PR finally fixes `__all__`.

python/distributed-ucxx/distributed_ucxx/__init__.py
```diff
@@ -2,9 +2,9 @@
 # SPDX-License-Identifier: BSD-3-Clause
 
 from .ucxx import (
-    UCXXBackend,
-    UCXXBackendLegacyPrefix,
-)  # noqa: F401
+    UCXXBackend as UCXXBackend,
+    UCXXBackendLegacyPrefix as UCXXBackendLegacyPrefix,
```
pentschev: This is really weird to me, why do we need to make them an alias with same name?

pyproject.toml
pentschev: Why is this moved to the root instead of `python/`? Is there a practical reason for that? Typically it looks more sensible to me that this is inside `python/`, where its immediate subdirectories are the projects that compose what `pyproject.toml` is handling.

python/ucxx/ucxx/__init__.py
```diff
@@ -27,17 +28,17 @@
     logger.debug("Setting env UCX_MEMTYPE_CACHE=n, which is required by UCX")
     os.environ["UCX_MEMTYPE_CACHE"] = "n"
 
-from . import exceptions, types, testing  # noqa
-from ._lib import libucxx  # type: ignore
-from .core import *  # noqa
-from .utils import get_address, get_ucxpy_logger  # noqa
+from . import exceptions as exceptions, types as types, testing as testing  # noqa: E402
```
pentschev: And again aliases with same name in this file.

python/ucxx/ucxx/__init__.py
```diff
@@ -27,17 +28,17 @@
...
+from ._lib import libucxx as libucxx  # noqa: E402
+from .core import *  # noqa: E402, F403
+from .utils import get_address as get_address, get_ucxpy_logger  # noqa: E402
```
pentschev: And then some are aliases (`get_address`), others aren't (`get_ucxpy_logger`), why?

pentschev (changes requested):

python/distributed-ucxx/distributed_ucxx/ucxx.py
```diff
@@ -433,13 +434,6 @@ def __init__(  # type: ignore[no-untyped-def]
 
         weakref.finalize(self, _finalizer, ep, resource_id)
 
-    def abort(self):
-        self._closed = True
-        if self._ep is not None:
-            self._ep.abort()
-            self._ep = None
-            _deregister_dask_resource(self._resource_id)
-
```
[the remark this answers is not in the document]
KyleFromNVIDIA: It's already defined from lines 636 to 641. Ruff was complaining about a duplicate definition. Both definitions were exactly the same, so I picked this one to remove.

pyproject.toml
KyleFromNVIDIA: Most of RAPIDS defines its tooling configuration in a `pyproject.toml` at the root of the repository rather than in `python`. This follows that convention.

python/ucxx/ucxx/__init__.py
KyleFromNVIDIA: I didn't do it with `get_ucxpy_logger` because it was being used elsewhere in the file and thus wasn't tripping up the linter. In hindsight it would have been better to do it with this one too, but let's resolve the discussion at https://github.com/rapidsai/ucxx/pull/545#discussion_r2505360867 before doing anything with it.

python/distributed-ucxx/distributed_ucxx/__init__.py
pentschev: It seems that `as` has the same effect as listing them in `__all__`, is that correct? I did not know that. If so, then I'm fine just leave `as` is (pun intended), presuming ruff will make the same suggestion (i.e., to use `as`) if/when we have the same sort of case occurring in the future, thus to prevent us from using `as` in half of the code base and `__all__` in the other half.

pentschev: Good point, +1 to Bradley's suggestion.

PR #559, opened 2025-12-09
Use static linkage for CUDA runtime

bdice:
## Summary
- Remove `cuda-cudart` from run requirements in `ucxx-tests` conda recipe

With static linking of the CUDA runtime, the runtime is embedded in the binaries and the `cuda-cudart` package is not needed at runtime.

Part of https://github.com/rapidsai/build-planning/issues/235

### DIFF
--- a/conda/environments/all_cuda-129_arch-aarch64.yaml
+++ b/conda/environments/all_cuda-129_arch-aarch64.yaml
@@ -10,7 +10,6 @@ dependencies:
 - c-compiler
 - cloudpickle
 - cmake>=3.26.4,!=3.30.0
-- cuda-cudart-dev
 - cuda-nvcc
 - cuda-version=12.9
 - cudf==26.2.*,>=0.0.0a0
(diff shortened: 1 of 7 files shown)


robertmaynard (changes requested):
I don't think we can do this at the ucxx level.

UCX depends on libcudart https://github.com/search?q=repo%3Aopenucx%2Fucx%20cudart&type=code

vyasr:
I think the real change we need here is to discuss statically linking cudart in UCX. This may also necessitate introducing a new approach in UCX for detecting the availability of the CUDA transport, but I would hope not. I haven't looked in a while, but IIRC the libuct_cuda.so binary is dynamically linked to the driver libcuda.so, and the dlopen of that transport governs whether CUDA support is available. Beyond that, if UCX also uses cudart symbols but statically links to it then we should be on the MVC happy path.

KyleFromNVIDIA:
Superseded by https://github.com/rapidsai/ucxx/pull/647.

PR #566, opened 2026-01-06
Update UCX pin to support 1.20, increase minimum to 1.18

pentschev:
Update UCX minimum to 1.18, building with 1.18.0 on CUDA 12 and keeping 1.19 as minimum for CUDA 13. The pin is also increased to support UCX 1.20, required to enable newer features, such as AWS EFA support.


pentschev:
Thanks Gil and Mads!

PR #567, opened 2026-01-08
wheel builds: react to changes in pip's handling of build constraints

mmccarty:
Contributes to https://github.com/rapidsai/build-planning/issues/242

Modifying `ci/build_wheel.sh` to avoid passing `--build-constraint` and `--no-build-isolation` together which results in an error from `pip`, however we want to keep environment variable `PIP_CONSTRAINT` set unconditionally.

Can be merged after rapidsai/gha-tools#237

### DIFF
--- a/ci/build_wheel.sh
+++ b/ci/build_wheel.sh
@@ -37,6 +37,9 @@ RAPIDS_PIP_WHEEL_ARGS=(
 if [[ "${PIP_NO_BUILD_ISOLATION:-}" != "0" ]]; then
     RAPIDS_PIP_WHEEL_ARGS+=(--build-constraint="${PIP_CONSTRAINT}")
 fi
+# unset that environment variable... it doesn't affect builds as of pip 25.3, and
+# results in an error from 'pip wheel' when set and --build-constraint is also passed
+unset PIP_CONSTRAINT
 
 rapids-pip-retry wheel \
     "${RAPIDS_PIP_WHEEL_ARGS[@]}" \


mmccarty:
Yeah, this is just a test for rapidsai/build-planning#242

@jameslamb - Is this what you had in mind?

jameslamb:
> Interesting error in `wheel-build-libucxx` @jameslamb
> 
> > ERROR: --build-constraint cannot be used with --no-build-isolation.

Ugh that is going to make this a bit more complicated. I was hoping we could just set these environment variables and be done with it, but that error message means we'll need to go case-by-case and do different things based on whether or not build isolation is used.

We really do want `--no-build-isolation` for anything that ends up doing a lot of compiling, because without it we get frequent cache misses from `sccache` (https://github.com/rapidsai/build-planning/issues/108).

So maybe the right mix here is like this?

* leave `rapids-init-pip` unchanged
* add command-line flag `--build-constraint "${PIP_CONSTRAINT}"` to every invocation of `pip wheel` in CI scripts

Kind of unfortunate that it'll require touching every repo, but that shouldn't be *too* bad and should allow the process of cross-PR testing to still work exactly as it did in https://docs.rapids.ai/resources/github-actions/#using-wheel-ci-artifacts-in-other-prs.

Do you agree? If so could you try something like that here?

ci/build_wheel.sh
```diff
@@ -21,11 +21,20 @@ cd "${package_dir}"
 sccache --stop-server 2>/dev/null || true
 
 rapids-logger "Building '${package_name}' wheel"
+
+# Only use --build-constraint when build isolation is enabled.
```
jameslamb:
```suggestion
# Only use --build-constraint when build isolation is enabled.
#
# Passing '--build-constraint' and '--no-build-isolation` together results in an error from 'pip',
# but we want to keep environment variable PIP_CONSTRAINT set unconditionally.
```

Let's expand this comment a bit to explain *why* this code is here.

ci/build_wheel.sh
```diff
@@ -21,11 +21,28 @@ cd "${package_dir}"
...
+#
+# Passing '--build-constraint' and '--no-build-isolation` together results in an error from 'pip',
+# but we want to keep environment variable PIP_CONSTRAINT set unconditionally.
+# PIP_NO_BUILD_ISOLATION=0 means "add --no-build-isolation" (ref: https://github.com/pypa/pip/issues/573
+if [[ "${PIP_NO_BUILD_ISOLATION:-}" != "0" ]]; then
+    RAPIDS_PIP_WHEEL_ARGS+=(--build-constraint="${PIP_CONSTRAINT}")
+fi
+# unset that environment variable... it doesn't affect builds as of pip 25.3, and
+# results in an error from 'pip wheel' when set and --build-constraint is also passed
```
jameslamb:
```suggestion
# unset PIP_CONSTRAINT (set by rapids-init-pip)... it doesn't affect builds as of pip 25.3, and
# results in an error from 'pip wheel' when set and --build-constraint is also passed
```

Let's clarify this a bit like I recommend in https://github.com/rapidsai/rmm/pull/2212#discussion_r2694900425, I think it'll help.

PR #568, opened 2026-01-12
Use limited Python API where possible

pentschev:
Use limited Python API to help address https://github.com/rapidsai/build-planning/issues/42.


cpp/python/src/future.cpp
```diff
@@ -95,10 +95,14 @@ PyObject* create_python_future()
   future_object = get_asyncio_future_object();
   if (future_object == NULL) { goto finish; }
   if (!PyCallable_Check(future_object)) {
+    PyObject* asyncio_encoded = PyUnicode_AsEncodedString(asyncio_str, "utf-8", "strict");
+    PyObject* future_encoded  = PyUnicode_AsEncodedString(future_str, "utf-8", "strict");
     PyErr_Format(PyExc_RuntimeError,
                  "%s.%s is not callable.",
```
da-woods: You'd probably be better just doing

```
"%U.%U is not callable.", asyncio_str, future_str);
```

rather than creating two new bytes objects.

cpp/python/src/future.cpp
```diff
@@ -215,9 +220,20 @@ PyObject* create_python_future_with_event_loop(PyObject* event_loop)
     goto finish;
   }
 
-  result = PyObject_CallMethodObjArgs(event_loop, create_future_str, NULL);
```
da-woods: CallMethodObjArgs is in the Limited API

cpp/python/src/future.cpp
```diff
@@ -246,25 +263,44 @@ PyObject* future_set_result_with_event_loop(PyObject* event_loop, PyObject* futu
...
+    PyObject* future_repr = PyObject_Repr(future);
+    PyObject* future_encoded =
+      future_repr ? PyUnicode_AsEncodedString(future_repr, "utf-8", "strict") : NULL;
+    PyObject* set_result_encoded = PyUnicode_AsEncodedString(set_result_str, "utf-8", "strict");
     PyErr_Format(PyExc_RuntimeError,
                  "%s.%s is not callable.",
-                 PyUnicode_1BYTE_DATA(future),
-                 PyUnicode_1BYTE_DATA(set_result_str));
+                 future_encoded ? PyBytes_AsString(future_encoded) : "<unknown>",
+                 set_result_encoded ? PyBytes_AsString(set_result_encoded) : "<unknown>");
```
jakirkham:
```suggestion
                 future, set_result_str);
```

cpp/python/src/future.cpp
```diff
@@ -296,10 +333,18 @@ PyObject* future_set_exception_with_event_loop(PyObject* event_loop,
     goto finish;
   }
   if (!PyCallable_Check(set_exception_callable)) {
+    PyObject* future_repr = PyObject_Repr(future);
+    PyObject* future_encoded =
+      future_repr ? PyUnicode_AsEncodedString(future_repr, "utf-8", "strict") : NULL;
+    PyObject* set_exception_encoded =
+      PyUnicode_AsEncodedString(set_exception_str, "utf-8", "strict");
```
jakirkham:
```suggestion
```

cpp/python/src/future.cpp
```diff
@@ -296,10 +333,18 @@ PyObject* future_set_exception_with_event_loop(PyObject* event_loop,
...
     PyErr_Format(PyExc_RuntimeError,
                  "%s.%s is not callable.",
```
jakirkham:
```suggestion
    PyErr_Format(PyExc_RuntimeError,
                 "%R.%U is not callable.",
```

cpp/python/src/future.cpp
```diff
@@ -296,10 +333,18 @@ PyObject* future_set_exception_with_event_loop(PyObject* event_loop,
...
-                 PyUnicode_1BYTE_DATA(future),
-                 PyUnicode_1BYTE_DATA(set_exception_str));
+                 future_encoded ? PyBytes_AsString(future_encoded) : "<unknown>",
+                 set_exception_encoded ? PyBytes_AsString(set_exception_encoded) : "<unknown>");
```
jakirkham:
```suggestion
                 future, set_result_str);
```

cpp/python/src/future.cpp
```diff
@@ -246,25 +263,44 @@ PyObject* future_set_result_with_event_loop(PyObject* event_loop, PyObject* futu
...
+                 future_encoded ? PyBytes_AsString(future_encoded) : "<unknown>",
+                 set_result_encoded ? PyBytes_AsString(set_result_encoded) : "<unknown>");
+    Py_XDECREF(future_repr);
+    Py_XDECREF(future_encoded);
+    Py_XDECREF(set_result_encoded);
     goto finish;
   }
 
-  result = PyObject_CallMethodObjArgs(
-    event_loop, call_soon_threadsafe_str, set_result_callable, value, NULL);
```
jakirkham: As David noted before, `PyObject_CallMethodObjArgs` is in the Python 3.11 Limited API and Stable ABI

Let's see how the discussion in the thread above unfolds: https://github.com/rapidsai/ucxx/pull/568#discussion_r2688299724

Based on that we can update this as needed

jakirkham (commented):
Thanks Peter! 🙏

Appreciate you getting this started. Also thanks Vyas and Gil for flagging these. And thanks David for sharing your feedback

Have added code suggestions building off of David's feedback to provide more context and ease integration. As GitHub suggestions doesn't allow commenting on removed lines, the suggestions are broken up a bit. Apologies that these can't fit in a single comment

There is one change we are unsure is needed. Would welcome more context and feedback in this thread: https://github.com/rapidsai/ucxx/pull/568#discussion_r2688299724

cpp/python/src/future.cpp
vyasr: This failed for me when I first attempted to get ucxx compiling with the limited API, which was over a year ago. I no longer have any traceback to share. It's certainly possible that I attempted compilation incorrectly in some way, or that I misinterpreted something.

pentschev:
Thanks @da-woods and @jakirkham for the reviews. I agree just passing the proper data format into the string without an explicit conversion is much better, thanks for pointing that out. Also I was misinformed about `PyObject_CallMethodObjArgs` not being part of the limited API and I overlooked checking it, so that was entirely unnecessary and is now reverted.

Now this PR only contains the string changes and is very slim, simultaneously making this part of the code compatible with the limited API to the best of my knowledge. Would appreciate if you could have another look.

da-woods:
Looks fine to me (as someone who doesn't know any of the surrounding code-base).

jakirkham (approved):
Thank you Peter for all your work here! Also thank you everyone who helped review 🙏

PR #581, opened 2026-01-30
Don't hard-code branch name in `RAPIDS.cmake`

KyleFromNVIDIA:
The branch name can be overridden by setting `rapids-cmake-branch`.

### DIFF
--- a/ci/release/update-version.sh
+++ b/ci/release/update-version.sh
@@ -148,13 +148,6 @@ for FILE in python/*/pyproject.toml; do
   done
 done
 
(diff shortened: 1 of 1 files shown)


jakirkham:
Thanks Kyle! 🙏

Was the branch actually hard-coded anywhere?

Asking as this removes the logic to update the branch. Though would have thought that there are some references to the branch in the project that also needed updating. However not seeing the latter in this PR. Am I missing something?

KyleFromNVIDIA:
It's hard-coded in `RAPIDS.cmake`, and `update-version.sh` was changing it. We don't need to change it, since it can be overridden.

PR #582, opened 2026-02-02
Fixes for C++ to Python future notifications

pentschev:
Fix missing reference counting for asyncio event loop in Python futures, `_asyncioEventLoop` pointer in `ucxx::python::Future` was stored without incrementing its reference count, creating a potential use-after-free bug if the Python garbage collector deallocates the event loop while the `Future` object still holds a pointer to it.

Added checks in `future_set_result_with_event_loop()` and `future_set_exception_with_event_loop()` to validate that the event loop pointer is not `NULL` and points to a valid Python object before attempting to call methods on it. During shutdown or in race conditions, the event loop may be deallocated while futures are still being processed. Without these checks, accessing a deallocated event loop can cause segmentation faults.


pentschev:
Thanks Mads!

PR #584, opened 2026-02-05
Test shared-workflows branch add-rtxpro6000

bdice:
Test PR for https://github.com/rapidsai/shared-workflows/pull/499.


bdice:
None of these failures seem specific to rtxpro6000 so I think this repo is sufficiently ready to turn on the new runners.

PR #616, opened 2026-03-31
Refactor `InflightRequests`

pentschev:
Refactor `InflightRequests` to replace the previous triple-mutex scheme (`std::recursive_mutex` + two `std::mutex` inside a heap-allocated `TrackedRequests` struct) and dual `std::map` containers with a single `std::mutex` and two `std::unordered_set` members. This eliminates the complex `try_lock` retry loop in `remove()` that was needed to handle the cancel-during-remove race: the new `cancelAll()` moves all requests out of the inflight set before calling `cancel()`, so any subsequent `remove()` from the completion callback is naturally a no-op. The `TrackedRequests` type is simplified from a mutex-bearing struct wrapped in `std::unique_ptr` to a plain value type holding two vectors, used solely as a transfer vehicle for `release()`/`merge()` operations. The result is significantly easier to follow: each public method acquires a single non-recursive mutex and performs one container operation, with no nested locking, no recursive mutex overhead, and no retry loops. The switch from `std::map` (red-black tree with per-node allocation and O(log n) operations) to `std::unordered_set` (O(1) amortized insert/erase) should also scalability for workloads with large numbers of concurrent inflight requests.

### DIFF
--- a/cpp/src/inflight_requests.cpp
+++ b/cpp/src/inflight_requests.cpp
@@ -62,7 +62,7 @@ size_t InflightRequests::cancelAll()
     std::lock_guard<std::mutex> lock(_mutex);
     for (auto& r : toCancel) {
       if (r && r->getStatus() == UCS_INPROGRESS)
-        _canceling.insert(std::move(const_cast<std::shared_ptr<Request>&>(r)));
+        _canceling.insert(r);
     }
   }
 
--- a/cpp/src/inflight_requests.cpp
+++ b/cpp/src/inflight_requests.cpp
@@ -61,8 +61,7 @@ size_t InflightRequests::cancelAll()
   {
     std::lock_guard<std::mutex> lock(_mutex);
     for (auto& r : toCancel) {
-      if (r && r->getStatus() == UCS_INPROGRESS)
-        _canceling.insert(r);
+      if (r && r->getStatus() == UCS_INPROGRESS) _canceling.insert(r);
     }
   }
 
--- a/cpp/src/inflight_requests.cpp
+++ b/cpp/src/inflight_requests.cpp
@@ -2,11 +2,7 @@
  * SPDX-FileCopyrightText: Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES.
  * SPDX-License-Identifier: BSD-3-Clause
  */
-#include <algorithm>
-#include <cstdlib>
-#include <cstring>
 #include <memory>
-#include <string>
 #include <utility>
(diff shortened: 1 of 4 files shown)


cpp/src/worker.cpp
```diff
@@ -538,13 +538,13 @@ size_t Worker::cancelInflightRequests(uint64_t period, uint64_t maxAttempts)
 
   if (inflightRequestsToCancel->getCancelingSize() > 0) {
     std::lock_guard<std::mutex> lock(_inflightRequestsMutex);
-    _inflightRequestsToCancel->merge(inflightRequestsToCancel->release());
+    _inflightRequestsToCancel->merge(std::move(inflightRequestsToCancel->release()));
```
madsbk:
```suggestion
    _inflightRequestsToCancel->merge(inflightRequestsToCancel->release());
```

cpp/src/inflight_requests.cpp
```diff
@@ -14,156 +16,89 @@ InflightRequests::~InflightRequests() { cancelAll(); }
...
-  dropCanceled();
-  size_t cancelingSize = 0;
   {
-    std::scoped_lock localLock{_mutex};
-    std::scoped_lock lock{_trackedRequests->_cancelMutex};
-    cancelingSize = _trackedRequests->_canceling.size();
+    std::lock_guard<std::mutex> lock(_mutex);
+    for (auto& r : toCancel) {
+      if (r && r->getStatus() == UCS_INPROGRESS)
+        _canceling.insert(std::move(const_cast<std::shared_ptr<Request>&>(r)));
```
madsbk: `const_cast` + `std::move` is UB.
```suggestion
        _canceling.insert(r);
```

madsbk (changes requested):
Nice simplification!

cpp/src/inflight_requests.cpp
pentschev: Nice catch!

pentschev (commented):
Thanks Mads!

pentschev:
Oh, I thought you had approved already, could you do so if you're satisfied @madsbk ?

madsbk (approved):

madsbk:
> Oh, I thought you had approved already, could you do so if you're satisfied @madsbk ?

Heh, I did as well :)

pentschev:
> > Oh, I thought you had approved already, could you do so if you're satisfied @madsbk ?
> 
> Heh, I did as well :)

Ha! No worries, thanks for the review! :)

PR #617, opened 2026-03-31
Support loopback (single-process) mode in `ucxx_perftest`

pentschev:
UCX may follow a different path when communicating within the same process. One example of that is the CUDA transport used should be `cuda_copy` instead of `cuda_ipc`. Therefore the ability to run in loopback (single-process) mode allows testing and benchmarking paths that cannot otherwise be tested.


madsbk (approved):
LGTM

PR #618, opened 2026-03-31
Support for RMM memory resources in `ucxx_perftest`

pentschev:
In addition to host and CUDA memory types introduce support for memory resources from RMM. This helps users benchmark more closely to their real use cases.

### DIFF
--- a/cpp/benchmarks/include/buffer_interface.hpp
+++ b/cpp/benchmarks/include/buffer_interface.hpp
@@ -5,8 +5,8 @@
 #pragma once
 
 #include <algorithm>
-#include <cassert>
 #include <iostream>
+#include <stdexcept>
 #include <memory>
 #include <string>
 #include <unordered_map>
@@ -331,7 +331,8 @@ struct HostBufferInterface : public BufferInterface {
   void verifyResults(size_t messageSize) override
   {
     for (size_t j = 0; j < (*bufferMap)[DirectionType::Send].size(); ++j)
-      assert((*bufferMap)[DirectionType::Recv][j] == (*bufferMap)[DirectionType::Send][j]);
+      if ((*bufferMap)[DirectionType::Recv][j] != (*bufferMap)[DirectionType::Send][j])
+        throw std::runtime_error("Host data verification failed at byte " + std::to_string(j));
   }
 
   // Static factory method to create host buffer interface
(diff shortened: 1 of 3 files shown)


cpp/benchmarks/include/buffer_interface.hpp
```diff
@@ -631,3 +682,192 @@ std::unique_ptr<CudaBufferInterfaceBase> CudaBufferInterfaceBase::allocateBuffer
...
+                                     recv.data(),
+                                     messageSize,
+                                     cudaMemcpyDeviceToHost,
+                                     recv.stream()),
+                     "RMM recv data copy for verification");
+  CUDA_EXIT_ON_ERROR(cudaStreamSynchronize(send.stream()),
+                     "RMM send stream synchronization for verification");
+  CUDA_EXIT_ON_ERROR(cudaStreamSynchronize(recv.stream()),
+                     "RMM recv stream synchronization for verification");
+  for (std::size_t j = 0; j < messageSize; ++j) assert(recvData[j] == sendData[j]);
```
madsbk: Is it intentional that the data is only checked in non-debug builds?

madsbk (approved):
LGTM

cpp/benchmarks/include/buffer_interface.hpp
pentschev: I think you mean "only checked in _debug_ builds (because of `assert()`)", right? But no, this was total oversight of my part, not just in RMM but also the existing validation for other types were using `assert()` and for the wireup in `perftest.cpp`. Fixed this in https://github.com/rapidsai/ucxx/pull/618/commits/870fc4318fd3c76158020670327df6beed633742 and everything should `throw` now. Thanks for catching that!

madsbk (approved):

pentschev:
Thanks Mads!

wence- (approved):

bdice (approved):
This means I'll need to add a PR to ucxx for the RMM CCCL memory resource migration -- I can tackle that after this merges.

pentschev:
> This means I'll need to add a PR to ucxx for the RMM CCCL memory resource migration -- I can tackle that after this merges.

Maybe that was true before as well? Note we already had [some limited use of RMM in UCXX](https://github.com/rapidsai/ucxx/blob/main/cpp/src/buffer.cpp).

PR #621, opened 2026-04-08
Fix error messages in a couple `__init__` methods

jakirkham:
A few classes raise in `__init__` as other methods are encouraged for their construction. However those methods all referenced `UCXListener`. IIUC they should reference the class they are creating. Guessing this just happened due to copy-paste. So have tried to fix these based on my understanding. Happy to change if needed


pentschev (approved):
Good catch. Thanks John!

jakirkham:
Thanks Peter! 🙏

PR #622, opened 2026-04-09
Add `getStringView` & deprecate `getString`

jakirkham:
Follow up to PR: https://github.com/rapidsai/ucxx/pull/392

Restore the original signature of `getString` to avoid a breaking change. Instead deprecate it with a planned removal in UCXX 0.51. Alongside it add `getStringView` as its replacement. Update internal code to instead use `getStringView`.

This provides a deprecation cycle for external users to move to the new behavior (if they are affected).

### DIFF
--- a/cpp/include/ucxx/address.h
+++ b/cpp/include/ucxx/address.h
@@ -126,9 +126,10 @@ class Address : public Component {
    *
    * @returns The underlying `ucp_address_t` handle.
    */
+  [[nodiscard]]
   [[deprecated(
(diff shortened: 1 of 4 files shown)


pentschev (approved):
Personally I think this is not absolutely necessary, see https://github.com/rapidsai/ucxx/pull/392#issuecomment-4212411270 with my reasoning. Regardless, I'm not opposed either, as this is good practice.

jakirkham:
Thanks Peter! 🙏

Yeah I didn't think so either. Though it was simple to do and it provided us another option if we wanted it

PR #623, opened 2026-04-09
Add `UCXAddress.tobytes` method

jakirkham:
Add a `cpdef` method `tobytes` to `UCXAddress` for efficient Python & Cython/C++ construction of `bytes` representations of `UCXAddress`. This allows for efficient C++ function calls whenever `tobytes` is used as is the case in several `UCXAddress` methods.

Also `tobytes` caches the resulting `bytes` object in `UCXAddress._bytes` for fast retrieval. That way a single `tobytes` call constructs the `bytes` representation and subsequent calls retrieves it for free. Ensure `__bytes__` makes use of `tobytes` along with all other `UCXAddress` methods needing the `bytes` representation.

Plus [`bytes` implementation of `__hash__`]( https://github.com/python/cpython/blob/9a89f1bf1e7bb819fe7240be779c99a84f47ea46/Objects/bytesobject.c#L1570-L1581 ) maintains a cache of the hashed value. Meaning that when `UCXAddress.__hash__` calls `hash` on the cached `bytes` object, the cached hash value for `bytes` can be used as well. So `hash`ing of `UCXAddress` benefits not only from caching the `bytes` object, but also the hash value. Meaning subsequent `hash` calls can retrieve the cached hash value.

In summary this should cutdown on the cost of repeated calls to things like `bytes`, `hash`, etc. on `UCXAddress` objects.

### DIFF
--- a/python/ucxx/ucxx/_lib/libucxx.pxd
+++ b/python/ucxx/ucxx/_lib/libucxx.pxd
@@ -43,8 +43,11 @@ cdef class UCXAddress:
         size_t _length
         ucp_address_t *_handle
         string_view _string
+        bytes _bytes
 
     cdef shared_ptr[Address] get_ucxx_shared_ptr(self) nogil
+
+    cpdef bytes tobytes()
 
 
 cdef class UCXWorker:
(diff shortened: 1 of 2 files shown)


pentschev (approved):
Makes sense, thanks John!

jakirkham:
Thanks Peter! 🙏

PR #624, opened 2026-04-09
Add missing `cimport` of `string_view`

jakirkham:
This file makes use of `string_view` on this line (and `string` previously)

https://github.com/rapidsai/ucxx/blob/57cd86e8cee06caffca8a563912784d1ea12d298/python/ucxx/ucxx/_lib/libucxx.pxd#L45

However neither was `cimport`ed.

This likely worked as `ucxx_api` had `cimport`ed both

https://github.com/rapidsai/ucxx/blob/57cd86e8cee06caffca8a563912784d1ea12d298/python/ucxx/ucxx/_lib/ucxx_api.pxd#L13-L14

And everything from `ucxx_api` is `cimport`ed into `libucxx.pxd`

https://github.com/rapidsai/ucxx/blob/57cd86e8cee06caffca8a563912784d1ea12d298/python/ucxx/ucxx/_lib/libucxx.pxd#L8

Nevertheless `string_view` should be `cimport`ed where it is used. So this adds the missing `cimport`.

### DIFF
--- a/python/ucxx/ucxx/_lib/libucxx.pxd
+++ b/python/ucxx/ucxx/_lib/libucxx.pxd
@@ -3,7 +3,9 @@
 
 
 from libc.stdint cimport uint64_t, uintptr_t
+
 from libcpp.memory cimport shared_ptr, unique_ptr
+from libcpp.string_view cimport string_view
 
 from .ucxx_api cimport *
 


pentschev (approved):
Thanks John!

jakirkham:
Thanks Peter! 🙏

PR #625, opened 2026-04-10
Drop intermediates in UCXX worker test

jakirkham:
Drops intermediate `bytearray` objects used to construct messages in favor of more inlined approaches like [`struct.pack_into`]( https://docs.python.org/3/library/struct.html#struct.pack_into ). By doing this there is no need to explicitly figure out padding (or constructing intermediates to hold space for padding). Instead construct frames with the needed size and write directly into them with `struct.pack_into` using the values unchanged.

Further as NumPy arrays are already used here. Make more consistent use of them for buffers of bytes (as opposed to `bytearray`). This effectively replaces `calloc` with `malloc` dropping unneeded initialization steps which will be overwritten by data later.

Lastly tidy the format string using [f-strings]( https://docs.python.org/3/reference/lexical_analysis.html#f-strings ).

### DIFF
--- a/python/ucxx/ucxx/_lib_async/tests/test_from_worker_address.py
+++ b/python/ucxx/ucxx/_lib_async/tests/test_from_worker_address.py
@@ -28,7 +28,7 @@ def _test_from_worker_address_server(queue, timeout):
         await ucxx.recv(address_size, tag=0)
 
         # Receive address buffer on tag 0 and create UCXAddress from it
-        address_size = np.empty(address_size[0], dtype=np.uint8)
+        remote_address = np.empty(address_size[0], dtype=np.uint8)
         await ucxx.recv(remote_address, tag=0)
         remote_address = ucxx.get_ucx_address_from_buffer(remote_address)
 
--- a/python/ucxx/ucxx/_lib_async/tests/test_from_worker_address.py
+++ b/python/ucxx/ucxx/_lib_async/tests/test_from_worker_address.py
@@ -134,7 +134,7 @@ def _pack_address_and_tag(address, recv_tag, send_tag):
         recv_tag,  # Recv Tag
         send_tag,  # Send Tag
         address.length,  # Address buffer length
-        address,  # Address buffer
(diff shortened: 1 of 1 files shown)


python/ucxx/ucxx/_lib_async/tests/test_from_worker_address.py
```diff
@@ -111,33 +111,30 @@ def _get_address_info(address=None):
...
         address_info["fixed_size_address_buffer_fmt"],
+        fixed_size_address_packed,  # Buffer to fill
+        0,  # Starting Offset
         recv_tag,  # Recv Tag
         send_tag,  # Send Tag
         address.length,  # Address buffer length
-        (
-            bytearray(address) + bytearray(address_info["padding_length"])
-        ),  # Address buffer + padding
+        bytes(address),  # Address buffer
```
jakirkham, line 137: As [`struct.pack` using format `s` only accepts `bytes` or `bytearray`]( https://github.com/python/cpython/blob/9a89f1bf1e7bb819fe7240be779c99a84f47ea46/Modules/_struct.c#L1936-L1941 ), for now convert `address` to `bytes`. This is a reasonably efficient path after PR: https://github.com/rapidsai/ucxx/pull/623

Though ideally this copy would be avoided. That may be better done with follow up work

pentschev (approved):
Thanks John!

python/ucxx/ucxx/_lib_async/tests/test_from_worker_address.py
jakirkham: Raised as a CPython bug: https://github.com/python/cpython/issues/148529

jakirkham: With some follow up work was able to avoid the need for this copy in PR: https://github.com/rapidsai/ucxx/pull/628

jakirkham:
Thanks Peter! 🙏

Have made some further improvements on this in PR: https://github.com/rapidsai/ucxx/pull/628

PR #628, opened 2026-04-13
Simplify header generation in UCXX worker test

jakirkham:
Follow up to PR: https://github.com/rapidsai/ucxx/pull/625

Move from a function generating constants used in message headers to a couple global constants used in message packing/unpacking functions. This reduces the lines of code. Also it benefits from building things like a `struct.Struct` object once and reusing it for serializing message header metadata.

Additionally this ensures zero-copy packing of `UCXXAddress` when packing/unpacking it. Thus eliminating some intermediate `bytes` objects that are otherwise created.

Lastly this avoids explicit writes that `struct.pack` did to fill in the padded bytes with `\0`. Similarly `struct.unpack` included both the address and the pad values, which it copied to a large `bytes` object containing both. Given the padding is only there to ensure the frame is of a specific size, there is no need to fill out the pad values or copy them around. As a result this change now ensures no extra effort is put into the pad values.

### DIFF
--- a/python/ucxx/ucxx/_lib_async/tests/test_from_worker_address.py
+++ b/python/ucxx/ucxx/_lib_async/tests/test_from_worker_address.py
@@ -17,6 +17,14 @@ from ucxx.testing import join_processes, terminate_process
 mp = mp.get_context("spawn")
 
 
+# Fixed frame size
+FRAME_SIZE = 10000
+
+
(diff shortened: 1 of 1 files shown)


pentschev (approved):
LGTM, thanks John.

jakirkham:
Thanks Peter! 🙏

PR #629, opened 2026-04-14
Fix decrementing active clients upon exception in listener callback

pentschev:
Wrapped the body of `_listener_handler_coroutine` in `try`/`finally` so that active_clients.dec(ident) and del ep always execute, even if:
* `exchange_peer_info` times out or is cancelled
* `endpoint.raise_on_error()` raises
* `func(ep)` (the callback) raises an unexpected exception
* The coroutine is externally cancelled

This ensures the coroutine frame never keeps `ctx` / `ApplicationContext` alive after the coroutine finishes (normally or abnormally).

Removed all four `@pytest.mark.flaky` decorators that were masking this bug in distributed-ucxx tests. With the listener fix in place, the retries should no longer be needed.

### DIFF
--- a/python/ucxx/ucxx/_lib_async/listener.py
+++ b/python/ucxx/ucxx/_lib_async/listener.py
@@ -142,9 +142,6 @@ async def _listener_handler_coroutine(
     ident,
     active_clients,
 ):
-    # def _listener_handler_coroutine(
-    #     conn_request, ctx, func, endpoint_error_handling, ident, active_clients
-    # ):
     # We create the Endpoint in five steps:
     #  1) Create endpoint from conn_request
     #  2) Generate unique IDs to use as tags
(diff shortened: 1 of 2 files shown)


python/ucxx/ucxx/_lib_async/listener.py
```diff
@@ -142,66 +142,66 @@ async def _listener_handler_coroutine(
...
-    del ep
+        # Removing references here to avoid delayed clean up
+        del ctx
+
+        # Finally, we call `func`
+        if inspect.iscoroutinefunction(func):
+            try:
+                await func(ep)
+            except Exception as e:
+                logger.error(f"Uncatched listener callback error {type(e)}: {e}")
```
TomAugspurger: Mostly just an FYI, but there's also [`logger.exception`](https://docs.python.org/3/library/logging.html#logging.Logger.exception) to include the traceback.

TomAugspurger (approved):

python/ucxx/ucxx/_lib_async/listener.py
pentschev: Oh nice, I wasn't aware of that one. Changed to that in https://github.com/rapidsai/ucxx/pull/629/commits/7785c5314c5707d11d3a6d31cfd7f9a2d809b5da .

pentschev:
Thanks Tom!

PR #630, opened 2026-04-14
Ensure distributed-ucxx `read()` aborts endpoint on frame errors

pentschev:
The else block of `read()` had a `try`/`except` `BaseException` that only wrapped the frame recv loop, leaving frame allocation and the `zip()` unpacking unprotected, so when corrupted metadata during shutdown caused a `ValueError` there, the endpoint was never aborted, leaking it and blocking `ucxx.reset()` at teardown. The fix widens that existing `try`/`except` `BaseException` to cover the entire `else` block, ensuring any error during frame processing properly aborts the endpoint and raises `CommClosedError`.

### DIFF
--- a/python/distributed-ucxx/distributed_ucxx/ucxx.py
+++ b/python/distributed-ucxx/distributed_ucxx/ucxx.py
@@ -524,9 +524,7 @@ class UCXX(Comm):
                 # "low-level" exception. The only safe thing to do is to abort.
                 # (See also https://github.com/dask/distributed/pull/6574).
                 self.abort()
-                raise CommClosedError(
-                    f"Connection closed by writer.\nInner exception: {e!r}"
-                )
+                raise CommClosedError("Connection closed by writer") from e
         else:
             try:
                 # Recv meta data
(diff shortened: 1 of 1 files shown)


python/distributed-ucxx/distributed_ucxx/ucxx.py
TomAugspurger: Maybe out of scope, but these days we should be using `raise ... from e` to properly chain the exceptions: https://docs.python.org/3/tutorial/errors.html#exception-chaining

TomAugspurger (approved):

python/distributed-ucxx/distributed_ucxx/ucxx.py
pentschev: It is a bit out-of-scope but I think that's a harmless change with a clear code quality improvement so it should be ok. Done in https://github.com/rapidsai/ucxx/pull/630/commits/c4e061586dafbeab16e8ecc169ca9148eb78edfc .

jakirkham (approved):
Thanks Peter! 🙏

pentschev:
Thanks for the reviews Tom and John.

PR #631, opened 2026-04-14
Do not run `ucxx` tests on changes only to `distributed-ucxx`

pentschev:
Optimize CI to avoid running `ucxx` test suite when changes are made only to `distributed-ucxx` package.


jameslamb (approved):
Nice!

PR #632, opened 2026-04-16
Surface subprocess exceptions in Python tests

pentschev:
Add `run_in_subprocess` to `ucxx.testing`: a thin wrapper for use as a `multiprocessing.Process` target that catches any unhandled exception, serialises its traceback into a caller-supplied `multiprocessing.Queue`, and re-raises so the process still exits with a non-zero code.

Extend `terminate_process` with an optional `error_queue` parameter: when a process exits non-zero and a queue is provided, the stored traceback is included in the raised `RuntimeError` rather than the generic "Process did not exit cleanly" message.

Update all subprocess-based tests to use this pattern so that the root cause of a failure is immediately visible in CI output instead of being silently swallowed by the subprocess boundary.

Add tests for both functions to the testing-utilities test suite.


madsbk (approved):
This is a very good idea :)

pentschev:
Thanks Mads!

PR #633, opened 2026-04-17
Fix subprocess error propagation in `test_benchmark_cluster`

pentschev:
This is a follow-up to https://github.com/rapidsai/ucxx/pull/632 which allows propagating exceptions and errors from the subprocesses back to the main pytest process, exposing necessary information from actual failures. It was an oversight that this adjustment was not done as part of https://github.com/rapidsai/ucxx/pull/632.


wence- (approved):

pentschev:
Thanks Lawrence.

PR #634, opened 2026-04-17
Fix `ApplicationContext` reference cycle and unhandled write exceptions

pentschev:
`UCXListener._cb_data["cb_args"]` held a strong reference to `ApplicationContext` creating a reference cycle that prevented proper cleanup and caused intermittent `ucxx.reset()` failures in the `ucxx_loop` fixture. The fix passes `weakref.ref(self)` in `cb_args` so the cycle never holds `ApplicationContext` alive, while `Listener` retains a direct strong reference (released by `close()`) to preserve the detection contract in `reset()`. `_listener_handler_coroutine` now dereferences the weakref immediately and clears `ctx` in `finally` to avoid holding `ApplicationContext` in cancelled coroutine frames. `UCXXListener.stop()` now calls `ucxx_server.close()` before dropping the reference for deterministic cleanup.

Additionally, `write()` now catches `BaseException` instead of `ucxx.exceptions.UCXError` so that `CancelledError` and `ValueError` mid-write also trigger `abort()` and return `CommClosedError` (consistent with #630), two f-strings in `ActiveClients` error messages had their missing f prefix added, and `gc.collect()` is called before `ucxx.reset()` in the test fixture as a safety net.

### DIFF
--- a/python/ucxx/ucxx/_lib_async/listener.py
+++ b/python/ucxx/ucxx/_lib_async/listener.py
@@ -150,10 +150,6 @@ async def _listener_handler_coroutine(
     #  3) Exchange endpoint info such as tags
     #  4) Setup control receive callback
     #  5) Execute the listener's callback function
-    if isinstance(ctx, weakref.ref):
-        ctx = ctx()
-        if ctx is None:
-            return
     active_clients.inc(ident)
     ep = None
     try:
--- a/python/ucxx/ucxx/_lib_async/listener.py
+++ b/python/ucxx/ucxx/_lib_async/listener.py
@@ -156,10 +156,7 @@ async def _listener_handler_coroutine(
     #  4) Setup control receive callback
     #  5) Execute the listener's callback function
 
-    # Dereference the weakref immediately so the UCXListener's cb_args tuple
-    # does not keep ApplicationContext alive through this coroutine's frame.
     ctx = ctx_weakref()
-    del ctx_weakref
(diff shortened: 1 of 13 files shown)


wence- (approved):

python/ucxx/ucxx/_lib_async/listener.py
```diff
@@ -148,6 +155,17 @@ async def _listener_handler_coroutine(
     #  3) Exchange endpoint info such as tags
     #  4) Setup control receive callback
     #  5) Execute the listener's callback function
+
+    # Dereference the weakref immediately so the UCXListener's cb_args tuple
+    # does not keep ApplicationContext alive through this coroutine's frame.
+    ctx = ctx_weakref()
+    del ctx_weakref
```
wence-: Why do you need to decref `ctx_weakref`. By definition it can't be keeping anything alive.

Also, I don't understand the comment.

pentschev: This was totally AI slop that I missed when committed, it's removed now. I'm still struggling with this PR, it seems I have in fact resolved the original issue (distributed-ucxx failing tests) as I haven't seen it pop up anymore for a while here, but simultaneously the fix seems to have broken the UCXX Python tests, so still trying to fix that and thus so many changes that may seem somewhat random still.

pentschev:
Superseded by #686 .

PR #637, opened 2026-04-22
Add option to set CPU affinity in `ucxx_perftest`

pentschev:
To test certain paths effectively (such as NIC selection on multi-NUMA node systems) it is necessary to set proper CPU affinity. This feature is also included in `ucx_perftest` that should be supported by the `ucxx_perfest` counterpart.


madsbk (approved):

pentschev:
Thanks Mads!

PR #638, opened 2026-04-23
Make builder constructors ABI stable with the pImpl idiom

pentschev:
Use the pImpl idiom as a first step towards #3 , and also to ensure upcoming expansions of constructors to use the builder pattern already follow the same path.

### DIFF
--- a/cpp/include/ucxx/experimental/builder_utils.h
+++ b/cpp/include/ucxx/experimental/builder_utils.h
@@ -1,0 +1,31 @@
+/**
+ * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+#pragma once
+
+#include <memory>
+
+/**
+ * @brief Define pImpl rule-of-five members and the implicit conversion operator for a builder.
+ *
+ * Place this in the `.cpp` file after the `Impl` struct definition. The builder class must
+ * have a `std::unique_ptr<Impl> _impl` member and a `build()` method.
+ *
+ * @param BuilderClass  the fully-qualified builder class name.
+ * @param TargetClass   the type that `build()` returns wrapped in `std::shared_ptr`.
+ */
+#define UCXX_BUILDER_PIMPL_DEFAULTS(BuilderClass, TargetClass)              \
+  BuilderClass::~BuilderClass()                                  = default; \
+  BuilderClass::BuilderClass(BuilderClass&&) noexcept            = default; \
+  BuilderClass& BuilderClass::operator=(BuilderClass&&) noexcept = default; \
+  BuilderClass::BuilderClass(const BuilderClass& other)                     \
+    : _impl(std::make_unique<Impl>(*other._impl))                           \
+  {                                                                         \
+  }                                                                         \
+  BuilderClass& BuilderClass::operator=(const BuilderClass& other)          \
+  {                                                                         \
+    if (this != &other) _impl = std::make_unique<Impl>(*other._impl);       \
+    return *this;                                                           \
+  }                                                                         \
+  BuilderClass::operator std::shared_ptr<TargetClass>() const { return build(); }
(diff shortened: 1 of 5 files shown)


cpp/src/experimental/worker_builder.cpp
```diff
@@ -1,39 +1,46 @@
...
+struct WorkerBuilder::Impl {
+  std::shared_ptr<Context> context;
+  bool enableDelayedSubmission{false};
+  bool enableFuture{false};
+};
+
+WorkerBuilder::WorkerBuilder(std::shared_ptr<Context> context) : _impl(std::make_unique<Impl>())
+{
+  _impl->context = std::move(context);
+}
```
nirandaperera: Nit. I think we could do this. Worst case it may need a ctr in `Impl`
```suggestion
WorkerBuilder::WorkerBuilder(std::shared_ptr<Context> context) : _impl(std::make_unique<Impl>{std::move(context)})
{}
```

cpp/src/experimental/context_builder.cpp
```diff
@@ -1,33 +1,39 @@
...
+  ConfigMap configMap{};
+  uint64_t featureFlags;
+};
 
-ContextBuilder& ContextBuilder::configMap(ConfigMap configMap)
+ContextBuilder::ContextBuilder(uint64_t featureFlags) : _impl(std::make_unique<Impl>())
 {
-  _configMap = std::move(configMap);
-  return *this;
+  _impl->featureFlags = featureFlags;
```
nirandaperera: Nit. same here. this can be pushed up to `make_unique` (I think)

cpp/include/ucxx/experimental/builder_utils.h
```diff
@@ -0,0 +1,31 @@
...
+  BuilderClass::BuilderClass(const BuilderClass& other)                     \
+    : _impl(std::make_unique<Impl>(*other._impl))                           \
+  {                                                                         \
+  }                                                                         \
+  BuilderClass& BuilderClass::operator=(const BuilderClass& other)          \
+  {                                                                         \
+    if (this != &other) _impl = std::make_unique<Impl>(*other._impl);       \
+    return *this;                                                           \
+  }                                                                         \
+  BuilderClass::operator std::shared_ptr<TargetClass>() const { return build(); }
```
nirandaperera, line 31: I feel like this could be a CRTP mixin. I'd avoid macros as much as possible because its impossible to read (at least for me) 😅

nirandaperera (changes requested):
some minor cosmetic comments

cpp/src/experimental/worker_builder.cpp
pentschev: Good idea, required also adding constructors. Done in https://github.com/rapidsai/ucxx/pull/638/commits/1158c548155501d5448a9a2cc02ed240b2860ac0

cpp/src/experimental/context_builder.cpp
pentschev: Also done in https://github.com/rapidsai/ucxx/pull/638/commits/1158c548155501d5448a9a2cc02ed240b2860ac0

cpp/include/ucxx/experimental/builder_utils.h
pentschev: I don't disagree macros are hard to read, but I think this one in particular is as easy as it can get, since it does essentially nothing complex other than replace `BuilderClass` with `TargetClass`.

As for your CRTP suggestion, I think it doesn't work for pImpl. First, a template needs to live in a header, and pImpl's entire point is that `Impl` is only complete in the .cpp file. A CRTP base can't generate destructors or copy/move operations for a pImpl class because it can't see the complete `Impl` type. The `unique_ptr<Impl>` destruction also needs a complete type.

With that said, the alternative here is to have all 6 methods here written in every .cpp file, I have previously considered that but it feels like unnecessary boilerplate. Currently this only exists in two files, but I'll soon extend the builder pattern to other classes, immediately following with the `Request` classes, which are going to add these 6 methods to 7 new classes, and thus we will already have these same entries 9 times. If you feel like this is terrible, I can remove the macro in favor of explicit instantiation, but I feel like this is one of the few cases where a macro is genuinely the right tool for the job.

nirandaperera (approved):

pentschev:
Thanks for the review, Niranda!

PR #639, opened 2026-04-27
Sanitize deserialization and allocation paths

pentschev:
Add input validation and bounds checking across several deserialization and allocation code paths.

* Replace `AmReceiverCallbackOwnerType` with a fixed-size type.
* Validate `_packedRemoteKeySize` against the actual serialized data length before allocating in `RemoteKey::deserialize`.
* Validate `nframes` against `HeaderFramesSize` in `Header::deserialize` to prevent consumers from iterating past the fixed-size arrays.
* Check `malloc` return value in the `HostBuffer` constructor to avoid undefined behavior on allocation failure.
* Reject empty addresses in `createAddressFromString`.

### DIFF
--- a/cpp/src/remote_key.cpp
+++ b/cpp/src/remote_key.cpp
@@ -138,7 +138,8 @@ void RemoteKey::deserialize(const SerializedRemoteKey& serializedRemoteKey)
 
   const size_t fixedFieldsSize =
     sizeof(_packedRemoteKeySize) + sizeof(_memoryBaseAddress) + sizeof(_memorySize);
-  if (_packedRemoteKeySize > serializedRemoteKeyData.size() - fixedFieldsSize)
+  if (serializedRemoteKeyData.size() < fixedFieldsSize ||
+      _packedRemoteKeySize > serializedRemoteKeyData.size() - fixedFieldsSize)
     throw std::overflow_error("Packed remote key size exceeds serialized data length");
 
   _packedRemoteKeyVector = std::vector<char>(_packedRemoteKeySize);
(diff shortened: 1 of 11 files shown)


cpp/src/remote_key.cpp
```diff
@@ -136,14 +136,14 @@ void RemoteKey::deserialize(const SerializedRemoteKey& serializedRemoteKey)
 
   ss.read(reinterpret_cast<char*>(&_packedRemoteKeySize), sizeof(_packedRemoteKeySize));
 
-  // Use a vector to store data so we don't need to bother releasing it later.
+  const size_t fixedFieldsSize =
+    sizeof(_packedRemoteKeySize) + sizeof(_memoryBaseAddress) + sizeof(_memorySize);
+  if (_packedRemoteKeySize > serializedRemoteKeyData.size() - fixedFieldsSize)
+    throw std::overflow_error("Packed remote key size exceeds serialized data length");
```
madsbk, line 143: The `RemoteKey::deserialize` bound check has an unsigned underflow flaw that silently allows the very payloads this PR is intended to reject through. Need something like:

```suggestion
  const size_t fixedFieldsSize =
    sizeof(_packedRemoteKeySize) + sizeof(_memoryBaseAddress) + sizeof(_memorySize);
  if (serializedRemoteKeyData.size() < fixedFieldsSize ||
      _packedRemoteKeySize > serializedRemoteKeyData.size() - fixedFieldsSize)
    throw std::overflow_error("Packed remote key size exceeds serialized data length");
```
Also, it would be good to add a test

madsbk (changes requested):

cpp/src/remote_key.cpp
pentschev: Great catch! Fix is applied, and a test has been added in https://github.com/rapidsai/ucxx/pull/639/commits/2ff221ab1ba63cbeb1915a84c7ba6b2cf66d7a99.

madsbk (approved):
Thanks @pentschev

pentschev:
Thanks for the review Mads!

Issue #112, opened 2023-10-30
`RequestTagMulti` should expose CUDA stream or synchronize device


pentschev:
Currently `RequestTagMulti` doesn't take into consideration CUDA synchronization which is an error. The user should be able to consume any received data or release any sent data immediately after the request completes, without having to care about synchronization. Another valid approach would be to let the user specify a CUDA async stream, and in that case release the synchronization requirement back to the user.

PR #641, opened 2026-04-29
Add support for using CCCL for CUDA buffers

vyasr:
This PR adds support for using CCCL instead of rmm for handling device memory.


cpp/examples/basic.cpp
```diff
@@ -236,32 +254,29 @@ auto verify_buffers(ucxx::Buffer* expected, ucxx::Buffer* actual)
...
+    // copy device buffer to host
     host_buffer.resize(buffer->getSize());
-    auto stream = rmm::cuda_stream_default;
-    assert(
-      cudaMemcpyAsync(
-        host_buffer.data(), buffer->data(), buffer->getSize(), cudaMemcpyDefault, stream.value()) ==
-      cudaSuccess);
-    stream.synchronize();
+    assert(cudaMemcpy(host_buffer.data(), buffer->data(), buffer->getSize(), cudaMemcpyDefault) ==
+           cudaSuccess);
```
pentschev: Why drop the async memcpy API?

cpp/src/buffer_cccl.cpp
pentschev, line 1: Do we really need a .cu file? I would very much like to keep UCXX free from requiring nvcc.

cpp/tests/buffer.cpp
```diff
@@ -58,6 +58,14 @@ TEST_P(BufferAllocator, TestType)
     ASSERT_EQ(buffer->getType(), ucxx::BufferType::Invalid);
 #else
     GTEST_SKIP() << "UCXX was not built with RMM support";
+#endif
+  } else if (_type == ucxx::BufferType::CCCL) {
+#if UCXX_ENABLE_CCCL
+    auto buffer = std::dynamic_pointer_cast<ucxx::CCCLBuffer>(_buffer);
+    ASSERT_EQ(buffer->getType(), _type);
+    return;  // CCCLBuffer does not expose release(); post-release assertions do not apply
```
pentschev: Why not?

README.md
```diff
@@ -120,6 +120,35 @@ $ ./benchmarks/ucxx_perftest -m cuda-async -s 1048576 -n 10 127.0.0.1
...
+**Additional CCCL Memory Types (with `-DUCXX_ENABLE_CCCL=ON`):**
+- `cccl-device` - CCCL device memory pool
+- `cccl-shared` - CCCL shared memory resource
+- `cccl-cuda-async` - CCCL CUDA async memory resource
+- `cccl-cuda-async-managed` - CCCL CUDA async managed memory resource
+
+**Requirements for CCCL Support:**
+- UCXX compiled with `UCXX_ENABLE_CCCL=ON`
+- CCCL library available (fetched automatically via CMake)
+- CUDA 12+ runtime
```
pentschev:
```suggestion
```

pentschev: We don't claim support for CUDA < 12.

cpp/CMakeLists.txt
```diff
@@ -43,13 +43,16 @@ option(BUILD_BENCHMARKS "Configure CMake to build benchmarks" OFF)
 option(BUILD_EXAMPLES "Configure CMake to build examples" OFF)
 option(BUILD_SHARED_LIBS "Build UCXX shared libraries" ON)
 option(UCXX_ENABLE_RMM "Enable support for CUDA multi-buffer transfer with RMM" OFF)
+# TODO: Flip UCXX_ENABLE_CCCL default to OFF once devcontainer builds pass -DUCXX_ENABLE_CCCL=ON
```
pentschev, line 46: Unclear when that is, does that require changes to UCXX, devcontainers, elsewhere?

cpp/examples/basic.cpp
pentschev: Still relevant.

cpp/include/ucxx/buffer.h
```diff
@@ -286,7 +285,72 @@ class RMMBuffer : public Buffer {
...
+#if UCXX_ENABLE_CCCL
+/**
+ * @brief Opaque implementation struct for CCCLBuffer (defined in buffer_cccl.cu).
+ * Users should NOT access this type directly.
+ */
+// CCCLBuffer uses PIMPL pattern (opaque impl defined in buffer_cccl.cu) because CCCL
+// template types (cuda::buffer<cuda::std::byte, cuda::mr::device_accessible>) require
+// nvcc compilation and cannot be forward-declared in a standard C++ header. Compare
+// with RMMBuffer which forward-declares rmm::device_buffer directly since rmm types
+// are standard C++ headers.
```
pentschev:
```suggestion
```

pentschev: I think the comment is unnecessary. I'm working on moving to Pimpl anyway to reduce ABI breakage.

cpp/include/ucxx/constructors.h
```diff
@@ -56,9 +57,11 @@ class Worker;
...
                                                        void* callbackArgs);
 
-[[nodiscard]] std::shared_ptr<Worker> createWorker(std::shared_ptr<Context> context,
-                                                   const bool enableDelayedSubmission,
-                                                   const bool enableFuture);
+[[nodiscard]] std::shared_ptr<Worker> createWorker(
+  std::shared_ptr<Context> context,
+  const bool enableDelayedSubmission,
+  const bool enableFuture,
+  const BufferType cudaBufferType = BufferType::Invalid);
```
pentschev:
```suggestion
[[nodiscard]] std::shared_ptr<Worker> createWorker(std::shared_ptr<Context> context,
                                                   const bool enableDelayedSubmission,
                                                   const bool enableFuture);
```

pentschev: Let's not do this change here. My intent is to keep this constructor stable and will eventually deprecate/remove in favor of the `WorkerBuilder`. For now we can keep RMM as the only choice with the original constructor, if someone wants to use CCCL they should use the new constructor and set via `cudaBufferType`.

cpp/include/ucxx/worker.h
```diff
@@ -157,10 +159,14 @@ class Worker : public Component {
...
    *                         `ucxx::Request`, currently used only by `ucxx::python::Worker`.
+   * @param[in] cudaBufferType the preferred buffer type for CUDA allocations. If
+   *                           `BufferType::Invalid` (default), will use CCCL if compiled
+   *                           with CCCL support, otherwise RMM if compiled with RMM support.
    */
   explicit Worker(std::shared_ptr<Context> context,
                   const bool enableDelayedSubmission = false,
-                  const bool enableFuture            = false);
+                  const bool enableFuture            = false,
+                  const BufferType cudaBufferType    = BufferType::Invalid);
```
pentschev:
```suggestion
                  const bool enableFuture            = false);
```

cpp/include/ucxx/worker.h
```diff
@@ -177,7 +183,8 @@ class Worker : public Component {
    */
   friend std::shared_ptr<Worker> createWorker(std::shared_ptr<Context> context,
                                               const bool enableDelayedSubmission,
-                                              const bool enableFuture);
+                                              const bool enableFuture,
+                                              const BufferType cudaBufferType);
```
pentschev:
```suggestion
                                              const bool enableFuture);
```

cpp/include/ucxx/worker.h
```diff
@@ -1015,11 +1046,15 @@ class Worker : public Component {
  *                                    progress thread.
  * @param[in] enableFuture if `true`, notifies the future associated with each
  *                         `ucxx::Request`, currently used only by `ucxx::python::Worker`.
+ * @param[in] cudaBufferType the preferred buffer type for CUDA allocations. If
+ *                           `BufferType::Invalid` (default), will use CCCL if compiled
+ *                           with CCCL support, otherwise RMM if compiled with RMM support.
```
pentschev:
```suggestion
```

cpp/include/ucxx/worker.h
```diff
@@ -1015,11 +1046,15 @@ class Worker : public Component {
...
  * @returns The `shared_ptr<ucxx::Worker>` object
  */
 std::shared_ptr<Worker> createWorker(std::shared_ptr<Context> context,
                                      const bool enableDelayedSubmission,
-                                     const bool enableFuture);
+                                     const bool enableFuture,
+                                     const BufferType cudaBufferType);
```
pentschev:
```suggestion
                                     const bool enableFuture);
```

cpp/src/worker.cpp
```diff
@@ -33,11 +33,21 @@ namespace ucxx {
 
 Worker::Worker(std::shared_ptr<Context> context,
                const bool enableDelayedSubmission,
-               const bool enableFuture)
+               const bool enableFuture,
+               const BufferType cudaBufferType)
```
pentschev: I've made suggestions on the header files, but it's getting to tedious to do so via the UI. :sweat_smile: 

Please remove this and just set `_cudaBufferType(BufferType::RMM)` here and below.

pentschev (changes requested):
Thanks @vyasr , overall this looks really good. I've left some questions and change requests. What do you feel about completely moving away from RMM to CCCL? If build and tests pass, should CCCL be considered stable? We might then be able to simply remove RMM in a follow-up PR I think, we probably don't need to keep it for a deprecation cycle.

cpp/CMakeLists.txt
vyasr: This requires a change to the rapidsai devcontainers repo to pass `-DUCXX_ENABLE_CCCL=ON` in the devcontainer build configuration. Once that's done, we can flip this default to OFF.

cpp/include/ucxx/constructors.h
vyasr: I'm not sure what new constructor you mean but relying on `setCudaBufferType` is fine with me.

cpp/src/buffer_cccl.cpp
vyasr: Done — renamed to `.cpp` in commit bfce02d.

cpp/tests/buffer.cpp
vyasr: Updated the test comment to clarify why release() is skipped for CCCLBuffer (PIMPL prevents direct pointer access).

README.md
vyasr: Removed.

cpp/include/ucxx/buffer.h
vyasr: Removed the PIMPL comment.

cpp/src/worker.cpp
vyasr: Applied all formatting suggestions — reverted the constructor signature change and using `setCudaBufferType` post-construction instead.

cpp/include/ucxx/worker.h
vyasr: Applied.

cpp/include/ucxx/worker.h
vyasr: Applied.

cpp/include/ucxx/worker.h
vyasr: Applied.

cpp/include/ucxx/worker.h
vyasr: Applied.

cpp/tests/buffer.cpp
vyasr: I didn't see an immediate use case for release. We weren't using the RmmBuffer version anywhere here, and if we were to add it, it would release to a raw pointer since the CCCL buffer type isn't part of the public API. I didn't want to add an unused API, especially if it's releasing via unmanaged pointers.

cpp/examples/basic.cpp
vyasr: Put back.

vyasr:
> Thanks @vyasr , overall this looks really good. I've left some questions and change requests. What do you feel about completely moving away from RMM to CCCL? If build and tests pass, should CCCL be considered stable? We might then be able to simply remove RMM in a follow-up PR I think, we probably don't need to keep it for a deprecation cycle.

I figure it would make more sense to get this merged to a point where we can do some testing in production before we drop RMM support entirely. It's one extra PR to do a deprecation before a removal but I figure it's worth it for some stability.

cpp/include/ucxx/constructors.h
pentschev: The "new"/experimental constructor is the [`WorkerBuilder`](https://github.com/rapidsai/ucxx/blob/main/cpp/include/ucxx/experimental/worker_builder.h), which uses the builder-pattern to allow extending the arguments without breaking API/ABI, which you have updated as expected [here](https://github.com/rapidsai/ucxx/pull/641/files#diff-6a41680baf3695da54f469d0861add50450ff80a49617f8294b0c6dcc3c46ed5R93-R99). I don't want to extend the "old" constructor anymore, and instead add new features only via the `WorkerBuilder`, meanwhile, the old should still work for existing parameters and current defaults.

cpp/include/ucxx/worker.h
```diff
@@ -492,6 +494,30 @@ class Worker : public Component {
...
+   *
+   * Configure which buffer type to use when allocating CUDA buffers for incoming
+   * multi-buffer tag receives.
+   *
+   * @param[in] bufferType  the preferred buffer type (must be `BufferType::RMM` or
+   *                        `BufferType::CCCL`).
+   *
+   * @throws std::invalid_argument if bufferType is not RMM or CCCL.
+   */
+  void setCudaBufferType(BufferType bufferType);
```
pentschev: Thinking a bit more about this, I don't think it's a good idea to expose `setCudaBufferType` publicly. This allows changing the buffer type _while_ the worker runs, and that may cause confusion while types are being constructed/extracted. We should still provide a way to set the buffer type, but only via the [WorkerBuilder here](https://github.com/rapidsai/ucxx/pull/641/files#diff-6a41680baf3695da54f469d0861add50450ff80a49617f8294b0c6dcc3c46ed5R93-R99).

pentschev:
> > Thanks @vyasr , overall this looks really good. I've left some questions and change requests. What do you feel about completely moving away from RMM to CCCL? If build and tests pass, should CCCL be considered stable? We might then be able to simply remove RMM in a follow-up PR I think, we probably don't need to keep it for a deprecation cycle.
> 
> I figure it would make more sense to get this merged to a point where we can do some testing in production before we drop RMM support entirely. It's one extra PR to do a deprecation before a removal but I figure it's worth it for some stability.

Ok, let's do that. I don't know if anybody is using that in practice, so we might as well just deprecate it immediately and remove it in 26.08. WDYT?

cpp/include/ucxx/worker.h
vyasr: Sounds good. I made that change. Hopefully this extra machinery is short-lived anyway and we can just delete all the rmm code paths early in 26.08.

vyasr:
> > > Thanks @vyasr , overall this looks really good. I've left some questions and change requests. What do you feel about completely moving away from RMM to CCCL? If build and tests pass, should CCCL be considered stable? We might then be able to simply remove RMM in a follow-up PR I think, we probably don't need to keep it for a deprecation cycle.
> > 
> > 
> > I figure it would make more sense to get this merged to a point where we can do some testing in production before we drop RMM support entirely. It's one extra PR to do a deprecation before a removal but I figure it's worth it for some stability.
> 
> Ok, let's do that. I don't know if anybody is using that in practice, so we might as well just deprecate it immediately and remove it in 26.08. WDYT?

Sure, done.

pentschev (approved):
LGTM, thanks @vyasr for putting in all the work for this! :smile:

cpp/benchmarks/include/buffer_interface.hpp
```diff
@@ -870,3 +907,176 @@ inline std::unique_ptr<RmmBufferInterfaceBase> RmmBufferInterfaceBase::createBuf
 }
 
 #endif  // UCXX_BENCHMARKS_ENABLE_RMM
+
+#ifdef UCXX_BENCHMARKS_ENABLE_CCCL
+
+namespace cccl_benchmark_detail {
+
+// CCCL benchmarks use nullptr (the default CUDA stream) for async operations because
+// CCCLBuffer is constructed with cudaStream_t{0} and does not expose a per-buffer stream.
```
bdice: > does not expose a per-buffer stream.

This seems wrong. CCCL buffers do have a stream member exposed: https://github.com/NVIDIA/cccl/blob/057cd14a42d381ae74369f4f6dc74fb3d216a81d/libcudacxx/include/cuda/__container/buffer.h#L706-L712

cpp/src/buffer_cccl.cpp
```diff
@@ -0,0 +1,59 @@
...
+  // Unlike RMM (which initializes context internally via its device resource setup),
+  // CCCL needs explicit initialization. cudaFree(0) is the standard zero-cost idiom.
+  static auto get_device_pool()
+  {
+    cudaFree(0);  // Ensure CUDA primary context is initialized
+    return ::cuda::device_default_memory_pool(::cuda::device_ref{0});
+  }
+
+  explicit CCCLBufferImpl(const size_t size)
+    : buffer{::cudaStream_t{0}, get_device_pool(), size, ::cuda::no_init}
```
bdice, line 36: I don't see constructors for other streams/devices? Are all CCCL buffers forced to live on the default stream on device 0? How does this work with the RMM buffer today?

cpp/benchmarks/include/buffer_interface.hpp
```diff
@@ -870,3 +907,176 @@ inline std::unique_ptr<RmmBufferInterfaceBase> RmmBufferInterfaceBase::createBuf
...
+                                     nullptr),
+                     "CCCL recv data copy for verification");
+  CUDA_EXIT_ON_ERROR(cudaStreamSynchronize(nullptr),
+                     "CCCL stream synchronization for verification");
+  for (std::size_t j = 0; j < messageSize; ++j)
+    if (recvData[j] != sendData[j])
+      throw std::runtime_error("CCCL data verification failed at byte " + std::to_string(j));
+}
+
+struct CcclDevicePoolBuffers {
```
bdice, line 954: We're inconsistent with naming. In some classes/enum values, we use CCCL in all caps. In others we're using Cccl casing.

python/ucxx/ucxx/_lib/libucxx.pyx
```diff
@@ -621,10 +697,26 @@ cdef class UCXWorker():
...
-                    UCS_MEMORY_TYPE_CUDA, rmm_am_allocator
-                )
+                if self._worker.get().getCudaBufferType() == BufferType.RMM:
+                    with gil:
+                        warnings.warn(
+                            "RMM CUDA buffer support is deprecated and"
+                            " will be removed in a future release. Use"
+                            " CCCL buffers instead (set"
+                            " UCXX_ENABLE_CCCL=ON and"
+                            " UCXX_ENABLE_RMM=OFF).",
```
bdice:
```suggestion
                            "RMM CUDA buffer support is deprecated and "
                            "will be removed in a future release. Use "
                            "CCCL buffers instead (set "
                            "UCXX_ENABLE_CCCL=ON and "
                            "UCXX_ENABLE_RMM=OFF).",
```

cpp/src/buffer_cccl.cpp
pentschev: Yes, today everything is running on the default stream. We haven't made it stream-aware yet https://github.com/rapidsai/ucxx/issues/112 .

cpp/benchmarks/include/buffer_interface.hpp
pentschev: This is due to existing inconsistency, https://github.com/rapidsai/ucxx/blob/main/cpp/include/ucxx/buffer.h uses `RMM*` and https://github.com/rapidsai/ucxx/blob/main/cpp/benchmarks/include/buffer_interface.hpp uses `Rmm*`. I think Vyas' choice was right, the burden on fixing consistency is on me on a follow-up PR. I need to do a full-run of UCXX consistency in various aspects now that we're prepping for upstreaming. For this PR I think we should leave it as it is, consistent with the files they live in.

cpp/benchmarks/include/buffer_interface.hpp
vyasr: https://github.com/rapidsai/ucxx/pull/641#discussion_r3227286904

python/ucxx/ucxx/_lib/libucxx.pyx
vyasr: Applied in 4a4cf75.

cpp/benchmarks/include/buffer_interface.hpp
vyasr: Fixed the comment in 4a4cf75 — replaced the inaccurate claim about CCCLBuffer not exposing a per-buffer stream with a note that UCX is not yet stream-aware, linking to #112.

bdice (approved):

PR #650, opened 2026-05-13
Restore distributed-ucxx flaky markers

pentschev:
Some distributed-ucxx tests have been failing recently since their flaky markers were removed as it was believed the changes were sufficient to fix the issues, they weren't. Restore the flaky markers until the root cause can be narrowed down and the tests ultimately fixed.


TomAugspurger (approved):

pentschev:
Thanks Tom!

PR #655, opened 2026-05-19
Remove deprecated `Address::getString()`

pentschev:
`Address::getString()` was deprecated in favor of the new `Address::getStringView()`. Removes that for UCXX v0.51/RAPIDS 26.08.

### DIFF
--- a/cpp/src/address.cpp
+++ b/cpp/src/address.cpp
@@ -2,9 +2,9 @@
  * SPDX-FileCopyrightText: Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES.
  * SPDX-License-Identifier: BSD-3-Clause
  */
+#include <cstring>
 #include <memory>
 #include <stdexcept>
-#include <string>
 #include <string_view>
 
 #include <ucxx/address.h>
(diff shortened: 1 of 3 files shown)


nirandaperera (approved):

pentschev:
Thanks Niranda and Samson!

cpp/src/address.cpp
```diff
@@ -2,9 +2,9 @@
  * SPDX-FileCopyrightText: Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES.
  * SPDX-License-Identifier: BSD-3-Clause
  */
+#include <cstring>
```
jakirkham, line 5: Where is this being used?

jakirkham (approved):
This looks reasonable to me. Thank you for the cleanup Peter! 🙏

Sorry I didn't get here quick enough. Had one minor question below

cpp/src/address.cpp
pentschev: In [`memcpy`](https://github.com/pentschev/ucxx/blob/4a56e41c25349cf6447162aa3d24894c66d5655c/cpp/src/address.cpp#L49)

jakirkham: Ah of course. Thanks Peter! 🙏

PR #656, opened 2026-05-19
Remove deprecated RMM `Buffer` support

pentschev:
The `Buffer` support for RMM was deprecated in favor of CCCL. Remove that for UCXX v0.51/RAPIDS 26.08.


cpp/benchmarks/CMakeLists.txt
```diff
@@ -45,7 +45,7 @@ function(ConfigureBench CMAKE_BENCH_NAME)
     target_link_libraries(${CMAKE_BENCH_NAME} PRIVATE CUDA::cudart_static)
   endif()
 
-  # RMM memory resources for CUDA benchmarks (requires UCXX_ENABLE_RMM and get_rmm.cmake)
+  # RMM memory resources for CUDA benchmarks.
   if(UCXX_BENCHMARKS_ENABLE_CUDA AND UCXX_ENABLE_RMM)
```
bdice: Do we need to remove more RMM references in this block?

cpp/tests/CMakeLists.txt
```diff
@@ -29,6 +29,10 @@ function(ConfigureTest CMAKE_TEST_NAME)
     ${CMAKE_TEST_NAME} PRIVATE ucxx GTest::gmock_main GTest::gtest_main
                                $<TARGET_NAME_IF_EXISTS:conda_env>
   )
+  if(UCXX_ENABLE_RMM)
```
bdice: Remove more RMM references here?

cpp/benchmarks/CMakeLists.txt
pentschev: We still support RMM benchmarks, they have nothing to do with the `Buffer` class. Maybe at some point we may want to remove that, but for now I think it's useful for us to have the option to compare between both, unless our plan is to completely remove RMM memory resources and have everyone rely on CCCL directly from 26.08.

cpp/tests/CMakeLists.txt
pentschev: Ditto.

cpp/benchmarks/CMakeLists.txt
vyasr: No, we're not quite there yet, so let's leave these benchmarks. Most likely we'll never lose all the mrs in rmm, but I am hopeful that we'll eventually lose buffers.

cpp/CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ option(BUILD_TESTS "Configure CMake to build tests" ON)
 option(BUILD_BENCHMARKS "Configure CMake to build benchmarks" OFF)
 option(BUILD_EXAMPLES "Configure CMake to build examples" OFF)
 option(BUILD_SHARED_LIBS "Build UCXX shared libraries" ON)
-option(UCXX_ENABLE_RMM "Enable support for CUDA multi-buffer transfer with RMM" OFF)
+option(UCXX_ENABLE_RMM "Enable RMM-backed test and benchmark code paths" OFF)
 # TODO: Flip UCXX_ENABLE_CCCL default to OFF once devcontainer builds pass -DUCXX_ENABLE_CCCL=ON
```
vyasr, line 47: We need to update the devcontainers to use CCCL [here](https://github.com/rapidsai/devcontainers/blob/main/features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml#L34), then we can change this flag here. That way CCCL support stays optional.

vyasr (changes requested):
The README needs to be updated too, it refers to building the library with rmm support.

pentschev:
> The README needs to be updated too, it refers to building the library with rmm support.

Fixed in https://github.com/rapidsai/ucxx/pull/656/commits/50820706488a3b067a2c244f28a45fe462822b03

cpp/CMakeLists.txt
pentschev: Thanks for pointing that, opened https://github.com/rapidsai/devcontainers/pull/705 to address that.

vyasr (approved):

cpp/CMakeLists.txt
nirandaperera: @pentschev Would it be too much to rename `UCXX_ENABLE_RMM` to `UCXX_ENABLE_RMM_TESTING`?
I had the same question as @bdice [here](https://github.com/rapidsai/ucxx/pull/656/changes#r3268679564)

nirandaperera (approved):
LGTM. It would be nice if we could rename `UCXX_ENABLE_RMM` to something more representative now (maybe UCXX_ENABLE_RMM_TESTING).
But I dont want to block the PR for that.

trxcllnt:
Adding [this](https://github.com/rapidsai/ucxx/pull/661/changes#diff-7caf2bdc6aa44337cf6e4b04392bc3e5940a7356512b4d525aba3bcda0be6c3bR32-R35) to `cpp/tests/CMakeLists.txt` should fix the `UCXX_TEST` linker errors:
```cmake
  if(UCXX_ENABLE_CCCL)
    target_link_libraries(${CMAKE_TEST_NAME} PRIVATE CUDA::cudart_static)
  endif()
```

pentschev:
> Adding [this](https://github.com/rapidsai/ucxx/pull/661/changes#diff-7caf2bdc6aa44337cf6e4b04392bc3e5940a7356512b4d525aba3bcda0be6c3bR32-R35) to `cpp/tests/CMakeLists.txt` should fix the `UCXX_TEST` linker errors:
> 
> ```cmake
>   if(UCXX_ENABLE_CCCL)
>     target_link_libraries(${CMAKE_TEST_NAME} PRIVATE CUDA::cudart_static)
>   endif()
> ```

Thanks Paul, done in https://github.com/rapidsai/ucxx/pull/656/commits/e8ff9b0e19514d3d88dc581e762c1b3287adb4bc.

> LGTM. It would be nice if we could rename `UCXX_ENABLE_RMM` to something more representative now (maybe UCXX_ENABLE_RMM_TESTING). But I dont want to block the PR for that.

Thanks Niranda, done in https://github.com/rapidsai/ucxx/pull/656/commits/7e8ca764e01eac9fc07966d152beceba40d976e8.

trxcllnt:
Looks like either a similar change needs to be made in `python/ucxx`, or ucxx needs to specify its `CUDA::cudart_static` dependency in the public interface when `UCXX_ENABLE_CCCL=ON`

trxcllnt:
```diff
diff --git a/python/ucxx/ucxx/_lib/CMakeLists.txt b/python/ucxx/ucxx/_lib/CMakeLists.txt
index 68922af..d3488bf 100644
--- a/python/ucxx/ucxx/_lib/CMakeLists.txt
+++ b/python/ucxx/ucxx/_lib/CMakeLists.txt
@@ -5,8 +5,10 @@
 # cmake-format: on
 # =================================================================================
 
+find_package(CUDAToolkit REQUIRED)
+
 set(cython_sources arr.pyx libucxx.pyx)
-set(linked_libraries ucxx::ucxx ucxx::python)
+set(linked_libraries ucxx::ucxx ucxx::python CUDA::cudart_static)
 
 rapids_cython_create_modules(
   CXX
```

PR #657, opened 2026-05-19
Update test to remove use of deprecated RMM API

pentschev:
https://github.com/rapidsai/rmm/pull/2387 removed `rmm.mr.get_current_device_resource_type`, update test using it accordingly.

### DIFF
--- a/python/distributed-ucxx/distributed_ucxx/tests/test_ucxx_config.py
+++ b/python/distributed-ucxx/distributed_ucxx/tests/test_ucxx_config.py
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023-2025, NVIDIA CORPORATION & AFFILIATES.
+# SPDX-FileCopyrightText: Copyright (c) 2023-2026, NVIDIA CORPORATION & AFFILIATES.
 # SPDX-License-Identifier: BSD-3-Clause
 
 from __future__ import annotations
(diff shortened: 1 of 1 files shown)


python/distributed-ucxx/distributed_ucxx/tests/test_ucxx_config.py
```diff
@@ -209,12 +214,10 @@ def test_ucx_config_w_env_var(ucxx_loop, cleanup, loop, protocol):
                     sleep(0.1)
 
                 # Check for RMM pool resource type
-                rmm_resource = c.run_on_scheduler(
-                    rmm.mr.get_current_device_resource_type
-                )
+                rmm_resource = c.run_on_scheduler(current_device_resource_class)
                 assert rmm_resource == rmm.mr.PoolMemoryResource
```
bdice, line 218: Rather than comparing type equality directly, should we use `isinstance`? Typically type equality isn’t the ideal solution.

bdice (approved):
I’m approving this as a direct replacement for the removed API. I wonder if there’s a better way to test this though.

python/distributed-ucxx/distributed_ucxx/tests/test_ucxx_config.py
pentschev: This was merged too quickly before I could address this suggestion, so doing it now in https://github.com/rapidsai/ucxx/pull/658/files#diff-961d0e1704627ac5b02528550ac6caf74df2f2e3b06ce96a678b7812e1d70242 .

PR #661, opened 2026-05-20
Update devcontainer `cacheFrom` image names to use ucxx version

trxcllnt:
I forgot ucxx uses a different version scheme, so this updates the image tags added in #659 to match the ucxx version instead of the RAPIDS version.


wence- (approved):

trxcllnt:
~ucxx-python configure failure should be resolved by https://github.com/rapidsai/devcontainers/pull/707~

trxcllnt:
`ucxx-python` configuration should succeed once https://github.com/rapidsai/ucxx/pull/656 is merged.

bdice (approved):

bdice (approved):

PR #664, opened 2026-05-27
Synchronize Python futures pool access

pentschev:
Before this change `populateFuturesPool()` read `_futuresPool.size()` outside the mutex, and `getFuture()` also read `_futuresPool.size()` outside the mutex. A concrete race can occur while the [Python notifier thread](https://github.com/rapidsai/ucxx/blob/main/python/ucxx/ucxx/_lib_async/notifier_thread.py) runs `Worker::populateFuturesPool()` and concurrently a request future is fulfilled by another thread via `Worker::getFuture()`.

### DIFF
--- a/cpp/python/src/worker.cpp
+++ b/cpp/python/src/worker.cpp
@@ -22,6 +22,23 @@ namespace ucxx {
 
 namespace python {
 
+namespace {
+
+class GilState {
+ public:
+  GilState() : _state(PyGILState_Ensure()) {}
+
+  GilState(const GilState&)            = delete;
+  GilState& operator=(const GilState&) = delete;
+
+  ~GilState() { PyGILState_Release(_state); }
+
+ private:
+  PyGILState_STATE _state;
+};
+
+}  // namespace
+
 Worker::Worker(std::shared_ptr<Context> context,
                const bool enableDelayedSubmission,
                const bool enableFuture)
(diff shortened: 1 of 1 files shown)


cpp/python/src/worker.cpp
```diff
@@ -52,24 +52,37 @@ std::shared_ptr<::ucxx::Worker> createWorker(std::shared_ptr<Context> context,
...
+    std::lock_guard<std::mutex> lock(_futuresPoolMutex);
+    if (_futuresPool.size() >= minPoolSize) return;
+    futuresNeeded = maxPoolSize - _futuresPool.size();
+  }
+
+  decltype(_futuresPool) newFuturesPool;
+  PyGILState_STATE state = PyGILState_Ensure();
+  for (size_t i = 0; i < futuresNeeded; ++i)
+    newFuturesPool.emplace(createFuture(_notifier));
+  PyGILState_Release(state);
```
wence-: question: Can `createFuture` or `emplace` throw? If yes, then we technically need some kind of cleanup for `PyGILState_Release`, I think.

cpp/python/src/worker.cpp
```diff
@@ -79,39 +92,48 @@ void Worker::clearFuturesPool()
...
                    shared_from_this().get());
-    std::lock_guard<std::mutex> lock(_futuresPoolMutex);
-    PyGILState_STATE state = PyGILState_Ensure();
     decltype(_futuresPool) newFuturesPool;
-    std::swap(_futuresPool, newFuturesPool);
-    PyGILState_Release(state);
+    {
+      std::lock_guard<std::mutex> lock(_futuresPoolMutex);
+      std::swap(_futuresPool, newFuturesPool);
+    }
```
wence-, line 117: question: we're no longer holding the gil while swapping out the futures, does that matter?

cpp/python/src/worker.cpp
pentschev: Ah, good catch! Yes, I think both can throw. Fixed that now with https://github.com/rapidsai/ucxx/pull/664/commits/f4853c81c459f0b025cae76db4fba62c5194e7ed .

cpp/python/src/worker.cpp
pentschev: No, only C++ queue ownership is swapped, no Python refcounts or Python API calls are made, so this should be safe without the GIL.

pentschev (commented):
Thanks @wence- . I've addressed your comments, please have another look.

wence- (approved):

pentschev:
Thanks Lawrence!

PR #666, opened 2026-05-28
Increase `test_many_servers_many_clients` connection timeout

pentschev:
`test_many_servers_many_clients` imposes higher strain on CPU due to the amount of endpoints being created simultaneously, potentially causing connection timeouts leading to occasional CI failures. This change increases the connection timeout for cases creating at least 50 endpoints which should help preventing those failures on slower runners.


madsbk (approved):

gforsyth:
admin merging this per PIC request -- it's only touching tests, so no risk to release stability

pentschev:
Thanks @madsbk and @gforsyth !

PR #668, opened 2026-05-28
Safely complete and release blocking endpoint close requests

pentschev:
`Endpoint::closeBlocking()` now avoids treating an in-progress UCX close request as completed, reports timeout when completion cannot be observed, and always frees the UCX request it owns before returning. This prevents premature close success and avoids leaking close requests.


madsbk (approved):
LGTM

pentschev:
Thanks Mads! :slightly_smiling_face:

PR #669, opened 2026-05-28
Synchronize progress thread state access

pentschev:
`ucxx::Worker` now tracks progress thread running/polling state and progress thread ID with atomics. This removes data races between progress thread start/stop and callers using `registerGeneric{Pre,Post}()`.

### DIFF
--- a/cpp/include/ucxx/worker.h
+++ b/cpp/include/ucxx/worker.h
@@ -4,6 +4,7 @@
  */
 #pragma once
 
+#include <atomic>
 #include <functional>
 #include <memory>
 #include <mutex>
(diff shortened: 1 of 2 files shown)


cpp/include/ucxx/worker.h
```diff
@@ -61,7 +62,10 @@ class Worker : public Component {
   std::unique_ptr<InflightRequests> _inflightRequestsToCancel{
     std::make_unique<InflightRequests>()};  ///< The inflight requests scheduled to be canceled
   WorkerProgressThread _progressThread{};   ///< The progress thread object
-  std::thread::id _progressThreadId{};      ///< The progress thread ID
+  std::atomic<bool> _progressThreadRunning{false};      ///< Whether the progress thread is running
+  std::atomic<bool> _progressThreadPollingMode{false};  ///< Whether the progress thread polls
+  std::mutex _progressThreadIdMutex{};                  ///< Mutex to access progress thread ID
+  std::thread::id _progressThreadId{};                  ///< The progress thread ID
```
wence-: Why not just `std::atomic<std::thread::id>`? `std::thread::id` is trivially-copyable and (it just wraps a native handle that is just an integer)

wence- (changes requested):
Not sure we need a mutex to protect the thread id, it seems like you could just use an atomic

cpp/include/ucxx/worker.h
pentschev: Good point. The progress thread ID is standalone state and is not protecting any compound invariant with other fields, so a mutex is unnecessary here. I replaced it with `std::atomic<std::thread::id>` in https://github.com/rapidsai/ucxx/pull/669/commits/8f60a72dbe5b9087b4571fcdf31209799c5f5e79 and kept acquire/release ordering consistent with the adjacent progress-thread state atomics.

wence- (approved):
Please update the PR description

pentschev:
> Please update the PR description

Good call. Done, and thanks for reviewing!

PR #677, opened 2026-06-03
Add builder for `Request` types

pentschev:
Adds experimental builder classes for all request types.

This is a breaking API change. `Request`-returning APIs now return builder objects. Existing code using `auto req = endpoint_or_worker->requestMethod(...)` will now deduce a builder type instead of `std::shared_ptr<Request>`, users should update such code to either call `.build()` (preferred, the old constructor will be removed soon) or explicitly request the pointer type:

```cpp
auto req = endpoint->tagSend(...).build();
std::shared_ptr<ucxx::Request> req = endpoint->tagSend(...);
```


cpp/include/ucxx/request_am.h
```diff
@@ -169,3 +169,6 @@ class RequestAm : public Request {
 };
 
 }  // namespace ucxx
+
+// Include experimental features
+#include <ucxx/experimental/request_am_builder.h>
```
nirandaperera: Can we move this include to the top of the file?

cpp/include/ucxx/request_endpoint_close.h
```diff
@@ -120,3 +120,6 @@ class RequestEndpointClose : public Request {
 };
 
 }  // namespace ucxx
+
+// Include experimental features
+#include <ucxx/experimental/request_endpoint_close_builder.h>
```
nirandaperera: Same here

cpp/include/ucxx/experimental/detail/register_inflight_request.h
```diff
@@ -0,0 +1,35 @@
...
+namespace detail {
+
+inline void registerInflightRequest(std::shared_ptr<Component> const& component,
+                                    std::shared_ptr<Request> const& req)
+{
+  if (auto ep = std::dynamic_pointer_cast<Endpoint>(component))
+    std::ignore = ep->registerInflightRequest(req);
+  else if (auto wk = std::dynamic_pointer_cast<Worker>(component))
+    std::ignore = wk->registerInflightRequest(req);
+}
```
nirandaperera: Nit. I'd rather do this. I think component does not need to be shared_ptr here. and req shared_ptr copy needs to be explict 

```suggestion
inline void registerInflightRequest(Component& component,
                                    std::shared_ptr<Request> req)
{
  if (auto ep = std::dynamic_cast<Endpoint&>(component))
    std::ignore = ep.registerInflightRequest(std::move(req));
  else if (auto wk = std::dynamic_cast<Worker&>(component))
    std::ignore = wk.registerInflightRequest(std::move(req));
}
```

cpp/src/experimental/request_am_builder.cpp
```diff
@@ -0,0 +1,40 @@
...
+{
+  auto req = ucxx::createRequestAm(
+    _endpoint, _requestData, _enablePythonFuture, _callbackFunction, _callbackData);
+  detail::registerInflightRequest(_endpoint, req);
+  return req;
+}
+
+RequestAmBuilder::operator std::shared_ptr<RequestAm>() const { return build(); }
+
+RequestAmBuilder::operator std::shared_ptr<Request>() const { return build(); }
```
nirandaperera: Couple of questions for all the builders.
Wouldnt this be prone to accidental calls to `build()` multiple times? Is calling `build` multiple times safe? If it is safe, we can introduce a CRTP to implement all these operator methods to all builders. If it is not safe, I think we should remove this.

cpp/src/endpoint.cpp
```diff
@@ -240,11 +247,19 @@ Endpoint::~Endpoint()
 std::shared_ptr<Request> Endpoint::close(const bool enablePythonFuture,
                                          EndpointCloseCallbackUserFunction callbackFunction,
                                          EndpointCloseCallbackUserData callbackData)
+{
+  return closeRequest(_endpointErrorHandling, enablePythonFuture, callbackFunction, callbackData);
```
nirandaperera: Nit. Move `std::funciton`s to avoid temp copies

cpp/include/ucxx/endpoint.h
```diff
@@ -365,9 +416,9 @@ class Endpoint : public Component {
                                         completion.
    * @param[in] callbackData            user-defined data to pass to the `callbackFunction`.
    *
-   * @returns Request to be subsequently checked for the completion and its state.
+   * @returns Builder to set optional parameters and submit the request.
    */
-  [[nodiscard]] std::shared_ptr<Request> amSend(
+  [[nodiscard]] experimental::RequestAmBuilder amSend(
```
nirandaperera: So, is it okay to return an experimental object now to a previous method that returned a production one?

nirandaperera (dismissed):
I mainly had some questions and nitpicks

cpp/include/ucxx/experimental/detail/register_inflight_request.h
pentschev: This is a good catch, however, both uses of `const&` were an oversight and I prefer instead to use the project's convention of pass-by-value instead. Taking `req` by value instead of `const&` and `std::move` is a strict improvement. For `Component`, we pay an extra cost of two extra atomic operations per request submission are immaterial in any case. The change is purely for consistency with the project's convention.

For completeness: we cannot move to use with the (potentially multiple) `dynamic_pointer_cast` calls. However, we accept the cost of two extra atomic operations per request submission since they are mostly immaterial here. The change is purely for consistency with the project's convention.

Furthermore, the original suggestion as written has a silent defect: `dynamic_cast<T&>` on a reference doesn't return null on failure — it throws `std::bad_cast`. So `if (auto ep = dynamic_cast<Endpoint&>(component))` would throw rather than fall through to the `Worker` branch. The correct pointer form would be `dynamic_cast<Endpoint*>(&component)`, but that introduces raw pointers contrary to the project's coding standard. The `shared_ptr` + `dynamic_pointer_cast` form is kept.

cpp/include/ucxx/request_am.h
pentschev: Done. I've made an exception and included the files that are not in the scope of this PR, since this is small and applies to those cases too.

cpp/include/ucxx/request_endpoint_close.h
pentschev: Done as well.

cpp/src/experimental/request_am_builder.cpp
pentschev: Good catch, calling `build()` more than once is not safe. Each call independently invokes the UCX factory and registers a new inflight request, which would post duplicate operations against the same buffer. I've added a `mutable bool _built` guard to the CRTP base: `markBuilt()` is called at the top of every `build()` and throws `std::logic_error` on a second call.

On the CRTP suggestion for the operator methods: the practical problem is that users commonly hold requests as the abstract `std::shared_ptr<Request>` rather than derived classes like `std::shared_ptr<RequestTag>` or `std::shared_ptr<RequestAm>`. This means both `operator std::shared_ptr<ConcreteType>()` and `operator std::shared_ptr<Request>()` are needed, not just one. The concrete-type operator can't be placed in the CRTP base without knowing the concrete request type (e.g. `RequestTag`), which would require each builder to declare a `result_type` alias that adds more boilerplate, not less. The `Request`-typed operator could go in the base in principle, but it would be instantiated in every translation unit that includes a builder header, and at that point `endpoint.h` (which is typically what pulls in builder headers) only forward-declares the concrete request types without their full class definitions. The implicit conversion from `shared_ptr<RequestTag>` to `shared_ptr<Request>` requires knowing the inheritance relationship, which is only available in the builder `.cpp` files where the full request headers are included. Keeping the operators in the `.cpp` files, as they are now, I think is the correct and safe placement.

cpp/src/endpoint.cpp
pentschev: Done.

cpp/include/ucxx/endpoint.h
pentschev: You're right, and this is a real tension I've accepted deliberately after careful consideration. Ideally a production method would not return an experimental type. The decision was to change the return type rather than add a separate builder-tagged overload (like `ep->amSend(buf, len, type, ucxx::builder)`) because the latter is more ergonomically disruptive to adopt the builder pattern.

The practical mitigation is that all existing code assigning to `std::shared_ptr<Request>` continues to compile unchanged via the builder's implicit conversion: the only callers affected are those using `auto`, where the compiler gives a clear error message pointing directly to where `.build()` is needed. The proper resolution is to promote the builder API out of `experimental::` soon, which would ideally happens in a future release. However, due to time constraints (mostly the project upstreaming), I don't want to delay making those inevitable breaking changes, so I'm thinking of making as many necessary breaking changes already in the current release cycle to avoid having to make breaking changes spread out through various release cycles.

The use of `experimental` namespace was more to follow the same initial pattern where I was still thinking through the limitations for `Context`/`Worker`, but I think they're fairly well designed and could be immediately promoted to non-experimental. If you think this is best, we can immediately move the `Request` builder constructors out of the experimental, I will most likely do all that in this version cycle anyway, but was expecting to hash out some other things first (like use from Python).

pentschev (commented):
Thanks @nirandaperera , I've addressed and/or responded your suggestions and questions.

cpp/include/ucxx/endpoint.h
pentschev: After more consideration, I realized I committed one fatal mistake, doing something like `auto req = endpoint->tagSend(...);` without explicitly calling `.build()` will not fail to build (which is what I expected) but will make the existing API essentially a no-op, that is effectively introducing a bug rather than a build failure, unless the user is casting directly to `std::shared_ptr<Request>` instead of `auto`.

With that said, I don't think this change is feasible, my hope was to reuse the `tagSend` (and others) method without requiring to rename them, but I don't think there's a good way to do so. Therefore, I decided to make two changes:
1. For this PR, revert the changes to methods that create `Request`s in `Endpoint`/`Worker`, essentially leaving it as before (no change to currently used API);
2. I will start another PR to add new `*Builder` request methods (like `endpoint->tagSendBuilder()`) that do not break the existing API to introduce builder constructor-based requests, and then deprecate the existing API, thus providing a less risky migration path to existing users.

madsbk (approved):
Looks good, a few nits:

* Mark all `create*` factory functions (and ideally `build()`) `[[nodiscard]]`.
* Mark the PR as non-breaking?
* `build()` and the conversion operators are `const` but have significant side effects, which I think is a bit misleading. I suggest making them non-`const`.
* `markBuilt()` consumes the builder before validation, so a `build()` that throws `ucxx::Error` (for example due to a null component) leaves the builder spent, and a retry throws `std::logic_error` instead. I think this is OK but it would be good with a note in the docstring.
* Maybe also note that the builders are not thread-safe

pentschev:
Thanks so much for the review @madsbk . I've addressed your suggestions, see below.

> Looks good, a few nits:
> 
> * Mark all `create*` factory functions (and ideally `build()`) `[[nodiscard]]`.

Good idea, done in https://github.com/rapidsai/ucxx/pull/677/commits/42a066abe1c2bd12c5e1492752c877c335f3b4dc.

> * Mark the PR as non-breaking?

To confirm, you really meant marking it _non-breaking_, right? It was already the case (it was previously marked _breaking_, but the breaking changes have since been removed). In any case I re-marked it _non-breaking_, in case you were seeing any GH weirdness.

> * `build()` and the conversion operators are `const` but have significant side effects, which I think is a bit misleading. I suggest making them non-`const`.

Excellent catch! I just recently changed made this modification and didn't really think of those side effects. Fixed now in  https://github.com/rapidsai/ucxx/pull/677/commits/08eaf4fc1c7fb896e6849f4bf2e9f3f48c09ddda. I went ahead and applied similar changes to `ContextBuilder`/`WorkerBuilder` to make them consistent, even though they do not strictly have the same non-`const` requirements (at least not today).

> * `markBuilt()` consumes the builder before validation, so a `build()` that throws `ucxx::Error` (for example due to a null component) leaves the builder spent, and a retry throws `std::logic_error` instead. I think this is OK but it would be good with a note in the docstring.

Done in https://github.com/rapidsai/ucxx/pull/677/commits/b0b4f4d1326348d641369e883e26ab119bd138e2.

> * Maybe also note that the builders are not thread-safe

Done in https://github.com/rapidsai/ucxx/pull/677/commits/22d12b7bd05bdaba32b56948fc56b96d0efcff48.

madsbk:
> To confirm, you really meant marking it _non-breaking_, right? It was already the case (it was previously marked _breaking_, but the breaking changes have since been removed). In any case I re-marked it _non-breaking_, in case you were seeing any GH weirdness.

Yes. I thought I saw it marked as `breaking` before, but now it shows up as `non-breaking` for me as well :)

PR #678, opened 2026-06-09
Backport fixes for Velox to release/0.50

bdice:
This PR backports fixes needed in Velox to `release/0.50`. A hotfix release is not required, we only need a branch based on `release/0.50` until Velox is updated to use CMake 4.x in its CI.

This reverts commit babdcfa0e5c26ebec146c12088b669efbecf1b95.


pentschev (commented):
This is why @wence- is always wary of CMake changes, especially the hard pins that we make. Is there a way we don't need to pin that hard on the lower end?

PR #680, opened 2026-06-11
Add explicit request builder APIs

pentschev:
Adds explicit `*Builder()` entry points for `Endpoint` and `Worker` request creation while preserving the existing legacy request APIs. Builder methods now take only required operation inputs, with optional settings such as Python futures, callbacks, and AM receiver callback metadata being configured through fluent builder setters before `.build()` or implicit conversion. Tests cover builder return types, rejected optional-argument call shapes, move-only behavior, and single-use enforcement.

Depends on #677 .


cpp/src/endpoint.cpp
```diff
@@ -614,6 +679,13 @@ std::shared_ptr<Request> Endpoint::streamSend(const void* const buffer,
     createRequestStream(endpoint, data::StreamSend(buffer, length), enablePythonFuture));
 }
 
+experimental::RequestStreamBuilder Endpoint::streamSendBuilder(const void* const buffer,
+                                                               size_t length)
+{
+  auto endpoint = std::dynamic_pointer_cast<Endpoint>(shared_from_this());
```
nirandaperera: Nit. casting failure is not tested. If this is guaranteed to be an `Endpoint`, why not use static_pointer_cast?

nirandaperera: Maybe `enabled_shared_from_this` needs to used at each child component level, and not in `Component` class.

cpp/tests/request_builders.cpp
```diff
@@ -65,6 +67,159 @@ void assertConversionRequiresNonConstBuilder()
...
+    ucxx::RequestCallbackUserData{}))>> : std::true_type {};
+
+template <typename T, typename = void>
+struct HasWorkerFlushBuilderFutureArg : std::false_type {};
+
+template <typename T>
+struct HasWorkerFlushBuilderFutureArg<T,
+                                      std::void_t<decltype(std::declval<T&>().flushBuilder(false))>>
+  : std::true_type {};
+
```
nirandaperera, line 222: Lets use concepts here? 
```suggestion
template <typename T>
concept HasEndpointAmSendBuilderReceiverCallbackArg = requires(T& t) {
  t.amSendBuilder(
    static_cast<const void*>(nullptr),
    size_t{},
    UCS_MEMORY_TYPE_HOST,
    std::optional<ucxx::AmReceiverCallbackInfo>{});
};

template <typename T>
concept HasEndpointAmRecvBuilderFutureArg = requires(T& t) {
  t.amRecvBuilder(false);
};

template <typename T>
concept HasEndpointMemPutBuilderRemoteKeyOffsetArg = requires(T& t) {
  t.memPutBuilder(
    static_cast<const void*>(nullptr), size_t{}, std::shared_ptr<ucxx::RemoteKey>{}, uint64_t{});
};

template <typename T>
concept HasEndpointMemGetBuilderRemoteKeyOffsetArg = requires(T& t) {
  t.memGetBuilder(
    static_cast<void*>(nullptr), size_t{}, std::shared_ptr<ucxx::RemoteKey>{}, uint64_t{});
};

template <typename T>
concept HasEndpointStreamSendBuilderFutureArg = requires(T& t) {
  t.streamSendBuilder(static_cast<const void*>(nullptr), size_t{}, false);
};

template <typename T>
concept HasEndpointStreamRecvBuilderFutureArg = requires(T& t) {
  t.streamRecvBuilder(static_cast<void*>(nullptr), size_t{}, false);
};

template <typename T>
concept HasEndpointTagSendBuilderCallbackArgs = requires(T& t) {
  t.tagSendBuilder(static_cast<const void*>(nullptr),
                   size_t{},
                   ucxx::Tag{},
                   false,
                   ucxx::RequestCallbackUserFunction{},
                   ucxx::RequestCallbackUserData{});
};

template <typename T>
concept HasEndpointTagRecvBuilderCallbackArgs = requires(T& t) {
  t.tagRecvBuilder(static_cast<void*>(nullptr),
                   size_t{},
                   ucxx::Tag{},
                   ucxx::TagMask{},
                   false,
                   ucxx::RequestCallbackUserFunction{},
                   ucxx::RequestCallbackUserData{});
};

template <typename T>
concept HasEndpointTagMultiSendBuilderFutureArg = requires(T& t) {
  t.tagMultiSendBuilder(
    std::declval<const std::vector<const void*>&>(),
    std::declval<const std::vector<size_t>&>(),
    std::declval<const std::vector<int>&>(),
    ucxx::Tag{},
    false);
};

template <typename T>
concept HasEndpointTagMultiRecvBuilderFutureArg = requires(T& t) {
  t.tagMultiRecvBuilder(ucxx::Tag{}, ucxx::TagMask{}, false);
};

template <typename T>
concept HasEndpointFlushBuilderFutureArg = requires(T& t) {
  t.flushBuilder(false);
};

template <typename T>
concept HasWorkerTagRecvBuilderCallbackArgs = requires(T& t) {
  t.tagRecvBuilder(static_cast<void*>(nullptr),
                   size_t{},
                   ucxx::Tag{},
                   ucxx::TagMask{},
                   false,
                   ucxx::RequestCallbackUserFunction{},
                   ucxx::RequestCallbackUserData{});
};

template <typename T>
concept HasWorkerTagRecvWithHandleBuilderCallbackArgs = requires(T& t) {
  t.tagRecvWithHandleBuilder(
    static_cast<void*>(nullptr),
    std::shared_ptr<ucxx::TagProbeInfo>{},
    false,
    ucxx::RequestCallbackUserFunction{},
    ucxx::RequestCallbackUserData{});
};

template <typename T>
concept HasWorkerFlushBuilderFutureArg = requires(T& t) {
  t.flushBuilder(false);
};
```

nirandaperera (commented):
some cosmetic suggestions

cpp/tests/request_builders.cpp
nirandaperera: Ah! ucxx is stil c++17 isnt it?

cpp/src/endpoint.cpp
pentschev: Changed the explicit builder methods to use `static_pointer_cast` for these self-casts in https://github.com/rapidsai/ucxx/pull/680/commits/f28dd1423a1e217b3ca143c7515c93c8ec1d3683 (cases added in this PR) and https://github.com/rapidsai/ucxx/pull/680/commits/6a5b64fba6c803a2d1ab40e3156106f3822f8983 (extending to pre-existing cases).

The failure path is not something we can meaningfully test through the supported API: these casts happen inside `Endpoint` / `Worker` member functions, so once `shared_from_this()` succeeds, the dynamic type is already an invariant of the member function being executed. Using `static_pointer_cast` makes that invariant explicit and avoids carrying an unchecked nullable result.

cpp/tests/request_builders.cpp
pentschev: Yes, it's C++17 and I will intentionally not plan on changing that until after upstreaming. Once that's done we might raise discussions on whether we want to remain on an older C++ version (to allow broader usability) or switch to a modern path.

pentschev (commented):
Thanks @nirandaperera , I've addressed your comments, please take another look.

madsbk (approved):
LGTM

pentschev:
Thanks for reviews!

PR #682, opened 2026-06-15
Fix worker tag drain request completion

fallintoplace:
## Summary
- store drain receive completion status in callback user data instead of the UCP request storage
- poll pending drain receives with `ucp_request_check_status()`
- free pending UCP requests after completion

## Why
When `ucp_tag_msg_recv_nbx()` returns a request pointer, `UCS_PTR_STATUS(status)` keeps treating that pointer as `UCS_INPROGRESS`. The old callback also wrote the completion status into the UCP request memory itself. That can corrupt the request and leave the drain loop spinning forever.

## Validation
- `git diff --check`
- Tried `cmake -S cpp -B cpp/build/local-check -G Ninja -DBUILD_TESTS=OFF -DBUILD_BENCHMARKS=OFF -DBUILD_EXAMPLES=OFF -DUCXX_ENABLE_CCCL=OFF`; configure is blocked locally because UCX is not installed (`ucxConfig.cmake` / `ucx-config.cmake` not found).

### DIFF
--- a/cpp/src/worker.cpp
+++ b/cpp/src/worker.cpp
@@ -80,14 +80,6 @@ Worker::Worker(std::shared_ptr<Context> context,
   setParent(std::dynamic_pointer_cast<Component>(context));
 }
 
-static void _drainCallback(void* request,
-                           ucs_status_t status,
-                           const ucp_tag_recv_info_t* /* info */,
-                           void* /* arg */)
-{
-  *reinterpret_cast<ucs_status_t*>(request) = status;
-}
-
 void Worker::drainWorkerTagRecv()
 {
   auto context = std::dynamic_pointer_cast<Context>(_parent);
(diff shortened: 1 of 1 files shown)


pentschev (changes requested):
Thanks for the contribution @fallintoplace .

After reviewing, I think this can be simplified a bit further. I think the callback/user_data path is not at all required for completion here.

`ucp_request_check_status()` is the proper way to poll a non-blocking request, and `UCS_PTR_STATUS(status)` intentionally returns `UCS_INPROGRESS` for valid request pointers. So I don't think this drain path needs a callback at all.

Could you remove `_drainCallback` and use something like this instead?

```cpp
ucp_request_param_t param = {.op_attr_mask = UCP_OP_ATTR_FIELD_DATATYPE,
                             .datatype     = ucp_dt_make_contig(1)};

ucs_status_ptr_t status =
  ucp_tag_msg_recv_nbx(_handle, buf.data(), info.length, message, &param);

if (UCS_PTR_IS_PTR(status)) {
  while (ucp_request_check_status(status) == UCS_INPROGRESS)
    progress();

  ucp_request_free(status);
} else if (UCS_PTR_IS_ERR(status)) {
  ucxx_debug("ucxx::Worker::%s failed to drain tag receive: %s",
             __func__,
             ucs_status_string(UCS_PTR_STATUS(status)));
}
```

That keeps the actual fix focused on the important pieces: detect request pointers with `UCS_PTR_IS_PTR()`, poll them with `ucp_request_check_status()`, and release them with `ucp_request_free()`.

pentschev (approved):
Thanks @fallintoplace !

PR #683, opened 2026-06-15
Avoid retaining listener tracebacks in UCXX log records

pentschev:
Pytest 9.1.0 now captures logs from non-propagating loggers, including UCXX's `ucx` logger. Listener errors logged with `logger.exception(...)` therefore leave traceback objects in captured `LogRecord.exc_info`, which can keep listener frames and UCXX context objects alive until pytest teardown completes.

This updates listener exception logging to format the traceback into the log message instead of storing it in `LogRecord.exc_info`. The visible traceback is preserved, but pytest no longer retains traceback frame references through fixture teardown.

Changes included:
- Add a helper for listener exception logging that preserves traceback text without retaining traceback objects.
- Use that helper for listener callback failures and listener handler failures.
- Add regression coverage for peer-info exchange failures and listener callback failures, verifying:
  - the exception text is still logged
  - `LogRecord.exc_info` is not retained
  - the listener context can be garbage collected

### DIFF
--- a/python/ucxx/ucxx/_lib_async/listener.py
+++ b/python/ucxx/ucxx/_lib_async/listener.py
@@ -6,6 +6,7 @@ import inspect
 import logging
 import os
 import threading
+import traceback
 import weakref
 
 import ucxx._lib.libucxx as ucx_api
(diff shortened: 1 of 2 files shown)


TomAugspurger (approved):

pentschev:
Thanks Tom!

PR #684, opened 2026-06-15
Fix progress thread stop timeout handling

fallintoplace:
## Summary
- make the progress-thread stop state atomic so it can be safely set from the caller
- keep the existing pre/post shutdown synchronization path
- if the stop-setting post callback times out, request a direct stop, repeatedly wake the worker, and fail explicitly if the thread still does not exit

## Why
`WorkerProgressThread::stop()` could cancel the timed-out post callback and then call `join()` even though `_stop` was never set. In that state the progress loop can continue forever and the caller blocks indefinitely.

## Validation
- `git diff --check`
- Tried `cmake -S cpp -B cpp/build/local-check -G Ninja -DBUILD_TESTS=OFF -DBUILD_BENCHMARKS=OFF -DBUILD_EXAMPLES=OFF -DUCXX_ENABLE_CCCL=OFF`; configure is blocked locally because UCX is not installed (`ucxConfig.cmake` / `ucx-config.cmake` not found).

### DIFF
--- a/cpp/include/ucxx/worker_progress_thread.h
+++ b/cpp/include/ucxx/worker_progress_thread.h
@@ -4,6 +4,7 @@
  */
 #pragma once
 
+#include <atomic>
 #include <functional>
 #include <memory>
 #include <mutex>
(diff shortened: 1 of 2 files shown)


pentschev (commented):
@fallintoplace this is a real bug, but I'm wondering if you can share more about your real use case. When did you encounter such bug, can you share a reproducer? I think your fix is a bit over-complicated, for example the `finished` flag and `std::terminate` are overreaching. Additionally, you have added no tests.

I have also been working on this and https://github.com/rapidsai/ucxx/pull/687 is my take to fix it without the parts that are unnecessary as described above, plus adding a test.

PR #686, opened 2026-06-16
Clear listener references before distributed-ucxx teardown

pentschev:
Fix a UCXX reset race in distributed-ucxx tests by clearing listener-handler local references before marking the listener client handler inactive.

Previously, `_listener_handler_coroutine` could decrement `active_clients` while its frame still held references to `ctx`, `conn_request`, `endpoint`, and `ep`. Tests waiting on `listener.active_clients == 0` could then proceed to `ucxx.reset()` while the listener coroutine frame still retained the `ApplicationContext`, causing "Trying to reset UCX but not all Endpoints and/or Listeners are closed".

This updates the listener cleanup path so those references are cleared before `active_clients.dec()`, including error paths such as `UCXMessageTruncatedError` followed by `endpoint.raise_on_error()`.

With the reset issue addressed, this also removes the distributed-ucxx flaky retry markers that were rerunning tests only for that UCX reset error. Stress-testing in CI (10 runs of all jobs in both `conda-python-distributed-tests` and `wheel-tests-distributed-ucxx`) showed no cases of the original failure occurred, signaling strongly of a real fix of the problem.


python/ucxx/ucxx/_lib_async/tests/test_listener.py
```diff
@@ -119,3 +132,42 @@ def noop_callback(ep):
...
+
+    await listener_mod._listener_handler_coroutine(
+        conn_request=_ConnectionRequest(),
+        ctx=_Context(),
+        func=lambda ep: None,
+        endpoint_error_handling=True,
+        connect_timeout=1,
+        ident=1,
+        active_clients=_ActiveClients(on_dec=check_listener_frame),
+    )
```
wence-, line 173: Hmm, very delicate.

wence- (approved):

python/ucxx/ucxx/_lib_async/tests/test_listener.py
pentschev: I'm assuming you're referring to the test strategy, and not necessarily about the functionality being covered here (although that could probably use some future re-evaluation too), but please correct me if I'm wrong.

Yes, agreed. The test is intentionally white-box because the bug depends on the exact ordering: `active_clients.dec()` must not run while the listener coroutine frame still holds UCXX/context references. A normal post-return weakref assertion would miss that ordering.

I can try to make this less frame-introspective by asserting weakrefs are cleared from inside the fake `dec()` callback instead of inspecting `f_locals`, but the test still needs to observe state at decrement time.

Let me know what you think.

pentschev:
@wence- thanks for the review. I'll go ahead and merge it as is but I can revisit the test discussion if you think it's worth it.

PR #687, opened 2026-06-22
Fix progress thread shutdown hang after stop callback timeout

pentschev:
Fix an intermittent progress-thread shutdown hang that can occur when `stopProgressThread()` is called before the progress thread reaches the generic post callback that normally sets the stop flag.

Previously, if the stop post callback timed out and was canceled before running, `stop()` could still call `join()` while `_stop` remained false. Once the progress thread resumed, it could continue running indefinitely, leaving teardown stuck.

Supersedes and closes #684

### DIFF
--- a/cpp/include/ucxx/worker_progress_thread.h
+++ b/cpp/include/ucxx/worker_progress_thread.h
@@ -5,6 +5,7 @@
 #pragma once
 
 #include <atomic>
+#include <cstdint>
 #include <functional>
 #include <memory>
 #include <mutex>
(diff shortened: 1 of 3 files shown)


cpp/include/ucxx/worker_progress_thread.h
```diff
@@ -46,9 +48,31 @@ typedef void* ProgressThreadStartCallbackArg;
...
+   */
+  struct StopConfig {
+    uint64_t callbackTimeoutNs;  ///< Maximum time to wait for each stop callback.
+    uint64_t signalIntervalNs;   ///< Interval for waking the worker while waiting.
+
+    /**
+     * @brief Create stop synchronization configuration.
+     *
+     * @param[in] callbackTimeoutNs maximum time in nanoseconds to wait for each stop callback.
+     * @param[in] signalIntervalNs interval in nanoseconds for waking the worker while waiting.
```
madsbk, line 63:
```suggestion
     * @param[in] signalIntervalNs interval in nanoseconds for waking the worker while waiting.
     *                             A value of 0 disables periodic wakeups: the worker is signaled
     *                             once before each wait and not again until the wait returns.
```

madsbk (approved):

wence-:
> Supersedes and closes https://github.com/rapidsai/ucxx/pull/687

This _is_ #687?

pentschev:
> > Supersedes and closes #687
> 
> This _is_ #687?

Fixed, was supposed to be #684.

wence- (approved):

fallintoplace (approved):

PR #692, opened 2026-07-01
Promote UCXX builders to stable API

pentschev:
Move the builder APIs out of the experimental namespace and make their headers and sources part of the stable `ucxx` namespace.

Standardize the public builder entry points on the `*Builder` naming scheme, expose parent-scoped builder helpers for child object creation, and update the C++ tests and includes to use the stable paths.

PR #693, opened 2026-07-01
Use builders from Python bindings

pentschev:
Switch the Cython declarations and Python wrapper implementation to call the stable builder APIs instead of the legacy non-builder constructors.

This keeps UCXX's Python layer on the replacement construction path before the old constructors are deprecated/removed.


TomAugspurger (approved):
Python / cython changes look good.

nirandaperera (approved):
After #692 is merged, this will be trivial (I think)

wence- (approved):

pentschev:
Thanks all for the reviews!

PR #694, opened 2026-07-01
Deprecate old constructors

pentschev:
Mark the legacy non-builder constructor APIs as deprecated and point each warning at the matching builder replacement.

The deprecation text states that these constructors will be removed in UCXX 0.52, and internal build targets define `UCXX_DISABLE_DEPRECATED_NON_BUILDER_CONSTRUCTOR_WARNINGS` where tests and examples still intentionally exercise the old API (to be also removed in UCXX 0.52).


nirandaperera (approved):
LGTM

wence- (approved):
Approving cmake

pentschev:
Thanks Niranda and Lawrence!

PR #699, opened 2026-07-03
Remove deprecations from friend declarations

nirandaperera:
Having `deprecated` on `friend` declarations are not allowed in clang.


nirandaperera:
It seems like clang is not in the dependencies in ucxx. So, I will add a `cpp_linters.sh` CI step separately in another PR.      This will prevent issues like this in the future.

bdice (approved):

pentschev:
Thanks Niranda!

pentschev (approved):

PR #701, opened 2026-07-06
test against oldest cuda-core, refactor testing dependencies configuration

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/285

This project is already not depending on `cuda-python` and carrying the same `cuda-core` floor I'm proposing across RAPIDS (`>=0.5.1`).

This proposes some other changes based on things I noticed while investigating that:

* adds explicit testing against the oldest versions of `cuda-core` and `numpy` the project supports (ref: https://github.com/rapidsai/build-planning/issues/81)
* refactors `dependencies.yaml` to reduce duplication between `ucxx` and `distributed-ucxx` testing dependencies
* cleans up some dependencies (see inline comments)

### DIFF
--- a/conda/recipes/ucxx/recipe.yaml
+++ b/conda/recipes/ucxx/recipe.yaml
@@ -238,7 +238,6 @@ outputs:
         - setuptools>=77.0.0
         - wheel
       run:
-        - cuda-core >=0.5.1
         - python
         - pyyaml >=6
         - rapids-dask-dependency ${{ rapids_version }}
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -22,10 +22,11 @@ files:
       - run_python_ucxx
       - run_python_distributed_ucxx
       - test_cpp
+      - test_python_common
       - test_python_ucxx
-      - test_python_distributed_ucxx
       - depends_on_cudf
       - depends_on_cupy
+      - depends_on_numba_cuda
       - depends_on_rmm
       - depends_on_ucx_run
   test_cpp:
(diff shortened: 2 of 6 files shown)


conda/recipes/ucxx/recipe.yaml
```diff
@@ -238,7 +238,6 @@ outputs:
         - setuptools>=77.0.0
         - wheel
       run:
-        - cuda-core >=0.5.1
```
jameslamb, line 241: `distributed-ucxx` does not use `cuda-core` directly at all. Only `ucxx` does.

```console
$ git grep -E 'from cuda.*import|import cuda'
python/ucxx/ucxx/_cuda_context.py:        from cuda.core import Device
python/ucxx/ucxx/_cuda_context.py:            from cuda.core.experimental import Device
```

dependencies.yaml
```diff
@@ -373,74 +384,46 @@ dependencies:
...
-              - &numba_cuda_cu12_test numba-cuda[cu12]>=0.22.1
-          - matrix:
-              cuda: "13.*"
-              cuda_suffixed: "true"
+              dependencies: "oldest"
             packages:
-              - &numba_cuda_cu13_test numba-cuda[cu13]>=0.22.1
-          # fallback to numba-cuda with no extra CUDA packages if 'cuda_suffixed' isn't true
+              - cuda-core==0.5.1
+              - numpy==2.0.*
```
jameslamb, line 411: A little hard to see in the diff, but this is constraining `cuda-core` and `numpy` to the oldest versions (referenced in this project's pinnings) in the `oldest-deps` CI jobs.

Similar to how other RAPIDS projects do this.

python/distributed-ucxx/pyproject.toml
jakirkham, line 26: Since `ucxx` is the only using `cuda-core` and already depends on it. There is no need for the `cuda-core` dependency here either (hence why it is dropped)

jakirkham (approved):
Thanks James! 🙏

Made one observational note for other readers (no action needed)

PR #703, opened 2026-07-07
Support binding listener to a specific IP address or hostname

### LINKED ISSUE #299: Allow specifying bind address to `ucxx::Listener`
Currently the `ucxx::Listener` constructor binds exclusively to all interface (`0.0.0.0`). It should be possible for the user to specify the IP address where to bind to.

rhua-nv:
Resolves #299 

## Description

Until now `ucxx::Listener` always bound to the wildcard address (all interfaces): the private `Listener` constructor called `ucxx::utils::get_addrinfo(NULL, port)` with a hardcoded `NULL` address, even though `get_addrinfo()` fully supports resolving a hostname or IP address.

This PR adds an optional `ipAddress` option to `ListenerBuilder`, allowing the listener to be restricted to a specific interface:

```cpp
auto listener = worker->listenerBuilder(port, callback, callbackArgs)
                  .ipAddress("127.0.0.1")
                  .build();
```

An empty address (the default) preserves the existing wildcard behavior, so this is fully backward compatible. Invalid addresses throw `ucxx::Error` (raised by the existing `get_addrinfo` error path).

Motivation: services embedding ucxx (e.g. NVIDIA Isaac Sim's UCX bridge) need a secure-by-default loopback-only listener. Without this option, the only alternative is bypassing ucxx and managing raw `ucp_listener_h` / `ucp_ep_h` handles directly, which loses ucxx's endpoint lifetime management.

Changes:
- `ListenerBuilder::ipAddress(std::string)` builder option.
- `detail::createListener` / private `Listener` constructor take the address and pass it to `get_addrinfo`.
- Deprecated non-builder `createListener` APIs are unchanged (wildcard).
- Tests: end-to-end bind-to-loopback send/recv, `getIp()` verification, invalid-address throw, and builder-chaining static checks.

The async/`_lib` Python APIs could expose the same option; left out of scope here to keep the change small, happy to do a follow-up PR if there is interest.

### DIFF
--- a/cpp/src/listener.cpp
+++ b/cpp/src/listener.cpp
@@ -18,6 +18,7 @@
 namespace ucxx {
 
 Listener::Listener(std::shared_ptr<Worker> worker,
+                   std::string ipAddress,
                    uint16_t port,
                    ucp_listener_conn_callback_t callback,
                    void* callbackArgs)
@@ -28,7 +29,7 @@ Listener::Listener(std::shared_ptr<Worker> worker,
   ucp_listener_params_t params = {
     .field_mask   = UCP_LISTENER_PARAM_FIELD_SOCK_ADDR | UCP_LISTENER_PARAM_FIELD_CONN_HANDLER,
     .conn_handler = {.cb = callback, .arg = callbackArgs}};
-  auto info               = ucxx::utils::get_addrinfo(NULL, port);
+  auto info = ucxx::utils::get_addrinfo(ipAddress.empty() ? nullptr : ipAddress.c_str(), port);
   params.sockaddr.addr    = info->ai_addr;
   params.sockaddr.addrlen = info->ai_addrlen;
 
@@ -68,11 +69,13 @@ Listener::~Listener()
 namespace detail {
 
 std::shared_ptr<Listener> createListener(std::shared_ptr<Worker> worker,
+                                         std::string ipAddress,
                                          uint16_t port,
                                          ucp_listener_conn_callback_t callback,
                                          void* callbackArgs)
 {
-  return std::shared_ptr<Listener>(new Listener(worker, port, callback, callbackArgs));
+  return std::shared_ptr<Listener>(
+    new Listener(worker, std::move(ipAddress), port, callback, callbackArgs));
 }
 
 }  // namespace detail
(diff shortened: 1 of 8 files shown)


cpp/src/listener.cpp
```diff
@@ -82,7 +85,7 @@ std::shared_ptr<Listener> createListener(std::shared_ptr<Worker> worker,
                                          ucp_listener_conn_callback_t callback,
                                          void* callbackArgs)
 {
-  return detail::createListener(std::move(worker), port, callback, callbackArgs);
+  return detail::createListener(std::move(worker), std::string{}, port, callback, callbackArgs);
 }
```
wence-, line 89: question: Why not also expose the specific address version in the public interface?

cpp/src/listener.cpp
```diff
@@ -68,11 +69,13 @@ Listener::~Listener()
 namespace detail {
 
 std::shared_ptr<Listener> createListener(std::shared_ptr<Worker> worker,
+                                         std::string ipAddress,
```
wence-: Is this only an ipAddress, or anything that getaddrinfo can resolve?

wence- (changes requested):

cpp/src/listener.cpp
wence-: Ah, they're deprecated, so no need.

cpp/src/listener.cpp
rhua-nv: Anything that `getaddrinfo` can resolve. I renamed `ipAddress` to `host`

cpp/tests/component_builders.cpp
```diff
@@ -531,6 +531,20 @@ TEST_F(ComponentBuilderTest, ListenerBuilder)
   ASSERT_GT(listener->getPort(), 0u);
 }
 
+TEST_F(ComponentBuilderTest, ListenerBuilderHost)
+{
+  auto builder = ucxx::ListenerBuilder(_worker, 0, listenerCallback, nullptr);
+  static_assert(
+    std::is_same<decltype(builder.host("127.0.0.1")), ucxx::ListenerBuilder&>::value,
+    "ListenerBuilder::host returns ListenerBuilder& for chaining");
+
```
wence-:
```suggestion
```

I don't think this is necessary, since otherwise `build` would not be available on the next line.

wence- (approved):
Tiny test cleanup suggestion. thanks

d4l3k:
@rhua-nv @wence- any plans to clean this up and land?

cpp/tests/component_builders.cpp
rhua-nv: removed

rhua-nv:
@wence- can you please review again?

wence- (approved):

pentschev (approved):
Thanks @rhua-nv . Just for everyone reading this, we're merging breaking changes to `createListener` that is a deprecated constructor. In the next few days that and all other deprecated `create*` constructor methods will be removed, as per the deprecation cycle that is to be fulfilled by UCXX 0.52. Therefore, please switch to using `ListenerBuilder` and other `*Builder`s.

PR #710, opened 2026-07-20
put a <2 ceiling on cuda-core dependency

jameslamb:
## Description

Follow-up to #701

Puts a ceiling of `<2` on the project's `cuda-core` dependency.

Starting with the 1.x series, `cuda-core` is following strict semantic versioning, and some folks working on the project recommended doing this defensively: https://github.com/rapidsai/build-planning/issues/285#issuecomment-5027243158

There isn't yet a 2.x release, this is just defensive to protect CI and users if there eventually is one.

PR #713, opened 2026-07-23
update most pre-commit hooks

jameslamb:
Updates most `pre-commit` hooks to their latest versions (skipped `clang-format` because it changes many files).

This is helpful on its own, but mainly opening this to test whether a CI failure I'm observing in https://github.com/rapidsai/ucxx/pull/710#issuecomment-5060226224 affects all PRs.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -325,9 +325,9 @@ dependencies:
           - dask-cuda==26.8.*,>=0.0.0a0
           - dask-cudf==26.8.*,>=0.0.0a0
           # UCX Build
-          - libtool
-          - automake
           - autoconf
+          - automake
+          - libtool
           # UCXX Build
           - pkg-config
           # Docs Build
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,18 +3,18 @@
 
 repos:
       - repo: https://github.com/pre-commit/pre-commit-hooks
-        rev: v5.0.0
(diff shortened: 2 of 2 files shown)


dependencies.yaml
```diff
@@ -325,9 +325,9 @@ dependencies:
           - dask-cuda==26.8.*,>=0.0.0a0
           - dask-cudf==26.8.*,>=0.0.0a0
           # UCX Build
-          - libtool
-          - automake
           - autoconf
+          - automake
+          - libtool
```
jameslamb, line 330: Just touching `.pre-commit-config.yaml` isn't enough here, because test jobs are skipped for PRs that only change that. So picked a tiny change in a more central file (alphabetizing this little list) to trigger a full run.

jameslamb:
First run had a single failure ([link](https://github.com/rapidsai/ucxx/actions/runs/30029399446/job/89284316572)) and that job passed on a re-run.

Between that and #710 CI eventually passing, I think we at least have enough evidence to say there isn't some deterministic CI-blocking issue happening.

We might as well just merge this since it already has the passing CI run, but if it ends up missing freeze no big deal, it was just for testing.

jameslamb:
> _if it ends up missing freeze no big deal, it was just for testing._

UCXX is now in code freeze and this didn't get a review. Don't think it's worth re-targeting and another CI run, I'll just close it.

PR #716, opened 2026-07-28
enforce 'yamllint' checks, fix verify-hardcoded-versions findings

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/305

Proposes enforcing `yamllint` checks here. My primary motivation is to catch correctness issues in `dependencies.yaml` files, like duplicate entries silently resolving to the last one or indentation mistakes leading to filters being ignored.

But this also has some side benefits for consistency, which makes it a bit easier to write automation.


cpp/include/ucxx/typedefs.h
```diff
@@ -20,12 +20,14 @@
 
 #include <ucp/api/ucp.h>
 
+// rapids-pre-commit-hooks: disable[verify-hardcoded-version]
```
jameslamb, line 23: Suppresses this unrelated warning that was blocking PR CI.

```text
verify-hardcoded-version.................................................Failed
- hook id: verify-hardcoded-version
- exit code: 1

In file cpp/include/ucxx/typedefs.h:27:79:
   [[deprecated(message " This non-builder constructor will be removed in UCXX 
0.52.")]]
warning: do not hard-code version, read from VERSION file instead

In file cpp/include/ucxx/typedefs.h:27:79:
   [[deprecated(message " This non-builder constructor will be removed in UCXX 
0.52.")]]
note: if this is intentional (as part of a docstring or deprecation notice), 
suppress it with rapids-pre-commit-hooks: disable-next-line - see 
https://github.com/rapidsai/pre-commit-hooks/blob/main/README.md#suppressing-fal
se-positives for details
```

([build link](https://github.com/rapidsai/ucxx/actions/runs/30385349709/job/90363206242?pr=716))

It's a small change, figured it'd be ok to just include here.

nirandaperera (approved):
LGTM

KyleFromNVIDIA (approved):

PR #727, opened 2026-08-18
Allow UCX 1.21.0

jakirkham:
Part of issue: https://github.com/rapidsai/build-planning/issues/314

Allow UCX 1.21.0 to be used with UCXX.


conda/environments/all_cuda-129_arch-aarch64.yaml
```diff
@@ -46,5 +46,5 @@ dependencies:
 - scikit-build-core>=0.11.0
 - setuptools>=77.0.0
 - sphinx>=8.1.0
-- ucx>=1.18.0,<1.21
+- ucx>=1.18.0,<1.22
```
pentschev: 1.22 is already out, so let's pin `<1.23`.

pentschev (changes requested):

conda/environments/all_cuda-129_arch-aarch64.yaml
jakirkham: Was holding off on this until we have it packaged in all formats

Doing this as an incremental step to update UCX consistently as we move towards the latest version

dependencies.yaml
```diff
@@ -662,11 +662,11 @@ dependencies:
           - matrix:
               cuda: "12.*"
             packages:
-              - ucx>=1.18.0,<1.21
+              - ucx>=1.18.0,<1.22
           # require at least 1.19.0 on CUDA 13 (and make that the default when 'cuda' filter is not provided)
```
pentschev: Let's bump minimum/build version to 1.19, then we can remove the different CUDA 12/13 handling.

pentschev (commented):
Unfortunately, it seems L4s are failing with either host-sided or device-sided OOMs, this looks related to UCX 1.21, since nightlies all succeeded. I don't immediately have any ideas in mind of what could have caused this, but I suspect there may either be a memory leak being exposed or UCX has somehow increased memory utilization. Given they're only reproducible in a certain instance type, it may be difficult to reproduce locally to debug as well. I do not have time to debug this immediately, but it is certainly we will need to tackle soon-ish since we will most likely need to upgrade to v1.22.1 (an unreleased version that  is planned to go out soon) still for RAPIDS 26.10.

pentschev:
I was able to reproduce the issue on a machine with a single L4, so not CI-exclusive.

pentschev:
I have found out that disabling CUDA IPC (`UCX_TLS=^cuda_ipc`) works around the problem, and it seems to be due to https://github.com/openucx/ucx/pull/10933 removing `UCX_CUDA_IPC_ENABLE_SAME_PROCESS=n`. Still investigating whether that's an intentional behavioral change or if it may be a bug.

jakirkham:
Thanks Peter! 🙏

Appreciate you doing that deep dive in UCX internals

Were you able to learn any more from your investigation about the intent of that change?

pentschev:
> Thanks Peter! 🙏
> 
> Appreciate you doing that deep dive in UCX internals
> 
> Were you able to learn any more from your investigation about the intent of that change?

Still on my to-do list, but had zero time since to work on it.

pentschev:
I have now identified the issue, see #733 . I have pushed https://github.com/rapidsai/ucxx/pull/727/commits/e011803553d22eb8355026a7a49398c87925ab00 with the workaround to this branch, if everything passes now we should be good to merge this.

pentschev (approved):
As discussed offline, let's then merge this as is and enable 1.21.0, and once we have wheels for 1.22.0 then we update the pin to support that too.

pentschev (approved):

jameslamb (approved):

jakirkham:
Thank you for looking into this Peter! Also thank you and James for the reviews! 🙏

PR #729, opened 2026-08-20
Prevent delayed requests from being submitted after cancelation

pentschev:
## Summary

Prevent a delayed UCXX request from being submitted after it has already been canceled.

This addresses intermittent hangs in `ucxx/_lib_async/tests/test_shutdown.py::test_client_shutdown[tag]`.

## Problem

During endpoint shutdown, a request can be canceled before its delayed-submission callback runs. At that point no UCP request exists, so the previous cancelation path did nothing. A later progress iteration could then submit the request after shutdown, leaving its Python future waiting indefinitely.

## Fix

Cancelation and submission are now serialized using the request mutex:

- Canceling before submission completes the request with `UCS_ERR_CANCELED`;
- `Request::populateDelayedSubmission()` skips submission for completed requests;
- Request-specific submission remains in private `populateDelayedSubmissionImpl()` overrides.

This ensures that either cancelation prevents submission or submission completes before normal UCP cancelation occurs.

Depends on #731

### DIFF
--- a/cpp/src/request.cpp
+++ b/cpp/src/request.cpp
@@ -1,5 +1,5 @@
 /**
- * SPDX-FileCopyrightText: Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES.
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: BSD-3-Clause
  */
 #include <chrono>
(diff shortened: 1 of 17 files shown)


cpp/src/request.cpp
```diff
@@ -87,7 +87,10 @@ void Request::cancel()
                        ucs_status_string(status));
     } else {
       ucxx_trace_req_f(_ownerString.c_str(), this, _request, _operationName.c_str(), "canceling");
-      if (_request != nullptr) ucp_request_cancel(_worker->getHandle(), _request);
+      if (_request != nullptr)
+        ucp_request_cancel(_worker->getHandle(), _request);
+      else
+        setStatus(UCS_ERR_CANCELED);
```
bdice:
```suggestion
      if (_request != nullptr) {
        ucp_request_cancel(_worker->getHandle(), _request);
      } else {
        setStatus(UCS_ERR_CANCELED);
      }
```

bdice (approved):

pentschev:
Thanks @bdice !

PR #730, opened 2026-08-20
Prevent address reuse in unreachable worker test

pentschev:
Fix intermittent failures in `test_from_worker_address_error[unreachable]`:

The test destroyed the server worker before initializing the client. This allowed the operating system to assign the server's released transport address to the new client worker. The supposedly stale address could therefore identify a live worker, causing the endpoint to remain alive until the test timed out.

A focused reproducer plausibly confirmed this mechanism: forcing TCP address reuse kept 3/3 stale connections alive, while preventing reuse made 20/20 connections report the expected unreachable error.

This change initializes the client before the server releases its transport resources.


madsbk (approved):

pentschev:
Thanks Mads!

PR #731, opened 2026-08-20
Receive queued tag messages after endpoint closure

pentschev:
## Summary

Allow `Endpoint.recv()` to recover a matching message already queued on the UCX worker when the endpoint closes or its receive is canceled.

The change reserves the queued message and receives it directly through the worker. It also:

- Uses the destination buffer length for handle-based receives.
- Safely consumes reserved message handles during cancellation.
- Synchronizes delayed submission to prevent duplicate request submission.

## Problem

A peer may close its endpoint after sending, leaving the message queued on the receiving worker. Previously, `Endpoint.recv()` could raise `UCXCloseError` or `UCXCanceled` even though the requested payload was available.

This change preserves that queued message and delivers it, while retaining cancellation when no matching message exists. Tests cover endpoint closure, undersized buffers, immediate completion, and cancellation behavior.

This is expected to resolve a problem exposed by #729 , therefore should be merged before that PR.

### DIFF
--- a/cpp/include/ucxx/worker.h
+++ b/cpp/include/ucxx/worker.h
@@ -1,5 +1,5 @@
 /**
- * SPDX-FileCopyrightText: Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES.
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: BSD-3-Clause
  */
 #pragma once
@@ -883,6 +883,19 @@ class Worker : public Component {
    * @returns Builder to configure optional parameters and submit the request.
    */
   [[nodiscard]] RequestTagBuilder tagRecvWithHandleBuilder(void* buffer,
+                                                           std::shared_ptr<TagProbeInfo> probeInfo);
+
+  /**
+   * @brief Create a builder for a bounded tag receive using a message handle.
+   *
+   * @param[in] buffer     a raw pointer to pre-allocated receive memory.
+   * @param[in] length     the size in bytes of the receive buffer.
+   * @param[in] probeInfo  the TagProbeInfo object containing message length and handle.
+   *
+   * @returns Builder to configure optional parameters and submit the request.
+   */
+  [[nodiscard]] RequestTagBuilder tagRecvWithHandleBuilder(void* buffer,
+                                                           size_t length,
                                                            std::shared_ptr<TagProbeInfo> probeInfo);
 
   /**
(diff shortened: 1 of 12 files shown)


cpp/include/ucxx/worker.h
```diff
@@ -885,6 +885,19 @@ class Worker : public Component {
   [[nodiscard]] RequestTagBuilder tagRecvWithHandleBuilder(void* buffer,
                                                            std::shared_ptr<TagProbeInfo> probeInfo);
 
+  /**
+   * @brief Create a builder for a bounded tag receive using a message handle.
```
wence-: What is a "bounded" tag receive?

cpp/tests/worker.cpp
```diff
@@ -37,6 +37,62 @@ struct ExtraParams {
...
+  bool released{false};
+};
+
+void waitForProgressThreadRelease(void* arg)
+{
+  auto& barrier = *static_cast<ProgressThreadStartBarrier*>(arg);
+  std::unique_lock<std::mutex> lock(barrier.mutex);
+  barrier.started = true;
+  barrier.condition.notify_one();
+  barrier.condition.wait(lock, [&barrier]() { return barrier.released; });
```
wence-, line 53: OK, so on the progress thread we start the barrier, notify a waiter and then wait for the barrier to be released. That drops the lock after checking and goes into a wait queue.

cpp/tests/worker.cpp
```diff
@@ -37,6 +37,62 @@ struct ExtraParams {
...
+
+class PausedProgressThread {
+ public:
+  explicit PausedProgressThread(std::shared_ptr<ucxx::Worker> worker) : _worker(worker)
+  {
+    _worker->setProgressThreadStartCallback(waitForProgressThreadRelease, &_barrier);
+    _worker->startProgressThread(true);
+
+    std::unique_lock<std::mutex> lock(_barrier.mutex);
+    _barrier.condition.wait(lock, [this]() { return _barrier.started; });
```
wence-, line 64: This waits until the progress thread has entered the barrier.

cpp/tests/worker.cpp
```diff
@@ -37,6 +37,62 @@ struct ExtraParams {
...
+  }
+
+  void resume()
+  {
+    {
+      std::lock_guard<std::mutex> lock(_barrier.mutex);
+      if (_barrier.released) return;
+      _barrier.released = true;
+    }
+    _barrier.condition.notify_one();
```
wence-, line 80: And this releases the progress thread.

wence- (approved):

cpp/include/ucxx/worker.h
pentschev: Bounded in this context meant the receive is limited by the explicitly supplied destination-buffer length. However, now I think this is a bad choice of words, so I rephrased without that term and explained better the truncation in the docstring, see https://github.com/rapidsai/ucxx/pull/731/commits/418e6d0524b6865843486cef699619179fd5a4cf.

pentschev:
Thanks Lawrence!

bdice (approved):
Packaging approval

PR #743, opened 2026-09-03
Remove deprecated C++ constructors for v0.52

pentschev:
Removes the deprecated C++ construction APIs as planned for UCXX 0.52.

Users should construct UCXX objects through the supported builder APIs. Internal construction support remains isolated from the public API, and examples and benchmarks now use the supported interfaces.

This is a breaking API change for code that still uses the removed `create*` constructors.

### DIFF
--- a/cpp/include/ucxx/context.h
+++ b/cpp/include/ucxx/context.h
@@ -1,5 +1,5 @@
 /**
- * SPDX-FileCopyrightText: Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES.
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: BSD-3-Clause
  */
 #pragma once
@@ -66,12 +66,13 @@ class Context : public Component {
   Context& operator=(Context&& o)    = delete;
 
   /**
-   * @brief Friend declaration for `ucxx::createContext` with parameters.
+   * @brief Allow the internal context factory to access the private constructor.
    *
-   * This friend declaration allows the standalone `ucxx::createContext` function to access
-   * the private constructor. See the public declaration for full documentation.
+   * This friend declaration allows `ucxx::detail::createContext` to access the private
+   * constructor.
    */
-  friend std::shared_ptr<Context> createContext(ConfigMap ucxConfig, const uint64_t featureFlags);
+  friend std::shared_ptr<Context> detail::createContext(ConfigMap ucxConfig,
+                                                        const uint64_t featureFlags);
 
   /**
    * @brief Allow ContextBuilder to access private constructor.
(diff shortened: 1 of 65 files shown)


cpp/include/ucxx/context.h
```diff
@@ -13,8 +13,8 @@
 
 #include <ucxx/component.h>
 #include <ucxx/config.h>
-#include <ucxx/constructors.h>
 #include <ucxx/context_builder.h>
+#include <ucxx/detail/constructors.h>
```
pentschev, line 17: For now, constructors remain technically part of the API, I did not want to make even more changes in this large PR. This is being simplified and made truly an implementation detail in the follow-up in https://github.com/rapidsai/ucxx/pull/744.

jameslamb (approved):

bdice:
I asked an agent to help me review. There are three things it suggested:
1. Missing a `RemoteKey` builder API for specific set of parameters -- do with this what you want. I don't know if this is important.
2. Clean up Cython -- this is a good idea, I recommend doing it in this PR.
3. Clean up docs -- this is a good idea, I recommend doing it in this PR.

_AI generated content below._

---------------------

### 1. [P2] Add a `RemoteKey`-based builder path for nonzero RMA offsets

**Locations:**
- `cpp/include/ucxx/endpoint.h:481-483`
- `cpp/include/ucxx/endpoint.h:517-519`
- Exposed by the migrations in `cpp/tests/request.cpp:1084-1089` and `cpp/tests/request.cpp:1171-1176`

The removed overloads supported all of the following together:

- a `std::shared_ptr<RemoteKey>` rather than a raw `ucp_rkey_h`;
- a nonzero `remoteAddrOffset`;
- builder-equivalent callback/future options.

The surviving `memPutBuilder(..., RemoteKey)` and `memGetBuilder(..., RemoteKey)` overloads always use the key's base address and provide no offset parameter or builder setter. The changed offset tests demonstrate the missing migration path: they now have to unwrap `RemoteKey` into `getBaseAddress() + offsetBytes` and `getHandle()` and select the raw-handle overload.

The operation remains technically possible, but the advertised builder replacement does not preserve the typed `RemoteKey` API and requires callers to bypass that abstraction. Since these methods have gone through a deprecation cycle specifically directing users to the builders, add either `RemoteKey` overloads that accept an offset or a `remoteAddressOffset()` option on `RequestMemBuilder`, then keep the offset tests on the `RemoteKey` path.

### 2. [P2] Remove deleted C++ members from the Cython declarations

**Location:** `python/ucxx/ucxx/_lib/ucxx_api.pxd:273-290`

The PR deletes `ucxx::Context::createWorker()` and `ucxx::Worker::getAddress()`, but the Cython declaration file still advertises both:

```cython
cdef cppclass Context(Component):
    shared_ptr[Worker] createWorker(...)

cdef cppclass Worker(Component):
    shared_ptr[Address] getAddress()
```

This does not fail the current Python build because neither declaration is presently referenced by a `.pyx` file; Cython only emits a C++ member call when the declaration is used. Any future in-tree use—or downstream Cython code using the shipped declaration—will generate calls to members that no longer exist and fail at C++ compilation.

Remove these declarations in the same PR. The binding already constructs workers through `ucxx::python::createWorker`, and address construction can be represented by the existing `AddressBuilder` declaration (or by declaring the corresponding builder-returning member if needed).

### 3. [P3] Update public documentation that still names the removed methods

**Locations:**
- `cpp/include/ucxx/endpoint.h:396-649`
- `cpp/include/ucxx/worker.h:786,805,973,997-998`
- `cpp/include/ucxx/memory_handle.h:159-160`
- `cpp/include/ucxx/remote_key.h:202`

The builder documentation repeatedly says that conversion invokes the same path as removed methods such as `amSend()`, `tagRecv()`, `flush()`, and `close()`. `Worker::queryAttributes()` still tells users to call removed `getAddress()`, and the RMA documentation still directs users to removed `Endpoint::memPut()` / `memGet()` methods.

After this PR those names are no longer valid public API or useful Doxygen targets. Reword the builder text to say that conversion submits/builds the request, and replace the remaining usage references with `addressBuilder()`, `memPutBuilder()`, and `memGetBuilder()`.

## Validation performed

- `git diff --check upstream/main...upstream/pr-743` — passed.
- Configured and built libucxx, the C++ tests, examples, and benchmarks with CMake 4.4.2, GCC 14.4, UCX 1.20.1, CUDA 13.3, and CCCL enabled — passed.
- `ctest --test-dir cpp/build/review5 --output-on-failure` with the UCX 1.20.1 runtime — passed (the `UCXX_TEST` target completed in 181.54 seconds).
- `ucxx_example_basic -P polling -p 23456` — completed successfully.
- `ucxx_perftest -l -t tag_lat -m host -P polling -n 10 -w 1 -v` — completed its validated loopback iterations; UCX logged connection-reset messages during endpoint teardown.
- GitHub's build, style, docs, conda, wheel, and test checks were all successful for the reviewed head SHA.

## Additional note

The new `cpp/include/ucxx/detail/constructors.h` is still installed and included by public headers, so factory declarations are not yet physically isolated from the installed header surface. The author already called this out in the PR discussion and linked follow-up [#744](https://github.com/rapidsai/ucxx/pull/744); I did not count it as a new finding here. If #743 can ship independently of that follow-up, the PR description should avoid claiming complete isolation until #744 lands.

bdice (approved):
All seems fine. I will help with PRs for the 3 items above.

bdice (approved):
Re-approving for C++

pentschev:
Thank you both!

PR #744, opened 2026-09-03
Simplify internal constructor access

### LINKED ISSUE #435: Better C++ API for constructors (and methods that return new `ucxx` objects)

pentschev:
Replace per-class factory friendships with a single internal constructor factory.

This keeps construction details out of the public API while preserving builder-based object creation and current listener host-binding support.

Closes https://github.com/rapidsai/ucxx/issues/435 .

Note for reviewers: this PR is based off of #743 , the novel changes are only https://github.com/rapidsai/ucxx/pull/744/commits/daf0e42ce40835d5a375f3fe9f3eb4cc18922597 and https://github.com/rapidsai/ucxx/pull/744/commits/51eedc844b9c8630508c3f96bc3d15a977ebe2be.


cpp/src/address.cpp
```diff
@@ -8,6 +8,7 @@
 #include <string_view>
 #include <utility>
 
+#include <internal/constructors.h>
```
bdice, line 11: In cudf we use double quotes for relative includes from src like this. Not sure how ucxx handles includes.

bdice (approved):

cpp/src/address.cpp
pentschev: I don't think we have it formally defined but we have been using `<>` for internals, it doesn't really seem to have a benefit one way or the other, so I'll just keep it as-is for now.

pentschev:
Thanks for the review @bdice !

PR #746, opened 2026-09-03
Remove deleted C++ methods from Cython declarations

bdice:
Depends on #743 and is based on its head commit. The novel changes are in 6ed1ba7.

Removes the stale Cython declarations for `Context::createWorker()` and `Worker::getAddress()`, which #743 removes from the C++ API. This prevents future Cython users from generating calls to members that no longer exist.

There is no runtime API impact.


jameslamb (approved):
approving for `packaging-codeowners` (packaging changes are minimial and non-controversial)

pentschev (approved):

PR #748, opened 2026-09-04
Make async listener teardown deterministic

pentschev:
Track listener handlers from submission through completion and wait for implicitly managed UCXX resources to be released before resetting the context. Cleanup is bounded and preserves the original test failure when fallback cleanup is necessary.

The previous implementation could miss handlers that were scheduled but had not started, allowing UCXX teardown to race with outstanding listener work. This closes that gap and improves diagnostics for remaining leaks, addressing the recurring context-reset failures observed on GB300.


python/ucxx/ucxx/_lib_async/listener.py
```diff
@@ -29,71 +30,93 @@ def _log_exception(message):
...
+    async def wait(self, progress=None) -> None:
+        """Wait until all handlers submitted so far have fully retired."""
+        while True:
+            with self._lock:
+                futures = tuple(self._futures)
+            if not futures:
+                return
+            if progress is not None:
+                progress()
+                await asyncio.sleep(1e-9)
```
pentschev: Why not 0?

.github/workflows/pr.yaml
pentschev, line 1: Explicit GB300 testing to PR was added for testing purposes, given this issue was being deterministically observed only in those instances. This change will be removed immediately prior to merging, but it's kept until the PR is approved in case more changes require testing.

python/ucxx/ucxx/_lib_async/listener.py
```diff
@@ -29,71 +29,90 @@ def _log_exception(message):
...
+            # The event loop may finish the coroutine before this call returns.
+            with self._lock:
+                future = asyncio.run_coroutine_threadsafe(coroutine, event_loop)
+                self._futures.add(future)
+        except BaseException:
+            coroutine.close()
+            raise
+
+        future.add_done_callback(
+            lambda completed: self._handler_done(completed, event_loop)
```
wence-, line 64: question: Should the handler itself obtain the event loop (with asyncio.get_running_loop() or however it's spelt) rather than accepting an event loop to run on?

wence- (approved):
One non-blocking question/suggestion.

Don't forget to drop the gb300 tests before final merge

python/ucxx/ucxx/_lib_async/listener.py
pentschev: The listener callback may run outside the asyncio event-loop thread, so `asyncio.get_running_loop()` is not reliably available there. There is also a race where the future may complete before `add_done_callback()` is registered, causing the callback to run immediately on the registering thread. Passing the loop captured when the listener was created ensures both submission and deferred cleanup are always scheduled on the correct event loop.

pentschev:
Thanks @wence- .

PR #764, opened 2026-09-22
Fix tag-multi request completion accounting

pentschev:
`tag_send_multi` now completes only after all locally submitted parts of the transfer have completed.

This fixes an incomplete-completion condition that could allow endpoint teardown to race with remaining tag-multi traffic, contributing to intermittent connection-reset failures. Includes CUDA coverage for expected and unexpected-message close ordering.


bdice (approved):

pentschev:
Thanks Bradley!

PR #765, opened 2026-09-22
Make Python teardown tests confirm peer delivery

pentschev:
Several Python client/listener tests sent their final response and immediately force-closed the endpoint. A successful UCXX send only reports UCX local completion, so the remote endpoint can still be waiting for the message when that close occurs. This can produce intermittent connection resets, particularly for multi-message tag transfers.

This change makes affected tests wait for an application-level receipt after the receiver has completed its work, before either side tears down the connection. It also removes scheduling races in probe and close-callback coverage, and adds clearer `recv_multi()` failure context.

The change does not alter UCXX’s delivery semantics or add a public API. It corrects test protocols that previously relied on a stronger guarantee than UCX provides.


TomAugspurger (approved):

bdice (approved):

pentschev:
Thanks Tom and Bradley, appreciate it!

PR #766, opened 2026-09-22
Confirm benchmark transfers before teardown

pentschev:
Benchmarks can finish their final transfer while the peer is still processing its response. If either side immediately closes its endpoint, this can surface as an intermittent connection reset rather than a valid completed benchmark.

This change adds a final receipt exchange before teardown, allowing the receiving peer to acknowledge that it processed the final response. The C++ unidirectional tag benchmark follows UCX's acknowledgement strategy, while the Python async and core backends apply the same ordering for tag and AM transfers. Failed receipt exchanges now terminate the Python benchmark cleanly instead of timing out.


python/ucxx/ucxx/_lib_async/tests/test_benchmark_backends.py
```diff
@@ -0,0 +1,154 @@
...
+        assert hasattr(response, "__cuda_array_interface__")
+        assert Array(response).cuda
+        cp.testing.assert_array_equal(cp.asarray(response), message)
+    finally:
+        await client.close()
+        listener.close()
+        await wait_listener_client_handlers(listener)
+
+
+def _cuda_available():
```
madsbk, line 96: Now CUDA gets initialized in the pytest process at collection time, I think this is bad?

python/ucxx/ucxx/benchmarks/backends/ucxx_async.py
```diff
@@ -191,6 +179,8 @@ async def run(self):
         if self.args.cuda_profile:
             xp.cuda.profiler.stop()
 
+        await _send_terminal_ack(ep, self.args.enable_am)
```
madsbk, line 184: The ACK makes the race much less likely, but I think it moves it rather than removing it?
As the new docstrings say, send completion is only local. So the side sending the ACK has the same problem the payload had. In the Python backends the client sends the ACK and exits right away. In C++ unidirectional the server sends it and then force-closes. Either way the 1-byte ACK can in principle still be dropped.

The failure is also worse than before. If `_recv_terminal_ack` raises in the `ucxx_async` server handler, `lf.close()` never runs and the server spins on `while not lf.closed` until the CI timeout. In C++ the client would hang in `waitRequests`.

Could the ACK sender flush the endpoint before teardown (`flushBuilder()` in C++, `ucp_ep_flush_nbx` gives remote completion)? Or if the remaining window is acceptable, I think we should at least close the listener in a `finally` so a failed ACK errors out instead of hanging?

python/ucxx/ucxx/_lib_async/tests/test_benchmark_backends.py
pentschev: Good catch. The collection-time `skipif` initializes CUDA in the pytest process, defeating the isolation provided by the subprocess. I moved CUDA availability detection into the subprocess and have the parent translate a dedicated unavailable status into a skip.

python/ucxx/ucxx/benchmarks/backends/ucxx_async.py
pentschev: Agreed on the async failure path. If the terminal ACK receive fails, the listener must still be closed so the server exits and reports the original error rather than timing out. I moved `lf.close()` into a `finally`.

Adding endpoint flush would have no use here, UCX endpoint flush does not flush tag or AM messages, so it cannot confirm delivery of this ACK. Furthermore, this benchmark intentionally retains force-close behavior because graceful close can block when a peer terminates abruptly, like we prefer in all of UCXX for the same reason. The C++ unidirectional flow also follows UCX perftest's acknowledgement strategy, see the [here](https://github.com/openucx/ucx/blob/5f041ecfecb04f3cc0fadd735b72c6d85d1210c5/src/tools/perf/lib/ucp_tests.cc#L936-L960).

The ACK improves the payload/teardown ordering but does not make local send completion a general remote-delivery guarantee. I'll adjust the wording accordingly.

pentschev (commented):
Thanks @madsbk , I've addressed and responded your comments, please have another look when you have a chance.

madsbk (approved):
Thanks @pentschev, LGTM

pentschev:
Thanks Mads!

